PR #190, opened 2022-09-01
fix libfreetype linking error on MacOS

Extremys:
All is in the title ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -445,6 +445,14 @@ if (SLIC3R_STATIC AND NOT SLIC3R_STATIC_EXCLUDE_CURL)
     endif()
 endif()
 
+IF(APPLE)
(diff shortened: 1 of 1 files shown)


lanewei120 (commented):
thanks for the commit

I have tried the building under MacOS,
it can be built successfully

have you tried to building freetype in the deps directory firstly?
as it is a newly added module, we need to update the building in deps

deftdawg:
Thanks for the PR, I think it has now been superseded by the following commit https://github.com/bambulab/BambuStudio/commit/5b7cf54cf5370afeaec8158ac6eeb61cc2f7c14e which included your fix as well as fixes for Linux to be able to build.

Extremys:
Ok thank you! I am closing so! :)

PR #251, opened 2022-09-11
Support independent external perimeter fan speed

jschuh:
This adds a new filament setting that is most useful for improving the surface finish of higher temperature filaments (e.g. PETG, PC, and ABS) while simultaneously using lower fan speeds (or no fan) for interior perimeters and infill (for better layer adhesion and part strength).

This is a port of [PrusaSlicer PR 2921](https://github.com/prusa3d/PrusaSlicer/pull/2921) that I've been using longterm with very good results.

### DIFF
--- a/src/libslic3r/GCode/CoolingBuffer.cpp
+++ b/src/libslic3r/GCode/CoolingBuffer.cpp
@@ -722,7 +722,11 @@ std::string CoolingBuffer::apply_layer_cooldown(
     new_gcode.reserve(gcode.size() * 2);
     bool overhang_fan_control= false;
     int  overhang_fan_speed   = 0;
-    auto change_extruder_set_fan = [ this, layer_id, layer_time, &new_gcode, &overhang_fan_control, &overhang_fan_speed]() {
+    bool external_perimeter_fan = false;
+    int  external_perimeter_fan_speed = 0;
+    bool in_external_perimeter = false;
+    auto change_extruder_set_fan = [ this, layer_id, layer_time, &new_gcode, &overhang_fan_control, &overhang_fan_speed, &external_perimeter_fan, &external_perimeter_fan_speed, &in_external_perimeter]() {
+        in_external_perimeter = false;
 #define EXTRUDER_CONFIG(OPT) m_config.OPT.get_at(m_current_extruder)
         int fan_min_speed = EXTRUDER_CONFIG(fan_min_speed);
         int fan_speed_new = EXTRUDER_CONFIG(reduce_fan_stop_start_freq) ? fan_min_speed : 0;
@@ -737,6 +741,8 @@ std::string CoolingBuffer::apply_layer_cooldown(
             close_fan_the_first_x_layers = 1;
         }
         if (int(layer_id) >= close_fan_the_first_x_layers) {
+            external_perimeter_fan = EXTRUDER_CONFIG(external_perimeter_fan);
+            external_perimeter_fan_speed = EXTRUDER_CONFIG(external_perimeter_fan_speed);
             int   fan_max_speed             = EXTRUDER_CONFIG(fan_max_speed);
             float slow_down_layer_time = float(EXTRUDER_CONFIG(slow_down_layer_time));
             float fan_cooling_layer_time      = float(EXTRUDER_CONFIG(fan_cooling_layer_time));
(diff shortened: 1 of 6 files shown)


SaltWei:
Very impressive feature. Could you also share the setting of wall fan speed for ABS, PETG, PC etc?
We want to evaluate and update the setting as well if merge this feature.

jschuh:
Sure, but I should state that I do not yet own a Bambu Labs printer to test with, so my input is based on experience with my own custom build of Prusa Slicer, where I've used this patch quite extensively over the years across a broad range of printers and filaments. Hopefully that's applicable enough here, but I want to be clear that I simply ported my code and did a quick sanity check of the output. I haven't really tested the ported code, and certainly not on the X1 hardware.

So, for my custom build of Prusa Slicer my strategy for filaments like ABS, PC, and PETG is to start with an existing filament preset, set `Min` to 0 and `External perimeter fan speed` to match the `Bridges fan speed` for that filament. Usually that simple change works well enough for any existing filament preset, but sometimes I have to fine tune a bit (e.g. raise `Min` a bit above zero or lower `External perimeter fan speed` a bit).

I have noticed that Bambu Studios presets run the fan cooling quite a bit higher than Prusa's (I assume this is related to the dual fan setup with the curtain fan used for heavy cooling), but I'd still take the same approach. So for PC, PETG, and ABS I would start by setting `Min fan speed threshold` to 0, and then set `External perimeter fan speed` to whatever value `Fan speed for overhangs` is. 

As an example, "Bambu PC" currently uses these settings:
* `Min fan speed threshold`: 10%
* `Fan speed for overhangs`: 80%

So, my suggestion would be:
* `Min fan speed threshold`: 0%
* `Fan speed for overhangs`: 80%
* `External perimeter fan speed`: 80%

At a minimum this is a good starting point. One thing I don't know about the X1 hardware/firmware is what the startup time on your fan is. I mostly use custom built printers running Klipper, where each printer has the [`kick_start_time`](https://www.klipper3d.org/Config_Reference.html#fan) set to a value that ensures it spins up fast from a stop. Depending on your spin-up time due to the fan hardware/firmware on the X1, you might not want to take `Min fan speed threshold` all the way down to 0%, and instead keep it at a very low speed like 10% or so.

SaltWei:
Great! Thank you very much for sharing these precious experience.
We will download this patch and conduct experiment on Bambu X1 hardware.

Will update after finishing.

SaltWei:
Hi @jschuh ， I think we miss something again.
The idea to set separate fan speed for outer wall is great. But I think maybe we don't need to add new options "external_perimeter_fan" and "external_perimeter_fan_speed". Reason is shown below.

BambuStudio already has options to force fan speed for overhang wall and bridge according to different overhang degree of wall. This is a kind of forcing cooling for walls.

force cooling

If we add "external_perimeter_fan" and "external_perimeter_fan_speed", it conflicts with the existing cooling strategy.

The right way is enhancing the existing cooling strategy, add 0% overhang threshold choice to indicate cooling for all external perimeter(including bridging perimeter and overhang perimeter).

jschuh:
Oh wow, I hadn't realized how much that code changed from the Prusa Slicer implementation, but taking a closer look I now see that it certainly can be used to provide all the necessary capability (and more). Sorry, I should have developed a better understanding of the Bambu Studio changes before trying to port my PR over. I hope I didn't waste too much of your time.

If it's not too much trouble, I do have one question about the UX. Right now [`overhang_fan_threshold`](https://github.com/bambulab/BambuStudio/blob/cb2e89d93e839a8900ddbfcdd6933c83ddb7584d/src/libslic3r/PrintConfig.cpp#L519) is implemented as an enum, so it isn't possible to set it to 0% (or any values outside the those enumerated). So, I wondered why it wasn't implemented as a normal percentage field that would allow the full range?

SaltWei:
> If it's not too much trouble, I do have one question about the UX. Right now [`overhang_fan_threshold`](https://github.com/bambulab/BambuStudio/blob/cb2e89d93e839a8900ddbfcdd6933c83ddb7584d/src/libslic3r/PrintConfig.cpp#L519) is implemented as an enum, so it isn't possible to set it to 0% (or any values outside the those enumerated). So, I wondered why it wasn't implemented as a normal percentage field that would allow the full range?

Very happy that you can provide this great idea and we think it's valuable.

About why it's enum. Because calculating the overhang degree of every segment of extrusion line by 2D graphics method is difficult and very computationally expensive, we sample the overhang degree into several gears.

We will continue the work to support force fan speed for all external perimeter, by adding 0% overhang threshold as special handling. It will be recorded as feature request issue.

SaltWei:
This feature has been supported in latest 1.4 release version and no defaultly been applied for PA-CF
image

PR #252, opened 2022-09-11
Don't use bridge flow over dense infill

jschuh:
Make infill bridging use the same void threshold check as other bridging spans. Reduces print time and improves layer adhesion over dense infill.

This is a port of [PrusaSlicer PR 4374](https://github.com/prusa3d/PrusaSlicer/pull/4374) that I've been using longterm with very good results.

### DIFF
--- a/src/libslic3r/PrintObject.cpp
+++ b/src/libslic3r/PrintObject.cpp
@@ -1569,10 +1569,6 @@ void PrintObject::bridge_over_infill()
     for (size_t region_id = 0; region_id < this->num_printing_regions(); ++ region_id) {
         const PrintRegion &region = this->printing_region(region_id);
 
-        // skip bridging in case there are no voids
-        if (region.config().sparse_infill_density.value == 100)
-            continue;
-
 		for (LayerPtrs::iterator layer_it = m_layers.begin(); layer_it != m_layers.end(); ++ layer_it) {
             // skip first layer
 			if (layer_it == m_layers.begin())
@@ -1582,6 +1578,19 @@ void PrintObject::bridge_over_infill()
             LayerRegion *layerm      = layer->m_regions[region_id];
             //BBS: enable thick bridge for internal bridge only
             Flow         bridge_flow = layerm->bridging_flow(frSolidInfill, true);
+
+            // Minimum width of a void for the extrusion above to be treated as a bridge.
+            // Scaled based on layer height, with a 3mm void at 0.2mm layer height as the baseline).
+            const float min_void_width = float(bridge_flow.scaled_width()) * 3.f * layerm->m_layer->height / 0.2;
+
+            // Skip bridging if the infill voids are small enough.
+            const float fill_density = region.config().sparse_infill_density.value / 100.0;
+            if (fill_density > 0.0) {
+                // Estimate the infill void width from infill density and extrusion width.
+                const float infill_extrusion_width = layerm->flow(frInfill).scaled_width();
+                if (min_void_width > (infill_extrusion_width / fill_density - infill_extrusion_width))
+                    continue;
+            }
 
             // extract the stInternalSolid surfaces that might be transformed into bridges
             Polygons internal_solid;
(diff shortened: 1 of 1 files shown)


SaltWei:
This is a useful feature and it can fix some issue I met. The issue I met is over-extrusion when using bridge flow over dense infill, because Bambu Studio always enable thick bridge for internal bridge.

This patch not only can fix the old issue I met, but also can speed printing up. Impressive and perfect.

But I notice that the sparse infill density to trigger this logic is about 26-28%. Is this possible to be too low?
Because Bambu Lab X1 machine is fast and we always use a high speed to print internal solid infill. I worry about such fast solid infill speed over sparse infill which density is lower than 30% is not safe enough.

Printing experiment has been started with different nozzle diameter and different filament.

SaltWei:
This is a useful feature and it can fix some issue I met. The issue I met is over-extrusion when using bridge flow over dense infill, because Bambu Studio always enable thick bridge for internal bridge.

This patch not only can fix the old issue I met, but also can speed printing up. Impressive and perfect.

But I notice that the sparse infill density to trigger this logic is about 26-28%. Is this possible to be too low?
Because Bambu Lab X1 machine is fast and we always use a high speed to print internal solid infill. I worry about such fast solid infill speed over sparse infill which density is lower than 30% is not safe enough.

Printing experiment has been started with different nozzle diameter and different filament.

jschuh:
The exact infill threshold depends on both the infill percentage and the extrusion/nozzle width (for wider extrusions the percentage is higher). If you take a closer look at the code you'll see that I used the same 3mm span threshold that Slic3r derivatives have always used for skipping normal bridging, so empirically it should be safe.

I've also been using this patch for a few years now on Vorons and other custom printers running Klipper, where I normally print infill around 200-300m/s and 2000-3000mm/s² acceleration. So, at least for my use it's always worked well, even at high printing speeds.

SaltWei:
> The exact infill threshold depends on both the infill percentage and the extrusion/nozzle width (for wider extrusions the percentage is higher).

I think maybe we miss some things. The layer height should also be a factor. For smaller layer height, should be more difficult to trigger this logic. I mean more higher sparse infill density is necessary to support solid infill which has smaller layer height.
Otherwise the first solid layer above sparse infill will be very bad.

SaltWei:
Here is my printing result.
I printed with 0.4 nozzle and 0.08 layer height. infill density is 28%. The left is without the patch. The right is with the patch. And we can see the right one and the first solid layer above sparse infill is very bad. All lines are broken.
008mm
After finishing the whole printing, the top surface of right one is still not good enough. But the left is OK. My top solid layer number is 9 layers and thick enough.

008mm1

jschuh:
That's a very good point. I generally print larger, structural parts, where I tend to use layer heights of 0.16 - 0.3. However, at smaller layer heights there just doesn't seem to be enough extrusion volume to bridge the span. (Sorry, I should have considered smaller layer heights, but I write this code mostly to solve my own issues.)

It would be easy for me to modify the patch to further scale the span threshold based on layer height. I'm thinking something like using 3mm span at .2mm layer height as the baseline, which would look like this:

```
adjusted_span = 3.0 * layer_height / 0.2.
```

That would put a 0.08mm layer height at a 1.2mm span (requiring a bit more than twice the infill density). This also means that the span would grow larger at higher layer heights, but intuitively that makes sense to me, since at larger layer heights there should be enough extrusion volume to bridge a larger span. Although, that would probably require some testing with bigger nozzles to properly verify performance at larger layer heights.

What do you think?

SaltWei:
> adjusted_span = 3.0 * layer_height / 0.2.

This equation is a good example. The smaller the layer height, the smaller the adjust span. But currently I think maybe more conservative method than this equation to calculate span is more safe. We will test with different conditions and give a conclusion.

This patch is great and a perfect direction to improve internal bridge issues both on quality and printing speed.

For what value span is fine, I (and other guys in our team) will also try to build one mathematical model.
Not easy because it must have relationship with cooling and mechanics of material and melting, and I am really not good at these parts.

jschuh:
Yeah, makes sense. I also took a look at the code again and it will need a little bit of restructuring to account for layer height. As currently written it just looks at the infill density for the current layer in a preamble loop, but if it needs to account for the height of the layer above I'll have to move the logic into the actual bridge detection loop. That shouldn't be hard, but I just won't have time to get to it until tomorrow or the day after.

I'm happy to do that and let your team figure out the best values for the span thresholds if that works for you.

jschuh:
Well, turns out it was a trivial change to scale by layer height. I hadn't looked closely at the code in a few years and must have overestimated the complexity. 

So, baselining on a 3mm gap at 0.2mm layer height gets you these cutoffs for bridging infill at a 0.45mm extrusion width:

| Layer Height  | Bridge infill threshold |
| ------------- | ------------- |
| 0.3mm | 20% |
| 0.2mm | 25% |
| 0.16mm |  32% |
| 0.08mm |  49% |

Certainly warrants some empirical testing to get the exact values, but intuitively this feels like it's in the right range.

celtare21:
Any updated on this?

PR #319, opened 2022-09-24
Fix an issue that bed temperature for other layers is not set properly

SoftFever:


lanewei120:
thanks for the submission

PR #518, opened 2022-10-26
Fix a crashing issue when the model's path has Unicode characters on third party printers.

SoftFever:
Steps to reproduce:
1. Load a model with Unicode characters in its path:  `c:\xxx\中文路径\a.stl`
2. Slice it and click the `Print` button
3. Application clashes.

-SoftFever


StoneLiBambulab:
Cherry-picked!

Thanks!
Stone Li

PR #617, opened 2022-11-11
Fix the spelling of nozzle

pdelvo:
I fixed some places where nozzle was spelled "noozle" instead of "nozzle" in the comments of generated G code


lanewei120:
thanks very much for these findings
as the start gcode are from our auto submit flow internally
we will fix it there

PR #696, opened 2022-11-22
Fix build with upstream dependencies

zhaofengli:
Hello there,

This PR fixes build against newer dependencies like Boost 1.79 shipped with Linux distributions (specifically, NixOS). For distribution packaging, it's expected for applications to use libraries from the package set instead of bringing in their own vendored versions.

- `Fix repeated calls to FindTBB in module mode.`: Upstreamed in PrusaSlicer as https://github.com/prusa3d/PrusaSlicer/commit/eddcebf929c507b0a37c96b29a20f877e405dc55

Ref: #12

PR #867, opened 2022-12-16
update to remove non-UTF-8 characters and translate comment

ajacocks:
This is a simple fix to work around being unable to build on Linux, due to a failure in a UTF-8 character check


ajacocks:
Maybe duplicates part of #696

lanewei120:
yes, already solved by another submit before

thanks for doing this

PR #873, opened 2022-12-17
BuildLinuxImage.sh: Remove workarounds that are no longer needed

zhaofengli:
Removing some workarounds that are no longer necessary.

Ref: #12


lanewei120:
oops, we have uploaded a commit which has the similiar function with this commit "[GUI_App: Create data dir if it doesn't exist on Linux]"
please help to revert it

thanks a lot for doing this!

deftdawg:
Was going to do a PR for this, both workarounds to AppImage's AppRun are no longer required because of better darkmode support and the fix to create the directory if it's missing https://github.com/bambulab/BambuStudio/commit/c25c1ba4eb0b267feac034618419385296166cd9 (this is the right way because BS would crash outside of an AppImage without it)

thx @zhaofengli!

PR #923, opened 2022-12-21
deps/wxWidgets: Disable EGL for wxGLCanvas

zhaofengli:
This works around https://github.com/supermerill/SuperSlicer/issues/1093 which causes the build plate to not be visible in some setups.

It's the same workaround used by Nixpkgs for its PrusaSlicer and SuperSlicer packages until it switched to Prusa's wxWidgets 3.1.4.

Ref: #12

### DIFF
--- a/deps/wxWidgets/wxWidgets.cmake
+++ b/deps/wxWidgets/wxWidgets.cmake
@@ -49,6 +49,7 @@ bambustudio_add_cmake_project(wxWidgets
         -DwxUSE_LIBJPEG=sys
         -DwxUSE_LIBTIFF=sys
         -DwxUSE_EXPAT=sys
+        -DwxUSE_GLCANVAS_EGL=OFF
 )
 
 if (MSVC)


fang64:
This PR fixes the graphical render issues on Ubuntu 22.04 LTS and I imagine other distros.

jwise:
I think a potentially better change might be to update the bundled GLEW to support EGL.  (This would also pipeclean support for Wayland.)  For some reason, the GLEW that's packed in is truly ancient; if it were updated to support EGL, it would probably fix this the other way around.

zhaofengli:
> I think a potentially better change might be to update the bundled GLEW to support EGL.

Nixpkgs carries the latest glew version (2.2.0) and the same issue exists there (*), so the underlying issue is elsewhere.

(*) Previously we were disabling EGL in wxWidgets as a workaround, but then switched to Prusa's 3.1.4 fork which doesn't have EGL enabled by default

jwise:
I think that Bambu Slicer overrides GLEW to use the internal one no matter what, even if there is a system GLEW?  So Nix carrying GLEW 2.2.0 I *think* shouldn't change anything.  (Now, that said, I'm reading around the internet that GLEW for EGL is incomplete, which is giving me serious 'how does anything ever work?' vibes.  Which features of GLEW does Slic3r/PS/SS/BS use?  Hard to know until I try it, I guess!)

Mostly I don't really want to paper over this because I am worried that the lack of Wayland support is causing me problems with the Gstreamer stuff I'm working on (#12), and disabling EGL would close the door on Wayland indefinitely, since Wayland doesn't have GLX.

fang64:
@jwise @zhaofengli I incorporated the change with a linux-only build option for the script for SoftFever's fork of this, I did incorporate those changes into my PR so it doesn't effect windows builds and doesn't require the library to change. At least atm.

jwise:
Yeah, I'm absolutely mired in the Gstreamer stuff and I have no idea if I'm going to ever emerge onto the EGL side of it.  Building with this temporarily is probably a reasonable plan.

jwise:
Try #1053?  #1053 enables EGL instead of GLX by upgrading GLEW.  You'll probably have to wipe your `build/CMakeCache.txt` to make it work, though (and you will have to rebuild `deps` -- I have a change coming to make that less painful, too, though).

fang64:
> Try #1053? #1053 enables EGL instead of GLX by upgrading GLEW. You'll probably have to wipe your `build/CMakeCache.txt` to make it work, though (and you will have to rebuild `deps` -- I have a change coming to make that less painful, too, though).

I'm building this all in docker/github actions I just blow away the build environment so no biggie

jwise:
@zhaofengli I think #1053 has been merged into master.  See if that works for you?  If so, we can close this one.

zhaofengli:
Sorry for the late response. I just tried it and it worked well! Thanks for the amazing work @jwise and I'll close this PR.

PR #1011, opened 2023-01-04
REPAIR for MAC

christianjou:
When do you plan to add REPAIR STL for MAC please ?


lanewei120:
it is not a simple job
I had to say it is in our plan, but don't have a fixed schedule

we have tried serveral repair functions before
however, there are lots of corner cases we can not do well

and we don't have a good solution currently

PR #1023, opened 2023-01-07
Update BambuStudio_de.po translations (till line 1000)

Huseriato:
As a native speaker, I have adapted all translations up to line 1000. Now I would like to see if my work is needed and appreciated. I would continue if this is so.

### DIFF
--- a/bbl/i18n/de/BambuStudio_de.po
+++ b/bbl/i18n/de/BambuStudio_de.po
@@ -20,7 +20,7 @@ msgid "Section view"
 msgstr "Abschnittsansicht"
 
 msgid "Reset direction"
-msgstr ""
+msgstr "Ausrichtung zurücksetzen"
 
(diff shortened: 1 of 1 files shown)


hliebscher:
pls show my file and correct them pls

https://github.com/bambulab/BambuStudio/pulls#:~:text=ago%20by%20christianjou-,fix%20the%20most%20german%20words,-%231008%20opened%204

lanewei120:
Thanks for doing this

I have sent this patch to our translation group
and wait their feedback

lanewei120:
@Huseriato 

Hello

I have confirmed with our translation team
He is doing the same thing and has already done some similar fix as in this patch

as our translation flow needs some time to update to the newest version
we have not submitted this modification yet

so I think you can stop this work
Sorry about this

Again, thanks very much for doing it~

PR #1028, opened 2023-01-08
SIDEWINDER X1

christianjou:
Hi, could you add the SIDEWINDER X1 please ?


lanewei120:
we will do it later
we can not pick all the commits from softever's git
there are lots of conflicts

for new machine support, you can file a issue in this github

close this pull request

PR #1047, opened 2023-01-10
test

lanewei120:


lanewei120:
conflicts

tidus1492:
Hi Lane, I'm wondering if there is any way to add additional fonts to Bambu Studio? Have tried many things with no luck.

bambu123:
For source code, see the 'Label.cpp', modify the default font to use
For Windows, install font to the operating system.
For Linux, also see the 'Label.cpp', where the font is installed at run time

PR #1053, opened 2023-01-11
Linux: update GLEW to 2.2.0, and enable EGL support in GLEW to match wxWidgets

jwise:
On Linux, wxGTK by default attempts to use EGL if it is available on the system, rather than GLX.  Unfortunately, the ancient version of GLEW that we packaged in did not support EGL, and even if it did, the configuration was not set up to enable EGL.  To solve this, we:

  * upgrade GLEW to version 2.2.0, from upstream GitHub

  * modify the Bambu build process to enforce that we use GLEW from the built dependency

  * remove the "extra" even older GLEW that was packaged

  * modify GLEW's CMake configuration to enable EGL support when it is available on the system (using the same test as wxWidgets uses to decide whether to enable EGL support); if EGL isn't available at compile time, both GLEW and wxWidgets will fall back on GLX

Note that you probably will have to blow away your CMakeCache for this to work correctly -- otherwise, you may end up with the system GLEW, if you have one installed (which is probably not what you want -- on Ubuntu, the system GLEW is GLX, not EGL).

This supersedes #923.


jwise:
cc: @lanewei120

lanewei120:
thanks for doing this

PR #1054, opened 2023-01-11
deps/wxWidgets: avoid patching too many times

jwise:
Each time the dependency build was run, previously, the build system attempted to patch wxWidgets after checking it out from Git.  The problem, of course, is that if this happened once, it would not succeed a second time, so the only workaround was to blow away the wxWidgets source tree.

The real solution to this is to create a BBL fork of wxWidgets (or to upstream the changes...).  But for now, we add a file to determine whether the patch has taken place already, and if it's there, we don't apply the patch again.  This will mean that all kinds of exciting things happen if you change Git revisions of wxWidgets or the patch changes (in those cases, you'll have to blow away the build), but at least this makes it possible to build twice in the same repository in the best case.

To update an existing checkout, run:

```
$ touch deps/build/dep_wxWidgets-prefix/src/dep_wxWidgets/WXWIDGETS_PATCHED
```

### DIFF
--- a/deps/wxWidgets/wxWidgets.cmake
+++ b/deps/wxWidgets/wxWidgets.cmake
@@ -18,10 +18,16 @@ else ()
     set(_wx_edge "-DwxUSE_WEBVIEW_EDGE=OFF")
 endif ()
 
+if (MSVC)
+    set(_patch_cmd ${PATCH_CMD} ${CMAKE_CURRENT_LIST_DIR}/0001-wxWidget-fix.patch)
+else ()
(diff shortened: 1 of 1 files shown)


jwise:
cc: @lanewei120

lanewei120:
this can only work on linux

on windows, "test" and "touch" can not work under the 'x86 Native Tools Command Prompt for VS 2019'

jwise:
Hmm, that's sad.  Ok, well, how about this change instead?  Same thing, just Mac and Linux only.  You're on your own for getting a stable build on Windows :)

lanewei120:
it is ok now

thanks for doing this

PR #1062, opened 2023-01-12
Boost: merge the PATCH_COMMANDs into one

jwise:
On my Linux machine, the new Boost patch seems to generate a Makefile that does not work (the `git init` gets passed to the `cmake copy` part of the patch).  Glue these together into one `PATCH_COMMAND` to allow the build to pass.

### DIFF
--- a/deps/Boost/Boost.cmake
+++ b/deps/Boost/Boost.cmake
@@ -128,8 +128,7 @@ ExternalProject_Add(
     URL_HASH SHA256=f22143b5528e081123c3c5ed437e92f648fe69748e95fa6e2bd41484e2986cc3
     DOWNLOAD_DIR ${DEP_DOWNLOAD_DIR}/Boost
     CONFIGURE_COMMAND "${_bootstrap_cmd}"
-    PATCH_COMMAND ${_patch_command}
-    PATCH_COMMAND git init && ${PATCH_CMD} ${CMAKE_CURRENT_LIST_DIR}/0001-Boost-fix.patch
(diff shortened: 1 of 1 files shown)


jwise:
Really??  That's so weird.  Why does `git init && ${PATCH_CMD} ...` work, but `${_patch_command} && git init && {PATCH_CMD}` not work?  I have to say, I don't really understand how Cmake ends up generating this patch stuff.

Uh, I guess I could make this change Linux/macOS-only...?  Does the original work for you at all on Linux without this change?

lanewei120:
previously the engineer only verifies on windows
I just tried on linux, without the change, it doesn't work:)

and on windows, i am also curiously why previous command works, but it really patches successfully...

lanewei120:
I think just add the platform flag is ok
could you add the platform judge and submit it again?

jwise:
@lanewei120 I probably won't get to it for another day or two, so if you guys get a chance first, ok with me.  There also seems to be another regression that submitting print jobs fails with a 'std::alloc error' but I need to more carefully track that down.

lanewei120:
already fixed it by https://github.com/bambulab/BambuStudio/pull/1097
@spuiuk , please fetch the newest codes from master

lanewei120:
> @lanewei120 I probably won't get to it for another day or two, so if you guys get a chance first, ok with me. There also seems to be another regression that submitting print jobs fails with a 'std::alloc error' but I need to more carefully track that down.

ok, we submitted it here

thanks

PR #1144, opened 2023-01-22
Change Filament across single plates and not globally

sisuwest:
Presently if I wanted to change filament on a single plate I can use the slider on the right and right click to select a new filament on a certain layer. If I have multiple plates, that same layer height color change affects all models on all plates. That filament change should only occur on that single plate. 

Here is a video showing what I'm talking about to another user. https://youtu.be/97ZZ5lvRf0U


sisuwest:
Wrong thread

PR #1164, opened 2023-01-24
fix crashes when switching timelapse and vase mode

SoftFever:


SoftFever:
#1165

lanewei120:
@SoftFever 
thanks very much

I have tried this patch
it still crashes when switch timelapse_mode with boost::any_cast

I just have a debug and commit a patch here
https://github.com/bambulab/BambuStudio/commit/69b884e5851ff2d54f59ed186133925ee6c0845b

Please use this one for a try

SoftFever:
@lanewei120 do you have a project file to repro the crash?

lanewei120:
just a new project is ok

timelapse_crash

SoftFever:
Hmm, I probably have other changes along with this change that fixed the crash in my folk.
You might want to check and test this project as well https://github.com/SoftFever/BambuStudio-SoftFever/issues/175

lanewei120:
yes, with that patch, it should be ok~

we also added a fix here, https://github.com/bambulab/BambuStudio/commit/69b884e5851ff2d54f59ed186133925ee6c0845b
it can fix this issue too:)

I am afraid there will be conflicts, 
we can merge this patch but can not merge that patch now...
anyway, both methods should work

thanks for the submission

PR #1197, opened 2023-01-28
Linux: MediaPlayCtrl: make playback more reliable with wxUSE_GSTREAMER_PLAYER for systems that have that

jwise:
On Linux, there are actually two ways to build a wxWidgets video player -- Gstreamer, and gstreamer-player.  (I have no idea why they did not just pick one.)  The gstreamer-player path is a little different, and did not reliably start the stream previously.  (The symptom is that when built on a system that has the `libgstreamer-plugins-bad1.0-dev` package installed, the video player would stay black and would say "Loading..." forever, but if you right-clicked on the play button, it would in fact start playing.)  This collection of hacks gets gstreamer-player working.


jwise:
cc: @lanewei120

lanewei120:
thanks for doing this

PR #1198, opened 2023-01-28
Address Sanitizer stability and correctness cleanups

jwise:
Some modern compilers, including GCC, Clang, and MSVC, support a mode called "address sanitizer" ("asan"), invoked with `-fsanitize=address`, which adds extra tracking around all memory accesses.  Compiling Bambu Studio with asan turned on revealed a handful of memory usage errors; some of them are not critical, but some of them do seem quite important and could result in stability issues on any platform, not just Linux.

While I was in here, I also tracked down the crash associated with Linux background startup, and I *think* I fixed it.  It's very difficult to reproduce, though, so it's hard to know for sure.

### DIFF
--- a/src/slic3r/GUI/GLCanvas3D.cpp
+++ b/src/slic3r/GUI/GLCanvas3D.cpp
@@ -1706,6 +1706,9 @@ float GLCanvas3D::get_collapse_toolbar_height()
     return collapse_toolbar.is_enabled() ? collapse_toolbar.get_height() : 0;
 }
 
+bool GLCanvas3D::make_current_for_postinit() {
+    return _set_current();
+}
 
 void GLCanvas3D::render(bool only_init)
 {
--- a/src/slic3r/GUI/GLCanvas3D.cpp
+++ b/src/slic3r/GUI/GLCanvas3D.cpp
@@ -2157,15 +2157,16 @@ void GLCanvas3D::reload_scene(bool refresh_immediately, bool force_full_scene_re
 
     struct GLVolumeState {
         GLVolumeState() :
-            volume_idx(size_t(-1)) {}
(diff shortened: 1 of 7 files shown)


jwise:
cc: @lanewei120

src/slic3r/GUI/GLCanvas3D.cpp
```diff
@@ -2368,8 +2372,21 @@ void GLCanvas3D::reload_scene(bool refresh_immediately, bool force_full_scene_re
...
         }
         for (int temp_idx = vol_idx; temp_idx < m_volumes.volumes.size() && !update_object_list; temp_idx++) {
-            if (!m_volumes.volumes[temp_idx]->is_wipe_tower)
+            // Volumes in m_volumes might not exist anymore, so we cannot
+            // directly check if they are is_wipe_towers, for which we do
+            // not want to update the object list.  Instead, we do a kind of
+            // slow thing of seeing if they were in the deleted list, and if
+            // so, if they were a wipe tower.
+            bool was_deleted_wipe_tower = false;
+            for (int del_idx = 0; del_idx < deleted_volumes.size(); delidx++) {
```
jwise, line 2381: How did this ever work?  I cannot understand what happened.  This *did* compile once, at least, I think, though.  I'll have a fixup commit for this in a moment.

jwise: Fixed.  I have no idea how I ever got this to compile to begin with.  Sigh.

lanewei120:
@jwise 
hello

how about the two suggestion mentioned above?

thanks

jwise:
@lanewei120 The comment above was from me apparently pushing an old version.  Should be ready to merge now.  I've been running this version stably on my own machine for a while.

PR #1202, opened 2023-01-29
Bambu GUI: be much more careful about calling Layout() and Refresh()

jwise:
Repainting the screen is expensive and causes lots of idle CPU usage when the printer is continuously sending updates over the network (or a timer is simply firing); in some cases, it can even cause a SVG to continuously be rerendered to a bitmap.  In general, we should avoid causing widgets to repaint if they don't change; we add some checks to see if anything has actually changed inside of a widget before we actually commit to repainting.

This is a first step towards taking much finer-grained control over when these operations are called, rather than unconditionally calling them at 1Hz.  In the future, we should remove the 1Hz timer entirely -- and only call Layout on things that need their layout updated -- but in the mean time, this change substantially reduces the amount of CPU time (and battery!) that Bambu Studio consumes in the background when nothing is happening (particularly important for laptop users!).

This also reduces the annoying flickering in the video playback on Linux, but it probably will not entirely solve it.

Roughly, there are two major types of changes in here:
* `MachineObject` gets a set of `display_dirty_` variables, which are used to represent whether machine state has changed that needs to be updated.
* Individual widgets try to remember their previous state, and if they have been asked to change but their contents would remain the same, they do nothing.

### DIFF
--- a/src/slic3r/GUI/DeviceManager.cpp
+++ b/src/slic3r/GUI/DeviceManager.cpp
@@ -2326,38 +2326,45 @@ int MachineObject::parse_json(std::string payload)
                     }
 
                     if (jj.contains("mc_remaining_time")) {
+                        display_dirty_subtask = 1;
                         if (jj["mc_remaining_time"].is_string())
                             mc_left_time = stoi(j["print"]["mc_remaining_time"].get<std::string>()) * 60;
                         else if (jj["mc_remaining_time"].is_number_integer())
                             mc_left_time = j["print"]["mc_remaining_time"].get<int>() * 60;
                     }
                     if (jj.contains("mc_percent")) {
+                        display_dirty_subtask = 1;
                         if (jj["mc_percent"].is_string())
                             mc_print_percent = stoi(j["print"]["mc_percent"].get<std::string>());
                         else if (jj["mc_percent"].is_number_integer())
                             mc_print_percent = j["print"]["mc_percent"].get<int>();
                     }
                     if (jj.contains("mc_print_sub_stage")) {
+                        display_dirty_subtask = 1;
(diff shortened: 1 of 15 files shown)


walterwongbbl:
hello, 
thanks for this submission, we have some suggestions about your submission.

src/slic3r/GUI/DeviceManager.cpp 
src/slic3r/GUI/DeviceManager.hpp
*We don't want to add UI-related code in DeviceManager. And we need layout at some time because we need to adapt the UI for different model of machines.

src/slic3r/GUI/Monitor.cpp
src/slic3r/GUI/StatusPanel.cpp
src/slic3r/GUI/UpgradePanel.cpp
*Some codes can reduce local refresh

src/slic3r/GUI/Widgets/AMSControl.cpp
src/slic3r/GUI/Widgets/Button.cpp
*Some detection of the same color code can be used

src/slic3r/GUI/Widgets/ComboBox.cpp
src/slic3r/GUI/Widgets/ImageSwitchButton.cpp
src/slic3r/GUI/Widgets/ProgressBar.cpp
src/slic3r/GUI/Widgets/SideTools.cpp
src/slic3r/GUI/Widgets/StepCtrl.cpp
*The effect is not obvious

src/slic3r/GUI/Widgets/TempInput.cpp
src/slic3r/GUI/Widgets/TextInput.cpp
*Reduces the refresh times of the control,  But the number of controls is very small. So the effect is not obvious.

src/slic3r/GUI/wxExtensions.cpp
*It seems to be effective, but a lot of tests need to be done.

We're going to reference some of your code，But not all. 
Because other printers and hardware need to be considered, which will lead to UI re-layout. 
We will consider these factors.  So It will be more convenient for us to modify.

PR #1484, opened 2023-03-14
V1.5.0

lanewei120:


lanewei120:
used for generation image, close it not

PR #1523, opened 2023-03-21
feat: add basic Czech language support `cs`

petrleocompel:
Added basic Czech language support. 

Did not change anything else than `docs/Localization_guide.md` `Scenario 2. How do I add a new language support` said. 

I think from this it can just improve by using and feedback from users.


petrleocompel:
The translation is not perfect but we can improve from this. I definetly should review a lot of them again. 

Currently I am starting use my version of Bamboo Studio (my printer got lost in shipping :disappointed: so still not using it fully) and I allready see some problematic translations.

petrleocompel:
Got some new translations, checked from app and fixed. Also some problematic parts which are correct in `po` but characters are probably not in font. 

Snímek obrazovky 2023-05-02 v 10 26 50


Later i guess I will rebase my branch to keep everything updated.

petrleocompel:
Branche was rebased to reflect current branche. Also i will compare base lang files to cover latest changes.

lanewei120:
Hello @petrleocompel 

we have added Czech language support in newest version(v1.8 public beta)

and we have written a guide for translations https://wiki.bambulab.com/en/bambu-studio/contribute-languages
please check it for a reference

you can update the translations in the future following this guide

PR #1534, opened 2023-03-25
Updated Spanish translation

mainmind83:
Updated Spanish translation


lanewei120:
thanks for doing this

PR #1639, opened 2023-04-19
Docker Run:  Add gstreamer plugins for video support

calh:
These two packages are needed to stream video from the webcam

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ RUN apt-get update && apt-get install  -y \
     extra-cmake-modules \
     file \
     git \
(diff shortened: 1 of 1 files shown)


calh:
Hmm, this build just failed because:
```E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libx/libxrandr/libxrandr-dev_1.5.2-0ubuntu1_amd64.deb  Connection failed [IP: 40.119.46.219 80]
```

I guess the hamsters that run Azure are on break.

lanewei120:
thanks for doing this

PR #1803, opened 2023-05-14
Correction translation german

Th199403:


ChrMaass (approved):

lanewei120:
hello

thanks very much for doing this

we have asked our German tranlator to check, and he confirmed this modification is better than before

however, as our translation has a seperate flow(to co-work with the translator over the world)
it is better for us to update these words's translation directly to our database(we will do it internal)
I am afraid we can not merge it directly

sorry about it

lanewei120:
we have added it in out local translation file which will be valid on next release

thanks very much for doing this
and now close it

PR #1882, opened 2023-05-31
Fix building with GCC13 (required to build on Arch and Fedora)

c2h5oh:
Arch and Fedora switched to GCC 13.x as default making it impossible to build BambuStudio on those distros.

Only minor cleanup was required to fix that.


lanewei120:
thanks for doing this~

PR #1906, opened 2023-06-08
Initialize locale before wxTranslations::Get()

zhaofengli:
This fixes a crash similar to <https://github.com/prusa3d/PrusaSlicer/issues/8299>.


lanewei120:
thanks for doing this

PR #2005, opened 2023-07-09
Fix mcut for GCC 13

afzalarsalan:
GCC 13 requires an explicit #include <cstdint> for the std::uint32_t  data type used by this file. Adding this include allows for the whole project to successfully be compiled on GCC 13


lanewei120:
thanks very much for doing this

PR #2080, opened 2023-07-18
Add missing newlines in change_filament_gcode

adripo:
Added missing newlines to change:

```gcode
{if old_filament_temp > 142 && next_extruder < 255}M104 S[old_filament_temp]{endif}
```

into

```gcode
{if old_filament_temp > 142 && next_extruder < 255}
M104 S[old_filament_temp]
{endif}
```


lanewei120:
thanks for doing this

PR #2112, opened 2023-07-27
BambuLab machine profile cleanup

adripo:
I dedicated some time to organize, reorder, remove unused options and add missing ones to make the code more readable and easier to edit.


adripo:
@lanewei120 during this process I noticed that the P1P profile has a different `layer_change_gcode`.
The only different between the layer_change used for all the other bbl printers is this one:
image

The P1P has `C11` instead of `C10` in `M971 S11 C11 O0`.
After doing some research I have only found documentation for C10, but nothing for C11.

Can you please let me know if this is a typo or the `C11` command is correct?
Thank you

QingZhangBambu:
@adripo good job, we will review these commits. C11 has the same functionality as C10, but for older versions of the protocol.

adripo:
Thank you @QingZhangBambu 
Do you think that C11 is still necessary or C10 can be used on all the printers?
I have also seen that the same command is used some row above in the same file for the 
```
;timelapse without wipe tower
M971 S11 C10 O0
```

lanewei120:
@adripo 
thanks very much for doing these,
most of the suggestions will be adopted

could you rebase to the newest codes and upload again?

also, as we developed an auto tool to adjust the file and option orders
we will submit a patch based on your patches later generated by the tool which may have a bit difference

tangketan:
> Thank you @QingZhangBambu Do you think that C11 is still necessary or C10 can be used on all the printers? I have also seen that the same command is used some row above in the same file for the
> 
> ```
> ;timelapse without wipe tower
> M971 S11 C10 O0
> ```

Sorry that I gave wrong information about C11 and C10 to @QingZhangBambu before. They are not the same.  C11 is for smooth timelapse while C10 is for traditional.

PR #2154, opened 2023-08-02
Fix bug template

hho:
This fixes a few issues with the template I noticed while filing a bug:
* Wrong home directory for MacOS and Linux
* Missing endquote after Linux directory leading to broken formatting
* Reference to … Cura?
* Inconsistent spacing


lanewei120:
thanks for doing this

PR #2182, opened 2023-08-05
Port over Orca Slicer enhancement to optimise sandwich wall mode when perimeters are shared with multiple islands(Arachne only). Fix quality issue with "Single wall on top surface" 

igiannakas:
Port over Orca Slicer enhancement to optimise sandwich wall mode when perimeters are shared with multiple islands(Arachne only).

[SoftFever#1352](https://github.com/SoftFever/OrcaSlicer/pull/1352)

Also ported over fix for single wall top surface quality issue:
https://github.com/SoftFever/OrcaSlicer/issues/1680

Thank you to Soft Fever for cleaning up & refactoring the code on this! And for the fix for the single wall quality issue

Detail for the inner-outer-inner optimisations:
This is a PR containing a change in approach to handling sandwich mode when using Arachne engine.

In a nutshell instead of treating the perimeters as inside to outside and re-ordering the last extrusion it flips it to treat them as outside to inside and swaps the order of printing the third wall with the outer perimeter so the order is internal, external, internal and then the remaining internal walls from the outside - in.

This means that when less than 3 extrusions exist the wall is treated as outside to inside, leading to a better surface finish as there is no variation due to printing an external perimeter with and without an adjoining internal perimeter

This is a revised version of the PR https://github.com/SoftFever/OrcaSlicer/pull/751 addressing issue https://github.com/SoftFever/OrcaSlicer/issues/745 with more extensive logic on the wall ordering to ensure that all cases where an external perimeter is present on an island are re-ordered correctly, even if there are multiple external perimeters within the same island.


QingZhangBambu:
OK, I will review and test that. Thank you for your work.

QingZhangBambu:
image
image

@igiannakas I test your commit which adjust "single wall top surface quality", found it may cause surface quality problem in this type of case.

QingZhangBambu:
image
image

igiannakas:
I've noticed the same for a few edge case scenarios. Maybe adding this as a config option is the best - as what Orca slicer has done.

QingZhangBambu:
@igiannakas Yes, we tend to do that.

QingZhangBambu:
@igiannakas i think the wall order adjustment could be apply to Classic mode, do you want to push a commit yourself or just i modify it?

igiannakas:
If you could modify the classic one that would be great as I am not familiar with that mode in the code.On 22 Aug 2023, at 10:20, BambuStudioBugProducer ***@***.***> wrote:﻿
@igiannakas i think the wall order adjustment could be apply to Classic mode, do you want to push a commit yourself or just i modify it?

igiannakas:
If you could modify the classic one that would be great as I am not familiar with that mode in the code.

I've also made a small tweak to the algorithm as it had an edge case where it was not reordering properly the perimeters of another island on the same layer.

See screenshots below

Before:
261783965-f61f8853-d1de-4ba6-a868-a5ca36f989df


After:
261783986-2cc10917-ad8c-478f-a8ce-ace74acd7c5e


Please use the code from the latest update.

igiannakas:
> @igiannakas if you don't mind, could you provide this model? I want to test it. image

Of course, it’s this one: https://files.printables.com/media/prints/56731/stls/563074_47d87a59-8f93-4d9d-a13b-4b8dc5438822/wheel_x4.stl

3 walls, inner outer inner wall ordering, 0.08 layer height.

QingZhangBambu:
The wall seq had been adjusted according to the shortest path, so the corner cases could not be avoided. for example, the initial wall order is 0111122, after sorting ( shortest path) -> 0121121, then adjust sandwich order->2011121

QingZhangBambu:
I tend to merge your commits about "change the order of sandwich wall seq". 
And I open a param to fix single wall quality in develop mode.

So this PR could not be merged directly. 
Do you prefer to push a PR only containing commits about "change the order of sandwich wall seq", or I cherry-pick cornered commits to our git repo ( the author and msg would be kept, you are still the author).

igiannakas:
> I tend to merge your commits about "change the order of sandwich wall seq". And I open a param to fix single wall quality in develop mode.
> 
> So this PR could not be merged directly. Do you prefer to push a PR only containing commits about "change the order of sandwich wall seq", or I cherry-pick concerned commits to our git repo ( the author and msg would be kept, you are still the author).

Up to you whatever is easiest for you! 

I’d love to see your take on the code as I’m very much interested in getting this mode to work even better! It offers a big improvement in quality.

QingZhangBambu:
@igiannakas thanks!

QingZhangBambu:
your code could be been seen in the next public release. Thanks for your contribution

igiannakas:
Your welcome! Looking forward to using it!

PR #2207, opened 2023-08-07
Fix typo during flow rate calibration failure

twangodev:
Change "droped" to dropped for flow rate calibration fail.

Discovered this typo during my flow rate calibration for my X1C


lanewei120:
thanks for doing this

PR #2333, opened 2023-08-18
Add Klipper/Moonraker compatible metadata and thumbnails to GCode for Non-Bambu Printers

slynn1324:
This is the biggest thing that I'm missing from using Bambu Studio for all of my printers.

Uncommented the Thumbnail generation block and conditionalized for only non-Bambu printers, and increased the thumbnail size from 50x50 to 300x300.

Added Moonraker compatible stats/config to the bottom of the GCode export, only for non-Bambu printers.   Moonraker searches the "footer" (last read-size block) for most of the values - and has particular pattern expressions.   I also considered patching Moonraker, but found that several of the settings (such as filament used stats) were not available anywhere in the GCode output, so a change was required here anyway.

Please adjust or expand as you see fit.  

Thanks!

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -1563,9 +1563,12 @@ void GCode::_do_export(Print& print, GCodeOutputStream &file, ThumbnailsGenerato
     }
 
     //BBS: add plate id into thumbnail render logic
-    //DoExport::export_thumbnails_to_file(thumbnail_cb, print.get_plate_index(), THUMBNAIL_SIZE,
-    //    [&file](const char* sz) { file.write(sz); },
-    //    [&print]() { print.throw_if_canceled(); });
+    //SL 
+    if ( ! print.is_BBL_Printer() ){
+        DoExport::export_thumbnails_to_file(thumbnail_cb, print.get_plate_index(), THUMBNAIL_SIZE,
+            [&file](const char* sz) { file.write(sz); },
+            [&print]() { print.throw_if_canceled(); });
+    }
 
 
     // Write some terse information on the slicing parameters.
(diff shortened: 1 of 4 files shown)


lanewei120:
Hello @slynn1324 

we will check this logic
and what does this commit do?
image

slynn1324:
> Hello @slynn1324
> 
> we will check this logic and what does this commit do? image

apologies… didn’t mean for that to pull into the PR.  That’s just an adjustment of the build I used to run only the podman build action, I’ll see if I can back that out of this branch.

XunZhangBambu:
@slynn1324 . Thanks for your PR!  I prefer to not  add Moonraker compatible stats/config . I have recently made modifications to the thumbnail . Do you mind if I close this PR and acknowledge your contributions in the commit submitted internally ？

slynn1324:
For the metadata - I understand as much of it is redundant with the existing information.  I'm good with going back to Moonraker and making adjustments to how it parses for metadata from BambuStudio.  However; it does not appear that the current gcode output includes "filament used [mm]" or "filament used [g]" in any format, and they would be immensely useful in populating the Moonraker/Mainsail interface.   Do you have a different suggestion for where/how to include those?  Maybe those two filament used values in the Header block instead?  

I don't see what you've done with the Thumbnail - that must not be submitted up yet - so interested to see what you've done there. 

As far as closing this PR and acknowledgement on your commit instead -- no problem.

XunZhangBambu:
Thanks. We will support thumnail in gcode for 3rd printers along with the ability to edit the thumnail size in next version. If you need 'filament used', i prefer to add it in Header block

Markus-1984:
@slynn1324 
Have you found a solution for the meta data import in Mainsail/Moonraker? I´ve got the same problem and found your FR here.

kens96:
I have been using Orca waiting for this to come on the next release of Bambu Studio, disappointed to have just downloaded 1.8.2.56 and see it still hasn't.

XunZhangBambu:
@kens96  Sorry, but we have added  thumbnails in GCode for Non-Bambu Printers. And you can change thumbnail size in develop mode. Try again?
image

kens96:
@XunZhangBambu I didn't realise it is only visible in Developer mode, I have enabled and can see it now!
Thank you

PR #2338, opened 2023-08-19
de translation

Staubgeborener:
First i only wanted to translate this `msgstr "smooth"` part but i saw, that there are even more missing translations (i.e. for the calibration stuff), so this will be a long road. But this could already been pulled.


Ibot-11:
Hey,
Thanks for your help. I’m currently doing the German translations for Bambu Lab. Sometimes the translations are one step behind. The next update should include all new translations.
If you find any faulty translations feel free to contact me via discord: ibot_11

PR #2350, opened 2023-08-21
Fix Ankermake ooze issue

just-trey:
- `before_layer_change_gcode` should not have been there. It made the layers pause and ooze.
- Cleaned up all the custom gcode to simplify
- reverted resolution as the change provided no real benefit


just-trey:
I am not sure if this is where you want this pr to point to, but the other changes are not in the master branch yet.

Issue #1707, opened 2023-04-29
Wrong encoding - cyrillic


AndylgTom:
Wrong encoding is set in many places - cyrillic letters are displayed as ?????? (some of them are on screenshots).
In the initial versions of prusa slicer there was the same problem, but now it has been fixed a long time ago.
Please fix this in BambuStudio.

![Screenshot 2023-04-14 003217](https://user-images.githubusercontent.com/20997012/235307604-c3f30ac9-133d-42de-8eb4-9d7031b1f524.png)
![Screenshot 2023-04-14 003542](https://user-images.githubusercontent.com/20997012/235307608-66a23d18-73a6-421d-8f48-718acb4c5092.png)
![Screenshot 2023-04-29 211519](https://user-images.githubusercontent.com/20997012/235307609-e21f9514-9b29-429a-a486-2c2157438042.jpg)

LiZ-Li-BBL:
How did you set the BambuStudio to cyrillic letters? I need to know the steps to reproduce this.

AndylgTom:
> How did you set the BambuStudio to cyrillic letters? I need to know the steps to reproduce this.

I just replace any language *.mo file (for example, sv, or whatever) with a Russian translation file, and then I choose this language in the program settings.

LiZ-Li-BBL:
We will handle it. can you help provide your Russian *.mo file to us?

AndylgTom:
> We will handle it. can you help provide your Russian *.mo file to us?

Okay, I just wanted to add it after you fix the encoding bug.
This is not the final translation  [BambuStudio.zip](https://github.com/bambulab/BambuStudio/files/11795942/BambuStudio.zip)

LiZ-Li-BBL:
We plan to add Russian to preferences settings in the future. You need to select the correct language from the preferences settings so that part of the text which appears as "?" will be loaded properly. 
Thank you for the translation files.

PR #2371, opened 2023-08-23
Normal russian translation for Bambu Studio v1.7.4

AndylgTom:
Please add a **normal** ready Russian translation for Bambu Studio v1.7.4 (after fixing text which appears as ??? https://github.com/bambulab/BambuStudio/issues/1707 . OrcaSlicer has already solved this problem.) .

This translation is based on the Prusa slicer translation (the translation of which I have been supporting for a long time).


AndylgTom:
Hi, when will the translation be added?

PR #2413, opened 2023-08-28
Add K1 max printer profile

vlotho:


XunZhangBambu:
Thanks,but i only see changes to the Creality.json in the commit .Changes to the actual printer,filament,process profiles are missing. Can you recommit with these changes included ?

XunZhangBambu (changes requested):
Please add the modifications in  machine,process,filament profiles

PR #2439, opened 2023-09-02
Add ModelIO support on macOS

merikesh:
This PR adds support for a range of 3D flle formats for import, making use of [ModelIO](https://developer.apple.com/documentation/modelio) on macOS:
This is a version of my [PR to PrusaSlicer](https://github.com/prusa3d/PrusaSlicer/pull/11029) that has been refactored for BambuStudio 

* [Universal Scene Description](https://openusd.org/release/index.html)
* [Alembic](https://www.alembic.io)
* [PLY](https://en.wikipedia.org/wiki/PLY_(file_format)#:~:text=PLY%20is%20a%20computer%20file,list%20of%20nominally%20flat%20polygons.)

## How does it work?

This change only affects macOS users, and since it uses a system framework, it introduces no extra library dependencies. ModelIO is supported since macOS 10.11 which matches the [minimum system requirements](https://help.prusa3d.com/article/minimum-system-requirements_305250) as well.

The way this PR works is that it uses ModelIO to convert the data to a temporary STL that is then imported by the existing STL importer. After import, it clears the temp file.

## Why would someone want this?

With this PR, I no longer need to use other software to convert from 3D formats that I need to deal with in my day to day job. I get a lot of assets in USD and Alembic formats, from content creators who use those for their workflows.
I could ask them to export OBJ/STL, but its extra work for them that they don't verify. I could do the conversion myself in Blender , but with this PR, it streamlines my workflow quite a bit.

I figure since the changes are so minimal, it would be something others find useful as well. I recognize that its not ideal that the functionality is only for macOS users, but I hope that it can be seen as a value add for them instead. USD files are very common for Augmented Reality, and both USD and Alembic are common for film assets as well.

I'm a huge fan of BambuStudio and the changes in this PR make the workflow into it really seamless

### DIFF
--- a/src/libslic3r/CMakeLists.txt
+++ b/src/libslic3r/CMakeLists.txt
@@ -390,6 +390,8 @@ set(lisbslic3r_sources
 if (APPLE)
     list(APPEND lisbslic3r_sources
 			MacUtils.mm
+            Format/ModelIO.hpp
+            Format/ModelIO.mm
         )
 endif ()
 
@@ -504,6 +506,12 @@ if(NOT WIN32)
         target_link_libraries(libslic3r fontconfig)
     endif()
 endif()
+
+if (APPLE)
+    find_library(FOUNDATION Foundation REQUIRED)
+    find_library(MODELIO ModelIO REQUIRED)
+    target_link_libraries(libslic3r ${FOUNDATION} ${MODELIO})
+endif ()
 
 if (TARGET OpenVDB::openvdb)
     target_link_libraries(libslic3r OpenVDB::openvdb)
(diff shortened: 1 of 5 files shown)


Noisyfox:
Are you interested in porting your PR to OrcaSlicer too? Much appreciated.

merikesh:
I’m not very familiar with OrcaSlicer, but I’d be open to doing that.
However I think I should wait till I can merge it into BambuStudio first so
that I don’t need to juggle two forks and two sets of possible notes at the
same time (in addition to the Prusa PR which I haven’t gotten any response
on, but now I use a Bambu X1C anyway so I made this PR)

Noisyfox:
Awesome!
BTW it should be pretty straight forward as Orca is a fork of BambuStudio with extra features.

bambu123 (approved):

lanewei120:
great job @merikesh 
thanks for doing this~

merikesh:
Thanks, and thank you for the quick approval.

PR #2771, opened 2023-10-21
Fix name of bambu lab a1 mini cover

astrolemonade:
This fixes(happens on systems where the name of the files are case sensitive: Linux):
image
I had to `git mv -f "Bambu LAb A1 mini_cover.png" "Bambu Lab A1 mini_cover.png"` since git doesn't care about casing.


lanewei120:
thanks for doing this

PR #2848, opened 2023-11-05
Allow line width of nozzle diameter * 2.5

GuzziRaz:
As it were, 1 mm would be disallowed but 0.99 would be allowed for 0.4 nozzle.  1 mm is the sane maximum and 0.99 is unnecessary tedious to write.

OrcaSlicer already has this as `>` (it also allows 5x nozzle size but I personally see no use for that considering the nozzle flat size of Bambu hotends).


SaltWei:
@lanewei120  +1 for this patch

lanewei120:
thanks for doing this

PR #2868, opened 2023-11-11
BambuLab machine profile cleanup sync

adripo:
Continue from #2112 that created some confusion after rebase.
I am cherry picking the original commits and updating with latest modifications.


adripo:
@QingZhangBambu @tangketan 

I finally finished the refactor of the BBL machine profile. It was an intricate task that demanded careful consideration of each setting. Please review and share any thoughts or suggestions!

Changes Made:

- Completed the comprehensive refactor of the BBL machine profile.
- Conducted a meticulous analysis of individual settings to discern commonalities and distinctions.
- Organized models into a dedicated folder for improved structure.
- Eliminated redundant files containing duplicated settings.
- Introduced a universal configuration for each machine, streamlining maintenance for nozzle-related settings with simplified configurations.
- Implemented an `.editorconfig` to standardize indentation across all configurations.

Future Steps:

- Perform a thorough analysis of gcode to optimize duplication across various nozzle sizes.
- Scrutinize process and filament configurations for further refinement.

XunZhangBambu:
@adripo  Awesome work！Thanks for your hard work and contribution. I notice that you modify the "common" profiles ,and this will affect almost all presets in BBL due to the inheritance mechanism .I have written a python  script  and later i  will check the differences between before and after modification .  And for the options order , we already have an internal sorting rule  and may conflict with your rules .

XunZhangBambu:
@adripo  For the adjustment of the preset hierarchy structure , we will make comprehensive adjustments in next version , and your commit will be taken into consideration . Do you mind if i donnot merge the PR  instantly and acknowledege your contributions in later commit ?

adripo:
> I notice that you modify the "common" profiles ,and this will affect almost all presets in BBL due to the inheritance mechanism 

@XunZhangBambu  That is correct, I removed all the unnecessary and always overwritten configurations. Now the common is "clean" and only contains the options that are shared between almost all the profiles. I created the Machine common to put all the options that are shared by all the nozzles of the same machine and tried to keep the nozzle profiles as similar as possible for easier maintenance.
If you check the diff between the final configuration (after all the inheritances) for each nozzle before and after my PR they should be identical (except for some filament reference that had duplicated configs).

The options order is not important, you can change it following the rules of the script.

adripo:
> @adripo For the adjustment of the preset hierarchy structure , we will make comprehensive adjustments in next version , and your commit will be taken into consideration . Do you mind if i donnot merge the PR instantly and acknowledege your contributions in later commit ?

Sure, not an issue on my side, just please keep them simpler, similar to my solution. I also found some human errors introduced because of poor maintainability of the current structure. This can be easily avoid when the nozzle files are more similar between them.

For example here the 0.8 profile references a 0.6 machine:
https://github.com/bambulab/BambuStudio/blob/master/resources/profiles/BBL/machine/Bambu%20Lab%20P1S%200.8%20nozzle.json#L31



 Also file naming doesn't follow any common rule
 
image

1) references to `common layer 1` and it is inherited only by `common layer 2` -> what is the advantage?
2) references to `common layer 2` and it is inherited by all the nozzle configs. Name is uncommon `3dp_001` -> could there be a 002? When?
3) references to a machine model and it has nothing in common with the machine configurations present in that directory
4) the actual nozzle configuration that overwrites lots of identical options from the common

Those are only some of the issues I identified and tried to solve with my PR.

Feel free to adapt it for a better solution or use it as a reference, but it would be a big advantage to have a simple and maintainable structure without duplicated options.
Also, another important point would be to create uniform start_gcode, layer_change, etc. for each machine (similar to what has been done to A1 mini) without many customizations per nozzle, and using nozzle size as a variable inside formulas when necessary. In the current state if a modification is necessary for the gcode of one machine it needs to be applied 4 times, one for each nozzle, maintaining the differences between them. As you can imagine this is time consuming and prone to human errors.

lanewei120:
hello @adripo 

I am sorry we are on the Spring Festival holidays,
and we will check it after we return to work

adripo:
well... okay...

PR #2879, opened 2023-11-13
uncomment 'filament used' lines in gcode generator

eropple:
I'm looking to build third-party tools that are compatible with Bambu Studio. One of the metrics that I'd like to track (and that other tools already do, e.g. Moonraker) is the amount of filament used. Both PrusaSlicer and OrcaSlicer emit this in the exact same way, but Bambu Studio comments it out; it'd be nice to revert this.

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -1468,10 +1468,10 @@ namespace DoExport {
 	    print_statistics.clear();
         print_statistics.total_toolchanges = std::max(0, wipe_tower_data.number_of_toolchanges);
 	    if (! extruders.empty()) {
-	        //std::pair<std::string, unsigned int> out_filament_used_mm ("; filament used [mm] = ", 0);
-	        //std::pair<std::string, unsigned int> out_filament_used_cm3("; filament used [cm3] = ", 0);
-	        //std::pair<std::string, unsigned int> out_filament_used_g  ("; filament used [g] = ", 0);
-	        //std::pair<std::string, unsigned int> out_filament_cost    ("; filament cost = ", 0);
+	        std::pair<std::string, unsigned int> out_filament_used_mm ("; filament used [mm] = ", 0);
+	        std::pair<std::string, unsigned int> out_filament_used_cm3("; filament used [cm3] = ", 0);
+	        std::pair<std::string, unsigned int> out_filament_used_g  ("; filament used [g] = ", 0);
+	        std::pair<std::string, unsigned int> out_filament_cost    ("; filament cost = ", 0);
 	        for (const Extruder &extruder : extruders) {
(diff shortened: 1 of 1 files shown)


THE-SIMPLE-MARK:
Please merge this PR! I'm having the exact same issue and I'm forced to use Prusa Slicer for the time being. (#3072)

THE-SIMPLE-MARK:
@XunZhangBambu Do we have any news regarding this?

THE-SIMPLE-MARK:
@eropple I've tested this implementation, however it does not work as expected. It seems that you only modified the old function which prints the stats to the console, however it never gets called because [that's commented out as well](https://github.com/bambulab/BambuStudio/blob/9d1cdead44ccb5894cb45316d2fdb6a5839e0fff/src/libslic3r/GCode.cpp#L2314-L2318).

What I'd suggest is to move the code you uncommented into [the part which writes the header block](https://github.com/bambulab/BambuStudio/blob/9d1cdead44ccb5894cb45316d2fdb6a5839e0fff/src/libslic3r/GCode.cpp#L1664-L1711) in the gcode. Obviously you would need to refactor the code a little to match the pattern being used for writing it to the file in the new function.

eropple:
I have a patched version that does work, yeah - will try to remember to
update my fork this week.

XunZhangBambu:
@eropple  Hello, uncommenting this code indeed writes some filament usage information in the G-code, but these details are not accurate. Accurate filament usage information can only be obtained after post-processing (due to custom G-code). @THE-SIMPLE-MARK offers a solution: use a placeholder here first and then write the correct data back during post-processing.

XunZhangBambu:
> > @eropple Hello, uncommenting this code indeed writes some filament usage information in the G-code, but these details are not accurate. Accurate filament usage information can only be obtained after post-processing (due to custom G-code). @THE-SIMPLE-MARK offers a solution: use a placeholder here first and then write the correct data back during post-processing.
> 
> Yeah that makes sense. Can you tell us where we need to place the code so it gets run in post-processing?

A simple way is to move the code block to the end of the gcode file. Add some placeholders to represent filament usage and 
deal with them in "process_gcode_line" function. The accurate usage information can be obtained in GCodeProcessor::UsedFilaments

THE-SIMPLE-MARK:
> > > @eropple Hello, uncommenting this code indeed writes some filament usage information in the G-code, but these details are not accurate. Accurate filament usage information can only be obtained after post-processing (due to custom G-code). @THE-SIMPLE-MARK offers a solution: use a placeholder here first and then write the correct data back during post-processing.
> > 
> > 
> > Yeah that makes sense. Can you tell us where we need to place the code so it gets run in post-processing?
> 
> A simple way is to move the code block to the end of the gcode file. Add some placeholders to represent filament usage and deal with them in "process_gcode_line" function. The accurate usage information can be obtained in GCodeProcessor::UsedFilaments

Alright. I will probably create a new PR and do this during this day to accelerate the process.

XunZhangBambu:
solved in #3090

PR #2949, opened 2023-11-24
unix: Use canonical mime-type for 3MF files

hadess:
model/3mf is the canonical mime-type, but keep the application/vnd.ms-3mfdocument alias.

See for details:
https://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/179


iZonex (approved):
Cool! Thanks

lanewei120:
thanks for doing this

PR #2960, opened 2023-11-25
deps: Update wxWidgets build options

hadess:


bambu123 (commented):
<html>
<body>

wxUSE_LIBMSPACK | Use libmspack library.
-- | --



</body>
</html>

hadess:
> wxUSE_XPM | Enable XPM reader for wxImage and wxBitmap classes. -- | -- wxUSE_XPM Enable XPM reader for [wxImage](https://docs.wxwidgets.org/latest/classwx_image.html) and [wxBitmap](https://docs.wxwidgets.org/latest/classwx_bitmap.html) classes.

`wxUSE_XPM` isn't the same as `wxUSE_LIBXPM`.

> wxUSE_LIBMSPACK | Use libmspack library. -- | --

I don't understand that comment.

BambuStudio seems to expect wxWidgets to be build with this option, but doesn't force it on when building wxWidgets. So one could potentially build all the way to BambuStudio and not be able to finish building because wxWidgets wasn't build with mspack support enabled.

If libmspack isn't actually needed (it looks like it actually isn't...), I can remove that commit from the PR, and remove it from the required build packages instead.

lanewei120:
why remove this one ”-DCMAKE_DEBUG_POSTFIX:STRING=“

hadess:
> why remove this one ”-DCMAKE_DEBUG_POSTFIX:STRING=“

Because nothing uses it, as mentioned in the cmake warning output that's in the commit message.

hadess:
> If libmspack isn't actually needed (it looks like it actually isn't...), I can remove that commit from the PR, and remove it from the required build packages instead.

I've disabled the mspack requirement now.

lanewei120:
CMAKE_DEBUG_POSTFIX is used on other platform, such as windows
there is no warnings on windows
image

hadess:
> CMAKE_DEBUG_POSTFIX is used on other platform, such as windows

OK, I've removed that bit.

lanewei120:
thanks for doing this

PR #3093, opened 2023-12-12
Fix printing times on same line in GCode file 

THE-SIMPLE-MARK:
Fix the `model printing time` and `total estimated time` statistics being on the same line in the `HEADER_BLOCK` in generated GCode files.

While this may seem simple, it can cause huge headaches when parsing programatically.


lanewei120:
just found this modification is already in out internal commit history
and included in 1.8.2

thanks for doing this

PR #3096, opened 2023-12-13
Fix a number of compilation problems

hadess:
Using `gcc version 13.2.0 (GCC)` in a Flatpak sandbox, I ran into a number of problems.

This isn't all the problems I've run into, but this is all the ones I can fix right now.


lanewei120:
thanks for doing this

PR #2516, opened 2023-09-16
Add retraction on filament change (Fixes #2506)

mtorromeo:
This is based on [experiments done by @LeonFisherSk_1256980](https://www.printables.com/model/582382-bambulab-profile-for-up-to-60-purge-reduction).

Hopefully this MR helps to make it clearer what are the profile changes that were applied in the experiment. This is not an exact copy-paste since I changed the comments (originally in german) and used `F{old_filament_e_feedrate}` to push back the filament and a retraction speed consistent with what was used elsewhere in the profile.

Fixes #2506

PR #1660, opened 2023-04-23
Improve filament gcode change speed (x2)

DzzD:
Here is a major performance improvment on filament gcode change :

1. changed pulsatile purge : bumping purging feedrate and using constant time pause between pulse not depending on purged volume
     => pulse can go higher than max filament volumetric speed as they are pulsatile and the nozzle have no resistance at its output.

2. removed unrequired moves and speed up other moves

3. removed wait of temperature as they are unecessary

Based on this thread about improving multi-color printing speed : 
https://forum.bambulab.com/t/ideas-on-boosting-speed-for-multi-color-printing-with-ams/10037/41?u=dzzd

PR #3097, opened 2023-12-13
Save the Environment by reducing purge through retracting filament during every Filament change! Less Poop! No lower reliability! Very simple change!

LeonFisherSkipper:
### **TLDR: This change would save huge amounts of waste with no reduction in print quality or reliability. No drawbacks at all. Thousands of people have tested and confirmed that it works. For details look in printables link.**

**This makes the Bambu printers much more competitive against the Prusa XL in multi color printing.** 

**Please for mother Nature implement these changes to safe the environment from unnecessary waste!**

**Comparison Before (Left) and After (right) the changes** :

IMG-4320

**Details:**

**I posted a [printables profile](https://www.printables.com/model/582382-bambulab-profile-for-up-to-60-purge-reduction) which reduces the purge volumes** without negativ effects in september and since then thousands of people have tested it without issues. Many people on [reddit](https://www.reddit.com/r/BambuLab/comments/18c5sqh/comment/kc9gt03/?utm_source=share&utm_medium=web2x&context=3) recommend this. Also In a video by [Teaching Tech](https://youtu.be/p3ZIM3megIU?si=PhgUzG7-S1dT1NH6) the top comment recommends this change to safe filament. 

I have personally tested this with problematic filament combinations like PVA and Nylon and even than it works flawlessly on thousands of filament changes. 

DzzD also included this change in a recent update to his [Pull request](https://github.com/bambulab/BambuStudio/pull/1660), 
but I think his other changes require more testing to make sure that the reliability stays high and blops don't fall onto the bed etc.

**I believe my changes should be added in the mean time before the filament change is speed up like DzzD suggested, as my changes dont require testing and are incredibly easy to implement.** 

The only changes required are 8 lines of G-code in the Nozzle profiles and optionally that in the flushing-volume calculator a flat 50 mm3 is subtracted from every combination. So that instead of the default 280, the new default would be 230. The same for auto-calculated values, where if the autocalculated value was 467 for example until now, the new value would be 467-50=417.

A other user already opened a [pull request](https://github.com/bambulab/BambuStudio/pull/2516), but I wanted to post it myself, as I am the author of the original Printables Post and I can answer questions here and hope to get the changes implemented.


iZonex:
@LeonFisherSkipper That pull request not for master branch? As I see 946 commits, that PR never will be accepted at all. Please only leave what must be changed with 1 commit.

LeonFisherSkipper:
@iZonex I linked the wrong branch. I marked the location where the changes would go now.

PR #3100, opened 2023-12-13
Save the Environment by reducing purge through retracting filament during every Filament change! Less Poop! No lower reliability! Very simple change!

LeonFisherSkipper:
**TLDR: This change would save huge amounts of waste with no reduction in print quality or reliability. No drawbacks at all. hundreds of people have tested and confirmed that it works. For details look in the printables link.
This makes the Bambu printers much more competitive against the Prusa XL in multi color printing.**

**Please for mother Nature implement these changes to safe the environment from unnecessary waste!**

**Comparison Before (Left) and After (Right) the changes :** -> Color changes transition much earlier.
IMG-4320

Details:

**I posted a [printables profile](https://www.printables.com/model/582382-bambulab-profile-for-up-to-60-purge-reduction) which reduces the purge volumes** without negativ effects in september and since then thousands of people have tested it without issues. Many people on [reddit](https://www.reddit.com/r/BambuLab/comments/18c5sqh/comment/kc9gt03/?utm_source=share&utm_medium=web2x&context=3) recommend this. Also In a video by [Teaching Tech](https://youtu.be/p3ZIM3megIU?si=PhgUzG7-S1dT1NH6) the top comment recommends this change to safe filament.

I have personally tested this with problematic filament combinations like PVA and Nylon and even than it works flawlessly on thousands of filament changes.

DzzD also included this change in a recent update to his https://github.com/bambulab/BambuStudio/pull/1660,
but I think his other changes require more testing to make sure that the reliability stays high and blops don't fall onto the bed etc.

**I believe my changes should be added in the mean time before the filament change is speed up like DzzD suggested, as my changes dont require testing and are incredibly easy to implement.**

The only changes required are 8 lines of G-code in the Nozzle profiles and optionally that in the flushing-volume calculator a flat 50 mm3 is subtracted from every combination. So that instead of the default 280, the new default would be 230. The same for auto-calculated values, where if the autocalculated value was 467 for example until now, the new value would be 467-50=417.

A other user already opened a https://github.com/bambulab/BambuStudio/pull/2516, but I wanted to post it myself, as I am the author of the original Printables Post and I can answer questions here and hope to get the changes implemented.

The changes in github I made are 20 times the same line of code for each nozzle profile of all the printers. This line extracts a extra 20mm of filament at the beginning to compensate when the filament is pushed back out at the end.

Than a couple of lines are changed in the fdm_bbl_3dp_001_common which deals with the filament changes and pushes back the filament at the end.

I haven't adapted the flushing volumes yet on github.


LeonFisherSkipper:
It may look like a lot of changes, but it only is 8 lines of code in total. 1 of these code lines is added to each of the 20 nozzle profiles for all printers, so a total of 21 files were changed.

THE-SIMPLE-MARK:
I +1 this. Perhaps we could further enhance this and make the printers remember their previous filament and not purge if the previous filament that was used is the same.

LeonFisherSkipper:
I updated the code to change the flush volume matrix. Every auto-calculated value gets 50 mm3 subtracted now at the end. I have problems compiling the code, so It would be great if someone could test it for me.

iZonex:
Look awesome, when we can start use this feature?

igiannakas:
To do this "correctly" it would need to be a text box to type in the flush volume to be subtracted as your gcode retracts 20mm but for example I'm using 15mm. Ideally this value should be entered in mm by the user, with a min of 0 and max of 20mm and exported to a gcode variable which is then templated into the filament layer change code so the filament retraction before the cut can be adjusted from the UI instead of the g-code. The mm value can be converted to a volume value using a simple formula (mm3 = mm * 2.4)

I dont think Bambu will merge this PR as is, however I hope they consider the above proposals for the future. I may take a stab at implementing this in Orca at some point.

igiannakas:
> I updated the code to change the flush volume matrix. Every auto-calculated value gets 50 mm3 subtracted now at the end. I have problems compiling the code, so It would be great if someone could test it for me.

This only happens with a 1x multiplier as you're modifying the base matrix before the multiplication is applied. Ideally you'll need another config variable and to be included in the calculations after the multiplier is applied.

ankurpatel82:
> To do this "correctly" it would need to be a text box to type in the flush volume to be subtracted as your gcode retracts 20mm but for example I'm using 15mm. Ideally this value should be entered in mm by the user, with a min of 0 and max of 20mm and exported to a gcode variable which is then templated into the filament layer change code so the filament retraction before the cut can be adjusted from the UI instead of the g-code. The mm value can be converted to a volume value using a simple formula (mm3 = mm * 2.4)
> 
> I dont think Bambu will merge this PR as is, however I hope they consider the above proposals for the future. I may take a stab at implementing this in Orca at some point.

I agree - unlikely Bambu Labs merges this - but maybe SoftFever will - testing there may nudge Bambu to merge.

LeonFisherSkipper:
@igiannakas Thanks for your input. I think you are right that Bambu cant merge this as is. I like your idea, but I have tested 20mm retractions on thousands of filament changes with many different materials. It worked perfectly on all of them for me. Bambulab would need to test this themselves in their test farms. It would be really great if you could try to implement it in Orca!

SaltWei:
We are do printing test for this PR. Thank you very much!
Will update the test result here when it finished

SaltWei:
We conducted extensive testing for nearly half a month from December 19, 2023 to January 5th, 2024, on various models of Bambu printers in our testing farm. This modification indeed reduces material waste during filament changes. Impressive!
But we also find some headache issues. 

A1mini:
Nozzle type: 0.4mm
Test materials: PETG / PLA SILK
Issue:
1. Frequent occurrence of incomplete cutting of PLA SILK filament.

X1C:
Nozzle type: 0.2mm / 0.4mm / 0.6mm / 0.8mm
Test materials: PETG / PLA SILK
Issues:
1. Frequent occurrence of filament deformation under extruder pressure.
2. Poor quality of wipe tower with 0.6mm+0.8mm nozzles.
3. Poor surface quality in the Y direction of the models 0.6mm+0.8mm nozzles.

P1P / P1S:
Nozzle type: 0.2mm / 0.4mm / 0.6mm / 0.8mm
Test materials: PETG / PLA SILK
Issues:
1. Poor quality of wipe tower with 0.6mm+0.8mm nozzles.
2. Poor surface quality in the Y direction of the models 0.6mm+0.8mm nozzles.

Example pictures of issues:
Incomplete cutting of filament.
109b5d6a-274c-47de-a830-08f53e7b4964
filament deformation under extruder pressure
730321a8-c27c-4c12-a906-6dc64fc0eec7
Poor surface quality in the Y direction
03bac159-0853-4e88-82fd-3e73dc04e9f6
Poor quality of wipe tower
4365f789-9e13-4b62-b66e-6f9c8250893e

SaltWei:
So if we only look at 0.4 nozzle, issues include:
Frequent occurrence of incomplete cutting of PLA SILK filament

leonFisher-Skipper:
@SaltWei thanks a lot for the extensive insight! 
This issues seem to all only happen for PLA silk. Did no issues occur with other materials? Just to be sure: did you only use 20mm of retraction? 

I think the problem could be fixed if the retraction was lowered to 10mm or even no retraction at all for Silk PLA.

LeonFisherSkipper:
@SaltWei thanks a lot for the extensive insight!
This issues seem to all only happen for PLA silk. Did no issues occur with other materials? Just to be sure: did you only use 20mm of retraction? With more in my testing normal PLA also was problematic.

I think the problem could be fixed if the retraction was lowered to 10mm or even no retraction at all for Silk PLA.

mzanetti:
> So if we only look at 0.4 nozzle, issues include:
> 
>     1. Frequent occurrence of incomplete cutting of PLA SILK filament
> 
>     2. Frequent occurrence of filament deformation under extruder pressure, which cause the filament to be stuck when unload or slip and no extrusion easily when load the filament again

Kinda sounds like the amount of pulling back the filament before cutting should be provided by the filament profile, so it could be reduced (or even disabled) for Silk filaments.

GuzziRaz:
I have also been running this for almost half a year with barely any problem. I did see the occasional cut/pull-back problem with PLA Silk, but only rarely.

> Kinda sounds like the amount of pulling back the filament before cutting should be provided by the filament profile, so it could be reduced (or even disabled) for Silk filaments.

Definitely. The Support W filament (and the likes) would need to be considered as well.

Also, if Bambu remains cautious, the whole thing could be an “advanced” setting, defaulting to disabled (as a checkbox - so eg. PLA Basic profile would have it as 20 mm and Silk at just 10 or maybe zero, but that global checkbox _could_ default to off. The user could then just enable it if they want to, without needing to edit dozens of filament profiles).

mduft:
I did test this for a few hundred swaps on my P1S, and for the most part all went well. I had an issue with an extruder clog with the silk filament. Now I disabled the profile while printing with silk filament - HOWEVER I had the exact same extruder clog *again* even without the profile. Also the clogging piece of filament looked exactly the same, just a little longer and it was easier to remove without actually opening the extruder assembly. 

Whats strange anyhow is that the clogging piece in both scenarios - with and without profile - was seemingly properly cut on the one side, and molten on the other side, so I actually *expected* that piece to be in there... I'm confused. I checked my blade, and that seems to be well sharp enough (I have the printer for only 2 weeks or so :D).

LeonFisherSkipper:
@saltwei could we please get another update? Will this be implemented? If yes when could this potentially happen? Was more work done? Thanks a lot for your effort! :)

XunZhangBambu:
@LeonFisherSkipper  We have added it as an experimental feature and it's only applicable to X&P series currently. Thanks for you effort！ We have submitted the code internally on your behalf, so this PR may not be merged.

mduft:
Is this in the current stable Bambu somehow? Where can I enable it?

PR #3139, opened 2023-12-16
More compilation fixes

hadess:
Follow-up from https://github.com/bambulab/BambuStudio/pull/3096

### DIFF
--- a/src/libslic3r/Arrange.cpp
+++ b/src/libslic3r/Arrange.cpp
@@ -17,6 +17,7 @@
 #define BOOST_NO_CXX17_HDR_STRING_VIEW
 #endif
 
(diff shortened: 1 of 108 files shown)


hadess:
Happy to split this up in smaller merge requests if it helps landing them.

lanewei120:
@hadess 

thanks very much for doing this,
could you merge thoses commit into one?

too much commits will cause the history overly lengthy

hadess:
> thanks very much for doing this,
> could you merge thoses commit into one?

Done. It's now 4 separate commits. I wanted to make sure that the biggest ones that didn't just require a new `include` get their own commits to explain.

PR #3172, opened 2023-12-19
Update PrintConfig.cpp to fix typo

WowSuchRicky:
Super minor fix. Same typo is in the Bambu (X1/C at least) firmware as well, but I do not see that as open source.


lanewei120:
thanks for doing this

PR #3201, opened 2023-12-22
fix: i18n french translation

dorianboulch:
In French, we only use the verb "jouer" (jouant) to play with toys or with somebody for example.
We can replace it with "En cours...", that means "In progress"


StoneLiBambu:
Thank you very much. Due to the fact that the .po source data for this file is maintained on our internal platform's database (for more details, please refer to https://wiki.bambulab.com/en/bambu-studio/contribute-languages), we are unable to directly merge this patch. However, we will submit this patch to our internal database and acknowledge it in release note.


Stone Li （旺仔）

lanewei120:
already included before

Issue #3175, opened 2023-12-20
0.8 nozzle profile mixed in with 0.6 presets


B-RE-B:
### Bambu Studio Version

1.8.2.56

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11 home

### Printer

P1S

### How to reproduce

I had only the P1S 0.4 nozzle preset enabled under Printer selection.  What caused me to find the issue was adding the P1S 0.6 nozzle...
When I added the 0.6 nozzle settings, one of the imported layer heights is actually named for a 0.8mm nozzle.  I'm unsure if its named wrong or if it's in the wrong place.
Also, are any of the 0.6 settings supposed to be for draft or fine settings ... they're all labelled "standard" except for one labelled for "strength"  This part maybe should have gone under a feature request instead of bug.

![Screenshot 2023-12-19 201721](https://github.com/bambulab/BambuStudio/assets/154389986/82e2a9fb-d72e-4e3b-a449-628fe594a441)
![Screenshot 2023-12-19 201743](https://github.com/bambulab/BambuStudio/assets/154389986/430a9194-1d68-4df9-866b-264bb5201b95)

### Actual results

One of the 0.6 nozzle presets for the P1S is named wrong or in the wrong place - says it's a 0.8 profile.

![Screenshot 2023-12-19 201832](https://github.com/bambulab/BambuStudio/assets/154389986/b917e07e-765e-4d98-aa25-87109cc4c92d)

### Expected results

The 0.6 nozzle profile for the P1S should only have layer height settings for that size nozzle, not a 0.8 preset.

### Project file & Debug log uploads

Not sure if this blank project file helps.  I hadn't done anything but add the 0.6 nozzle size, when I saw the error.

[TEST - Error report.zip](https://github.com/bambulab/BambuStudio/files/13721862/TEST.-.Error.report.zip)

### Checklist of files to include

- [X] Log file
- [X] Project file

XunZhangBambu:
Oh, Thanks! You found! That's a mistake. I will fix it

PR #3259, opened 2023-12-30
Update 0.48mm Standard @BBL X1C 0.8 nozzle.json

SufficentMuffin:
Removed P1S 0.6mm nozzle from process profile.


hadess:
This looks like spam.

SufficentMuffin:
Please excuse that I gave no further explanation:

If you choose Bambu´s P1S with the 0.6 mm nozzle the Slicer will show the 0.48 mm layer height profile for it.
This should not be the case since the max layer height for 0.6 mm nozzle is 0.42 mm.
If you choose the 0.48mm profile and slice an object it will throw an "error" whichs says the layer height was changed back to 0.42mm.

XunZhangBambu:
Thanks. We have already fixed this in next version. Same problem as #3175

PR #3268, opened 2023-12-31
Fix "Beta Version" appearing as "Internal Version"

hadess:
The value of BBL_INTERNAL_TESTING is reversed and should be set to 1 when requested instead of 1 when not requested.

Fixes: 79480405a022 ("ENH: update version to 1.3.0")

Note: pretty sure I got this right, in my testing, without `-DBBL_RELEASE_TO_PUBLIC=1` in the build flags, it would default to an "internal testing" version instead of a "beta version".

### DIFF
--- a/version.inc
+++ b/version.inc
@@ -7,6 +7,8 @@ set(SLIC3R_APP_KEY "BambuStudio")
 if(NOT DEFINED BBL_RELEASE_TO_PUBLIC)
 set(BBL_RELEASE_TO_PUBLIC "0")
 endif()
+# Note that defining BBL_INTERNAL_TESTING will create a beta version
+# not an "internal testing" version (the default)
 if(NOT DEFINED BBL_INTERNAL_TESTING)
 set(BBL_INTERNAL_TESTING "1")
 endif()


bambu123 (commented):
Why?

bambu123:
Default is Internal version. Nothing to do with BBL_RELEASE_TO_PUBLIC

hadess:
> Why?

It would be great if you could tell me which part of commit message is unclear.

> Default is Internal version. Nothing to do with BBL_RELEASE_TO_PUBLIC

It doesn't have "Nothing to do with BBL_RELEASE_TO_PUBLIC". This is the code:
```
if(NOT DEFINED BBL_RELEASE_TO_PUBLIC)
set(BBL_RELEASE_TO_PUBLIC "0")
endif()
if(NOT DEFINED BBL_INTERNAL_TESTING)
set(BBL_INTERNAL_TESTING "1")
endif()
```

Setting `BBL_INTERNAL_TESTING` will make the app be a "beta version". I'm guessing setting `BBL_INTERNAL_TESTING` should make the app be in "internal testing" or the name of the flag should be changed to something else.

bambu123:
> The value of BBL_INTERNAL_TESTING is reversed and should be set to 1 when requested instead of 1 when not requested.
> 
> Fixes: [7948040](https://github.com/bambulab/BambuStudio/commit/79480405a022a63b6013ba82e913858802630e41) ("ENH: update version to 1.3.0")
> 
> Note: pretty sure I got this right, in my testing, without `-DBBL_RELEASE_TO_PUBLIC=1` in the build flags, it would default to an "internal testing" version instead of a "beta version".

BBL_INTERNAL_TESTING only affect some version identifiers. BBL_RELEASE_TO_PUBLIC remove debugging features that do not need to be published. The two are different, and the names are indeed somewhat inaccurate.

hadess:
> The two are different, and the names are indeed somewhat inaccurate.

What would you want me to rename `BBL_INTERNAL_TESTING` to? Is `BBL_BETA_VERSION` good for you?

bambu123:
> > The two are different, and the names are indeed somewhat inaccurate.
> 
> What would you want me to rename `BBL_INTERNAL_TESTING` to? Is `BBL_BETA_VERSION` good for you?

This is also related to our construction of CI, and it is recommended not to change it.

hadess:
> This is also related to our construction of CI, and it is recommended not to change it.

I've documented the fact that the naming is reversed, because it's really confusing...

lanewei120:
Hi @hadess 

I agree with you that this name is confusing
however, as it is already used in our internal patch/daily/release building system
I am afraid we can not change it:)

anyway, thanks for doing this~

hadess:
> however, as it is already used in our internal patch/daily/release building system
> I am afraid we can not change it:)

I didn't change it, I documented it.

lanewei120:
> > however, as it is already used in our internal patch/daily/release building system
> > I am afraid we can not change it:)
> 
> I didn't change it, I documented it.

ok, so it is ok, thanks for doing this

PR #3271, opened 2024-01-01
Fix garbled text when display scalling > 300%

### LINKED ISSUE #2626: Font issues info screen unreadable
### Bambu Studio Version

1.7.6.92

Bambu Lab Official website

### OS version

Windows 10 Version	10.0.19045 Build 19045

Core i5 gen 4 16GB RAM

### Printer

Creality Ender 3S1

### How to reproduce

Font unreadable see screenshot 
Bambu Studio
picture

### Actual results

Bambu Studio

### Expected results

Expect the font to be readable

[Lid5_short.zip](https://github.com/bambulab/BambuStudio/files/12739959/Lid5_short.zip)
[BambuStudio.zip](https://github.com/bambulab/BambuStudio/files/12739961/BambuStudio.zip)

- [X] Log file
- [X] Project file

lllucius:
On Windows, if your display scaling is greater than 300%, the text rendered by imgui will be garbled. Allowing imgui to do the scaling instead of making the font to large corrects the issue.

This fixes #2626

### DIFF
--- a/src/slic3r/GUI/GLCanvas3D.cpp
+++ b/src/slic3r/GUI/GLCanvas3D.cpp
@@ -6428,7 +6428,7 @@ void GLCanvas3D::_resize(unsigned int w, unsigned int h)
     m_last_w = w;
     m_last_h = h;
 
-    const float font_size = 1.5f * wxGetApp().em_unit();
+    const float font_size = 1.0f * wxGetApp().em_unit();
 #if ENABLE_RETINA_GL
     imgui->set_scaling(font_size, 1.0f, m_retina_helper->get_scale_factor());
 #else
--- a/src/slic3r/GUI/GLCanvas3D.cpp
+++ b/src/slic3r/GUI/GLCanvas3D.cpp
@@ -6428,7 +6428,17 @@ void GLCanvas3D::_resize(unsigned int w, unsigned int h)
     m_last_w = w;
     m_last_h = h;
 
-    const float font_size = 1.0f * wxGetApp().em_unit();
+    float font_size = wxGetApp().em_unit();
+
+#ifdef _WIN32
+    // On Windows, if manually scaled here, rendering issues can occur when the system's Display
+    // scaling is greater than 300% as the font's size gets to be to large. So, use imgui font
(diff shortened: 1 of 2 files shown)


LiZ-Li-BBL:
Just scale the font directly rather than loading the font with appropriate size, will result in the font texture being blurred and difficult to read. We think this part of modification should only be effective if the screen scaling >= 300% on Windows.

lllucius:
I could certainly do the conditional, but to my blind eyeball the text looks the same to me when using 150% scaling.  Keep in mind that I simply may not be able to see the difference, so I'd be happy to do the conditional if that's what y'all want.

When looking at the screenshots, it's the text handling by imgui that is being scaled here, not that of wxWidgets, so all the text on the right.

Here's a screenshot from an unmodified Orca Slicer 1.8.1 at 150% display scaling:

original 1 8 1

And this is from the modified 1.9.0:

Fixed 1 9 0

Incidentially, this is what it looks like in unmodified 1.8.1 at 350% display scaling:

original 1 8 1 at 350%

lllucius:
> Just scale the font directly rather than loading the font with appropriate size, will result in the font texture being blurred and difficult to read especially at a lower screen scaling. We think this modification should only take effect when the screen scaling >= 300% on Windows.

Just build Orca on Linux and I definitely see what you mean.  Will make it conditional as you say.

lllucius:
> Just scale the font directly rather than loading the font with appropriate size, will result in the font texture being blurred and difficult to read especially at a lower screen scaling. We think this modification should only take effect when the screen scaling >= 300% on Windows.

Pull request has been updated to include the changes you requested...only on Windows and only if the system's scaling is greater than 300%.

LiZ-Li-BBL (approved):

lllucius:
This is great news LiZ-Li-BBL!  Thaniks much!

hadess:
> It didn't apply only on Windows, it seems that it applied on non-macOS, which broke Linux: #3416

Looks like only one of the 2 commits in the PR landed in the stable v1.8.4 branch, so this is a process problem...

PR #3312, opened 2024-01-05
FIX: Fix build error with wxWidgets 3.2.4

hadess:
Make initializers explicit to avoid ambiguous wxArrayString overloads.

```
src/slic3r/GUI/PhysicalPrinterDialog.cpp: In member function ‘void Slic3r::GUI::PhysicalPrinterDialog::build_printhost_settings(Slic3r::GUI::ConfigOptionsGroup*)’: src/slic3r/GUI/PhysicalPrinterDialog.cpp:332: error: call of overloaded ‘set_values(<brace-enclosed initializer list>)’ is ambiguous
  332 |         choice->set_values({ m_config->opt_string("printhost_port") });
      |
```

See https://github.com/prusa3d/PrusaSlicer/pull/11769


lanewei120:
thanks for doing this

Issue #2805, opened 2023-10-30
Suggestion: Update to wxWidgets 3.2.3 to Fix webkit-gtk 4.x Dependency Issues on Linux


mwgabby-li:
I was having some trouble building this on Gentoo Linux, and after working through various issues, I finally determined the root cause.

Basically, wxWidgets 3.1.5 doesn't support webkit-gtk version 4.x, which is the only version available on my system. If you upgrade to 3.2.3, though, the latest version, wxWidgets includes support for this version of webkit-gtk. This solves the dependency issue and allows building on Gentoo and probably any other distros that use webkit-gtk 4.x.

I noticed two other issues when making this change:

- I had build errors related to the PNG library. I removed this dependency from the build system and had it use my system's version. This may or may not be related to the change I made.
- I had to comment out `#define NANOSVG_IMPLEMENTATION` and `#define NANOSVGRAST_IMPLEMENTATION` from `BitmapCache.cpp`, as the contents of this single-header library were apparently already being included from somewhere else. I did not investigate why this was necessary.

Locally, just to get it working, I also didn't bother updating your wxWidgets patch. Much to my surprise, the project still built and linked fine.
The program still seems to work fine with wxWidgets 3.2.3, but I don't really have a baseline to compare with, as I wasn't able to get it working before making these changes.

PR #3340, opened 2024-01-09
deps: Update to wxWidgets 3.2.4

hadess:
This update fixes a large number of bugs, including cutting down on the number of run-time critical messages on startup.

This update also contains the NanoSVG build control options added in version 3.1.7.


hadess:
I need to update `0001-wxWidget-fix.patch`.

hadess:
How can I verify that both wxwidgets and bambustudio were built with nanosvg support?

lanewei120:
as wxWidgets is a base module, which has lots of known issues on windows&&macos&&linux
I am afraid we can not upgrade it to a new version easily

we need to do lots of tests before the upgrading
and we will do it in next big version

anyway, @bambu123 please check the above question

hadess:
Note that some changes/fixes are going to be necessary upstream to use this patch:
- there's a change between 3.1.5 and 3.1.6 that broke the "Flushing Volumes" dialogue
- nanoSVG support will need to be verified

hadess:
> there's a change between 3.1.5 and 3.1.6 that broke the "Flushing Volumes" dialogue

See https://github.com/wxWidgets/wxWidgets/issues/24249

lanewei120:
no idea currently

there is a patch "deps/wxWidgets/0001-wxWidget-fix.patch" applied to wxWidgets
and there are lots of the conflicts with 3.2.4
how did you solve those conflicts?

and maybe you can try to remove this patch and check whether the issue is there or not
if no issue, then we can focus on this patch file

lanewei120:
a similiar issue before https://github.com/bambulab/BambuStudio/issues/2805

hadess:
> there is a patch "deps/wxWidgets/0001-wxWidget-fix.patch" applied to wxWidgets
> and there are lots of the conflicts with 3.2.4
> how did you solve those conflicts?

By manually applying all the changes to the upstream wxWidgets.

> and maybe you can try to remove this patch and check whether the issue is there or not if no issue, then we can focus on this patch file

This isn't the problem, we don't apply it in the Flatpak.

The preferred way to dealing with this patch would be to upstream all the fixes and changes.

lanewei120:
just tried on my windows
after building deps successfully
it can not pass compile at cmake stage

-- WXWIN environment set to:
CMake Error at C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find wxWidgets (missing: html adv gl core base webview aui net
  media) (Required is at least version "3.1")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.20/Modules/FindwxWidgets.cmake:1025 (find_package_handle_standard_args)
  src/CMakeLists.txt:50 (find_package)


See also "F:/work/projects/bambu_release/bamboo_slicer/build/CMakeFiles/CMakeOutput.log".
See also "F:/work/projects/bambu_release/bamboo_slicer/build/CMakeFiles/CMakeError.log".

hadess:
> after building deps successfully

If it can't find wxwidgets, then the deps aren't built and installed correctly. It's possible that newer wxwidgets have new or newer dependencies that you'll need to install and require to successfully test this.

I can't help you with the build though, I've never used Windows to do development.

fr3ak2402:
@hadess Did you also manage to use wxwidgets 3.2.4 with version 1.9.7.52?
I would like to try this under Windows 11.

hadess:
> @hadess Did you also manage to use wxwidgets 3.2.4 with version 1.9.7.52?

I haven't updated it since. Right now, we use the same version used in the stable version of wxWidgets with a couple of patches:
https://github.com/flathub/com.bambulab.BambuStudio/blob/master/com.bambulab.BambuStudio.yml#L168

BambuStudio 1.10.0 beta uses wxWidgets from the `master` branch, which I find quite surprising. We'll see whether they keep using it for the next stable.

hadess:
> BambuStudio 1.10.0 beta uses wxWidgets from the `master` branch, which I find quite surprising. We'll see whether they keep using it for the next stable.

That same branch is used in the stable versions of Bambu Studio, and the Flatpak uses the latest stable version of wxWidgets:
https://github.com/flathub/com.bambulab.BambuStudio/pull/84
So closing this MR.

PR #3455, opened 2024-01-22
All printers: Change printable height to 256 mm

GuzziRaz:
The Z-hop problem was already taken care of so this limitation is obsolete.

### DIFF
--- a/resources/profiles/BBL/machine/fdm_machine_common.json
+++ b/resources/profiles/BBL/machine/fdm_machine_common.json
@@ -74,7 +74,7 @@
     "min_layer_height": [
         "0.08"
     ],
-    "printable_height": "250",
+    "printable_height": "256",
     "extruder_clearance_radius": "65",
     "extruder_clearance_height_to_rod": "36",
     "extruder_clearance_height_to_lid": "140",


SaltWei:
Considering the frequent debris under the heated bed, using 256mm z hight for all users might damage the heated bed.

GuzziRaz:
> Not all users have the habit of cleaning the chassis. Considering the frequent solid waste under the heated bed, using 256mm z height for all users might damage the heated bed.

So you decided MY printable height should be less than advertised, because stupid users might not clean their printers? That is ridiculous.

vanowm:
> The Z-hop problem was already taken care of so this limitation is obsolete.

How was it fixed?

@mytechguyri not false advertising if hardware is capable, which it is. Nobody stops you from following [FAQ](https://wiki.bambulab.com/en/knowledge-sharing/print-volume-limitations) and modify your settings to remove any limitations. Note, that if you have any of these "dust caps" on z-axis screws, they will limit max height with very pleasant screeching sound and ruined top of the model (guess how I know that...)

GuzziRaz:
> > The Z-hop problem was already taken care of so this limitation is obsolete.
> 
> How was it fixed?

The printer can now do z-hops up to a certain height. Before this you had to turn it off for the entire print if you wanted to use the full height.

PR #3602, opened 2024-02-17
Update and rename Generic PLA.json to Hacthbox PLA.json

cCloudTVEPG:


XunZhangBambu (changes requested):
This shouldn't be renamed. If you want to rename it, just save the profile as personal profile locally

PR #3778, opened 2024-03-26
Turkish Translation Updates for 01.09.00.60 "BambuStudio_tr.po" add pls

fatih5228:
Hi. Can you add Turkish language pack?


lanewei120:
thanks for doing this, we will add the logic soon

PR #3783, opened 2024-03-27
1.9.0 Public Beta compilation fixes

hadess:


lanewei120:
thanks for doing this

PR #3843, opened 2024-04-08
master_build_1.9.0.69

lanewei120:
Change-Id: I6344dbdc54e30aae7de4aaddd7b929121aa5a678


Wiffman:
Hello,

I'm extremely new to github and I don't know if I'm doing this right. But I wanted to make a suggestion to add something to Bambulab Studio. My suggestion is, can the slice plate button get made into two buttons. A slice all button and slice plate button, instead of having a pull down menu for only two options please. I get it for the print button to have a drop down menu because it has six options. But the slice plate button only has two options and I use those two way more often then the print plate drop down. so its a bit annoying to click the slice plate drop down just to switch to slice all or slice plate. I feel like a lot of people deal with the same thing as I do with the slice drop down menu. Here is how I see the buttons layout could be better, from left to right you would have the upload button then have another button that is the slice all button then the slice plate button then have the print plate button with its drop down menu like it already is. Thank you for hearing me out and I think this is a small but worthy addition to the software.

lanewei120:
@Wiffman 
thanks for the suggestion
as the topbar space is limited, especially on some small monitor
we tried to reduce the button count

we will discuss this with our GUI designer

Wiffman:
I work on 1080p 22 inch monitors that i picked up at goodwill for $25 each and it still has room. I know thats right in the middle because 1080p isnt the biggest nor the smallest. I feel like most people are using 1080p monitors. Do you have any way of knowing what is the most common resolutions? if not you should do a poll and see

PR #3875, opened 2024-04-12
Update BambuStudio_tr.po

fatih5228:
Update Turkish language


lanewei120:
thanks for doing this

PR #3928, opened 2024-04-22
Update fdm_filament_pla.json

smwoodward:
When ever there is an object on the side of the plate nearest the aux fan, it always curls when the cooling fan is at 70%.  50% has worked just as well on other parts being printed and the parts near the edge of the bed don't warp.


XunZhangBambu:
@smwoodward 
Thank you. Since this is the parent node of the PLA filament, modifying this parameter will affect all parameters of the PLA preset, so it needs to be thoroughly tested. Until then, this commit may not be able to be merged.

PR #3942, opened 2024-04-24
Update BambuStudio_tr.po

fatih5228:
The newly added words have been translated into Turkish. Translation errors have been corrected.


lanewei120:
thank you

PR #3948, opened 2024-04-24
Fixed logic error in PLA jamming gcode for P1S

afriede:
Changed the P1S gcode for preventing PLA jamming from:
(swapped the inner if and elsif around)

```gcode
;=============turn on fans to prevent PLA jamming=================\n
{if filament_type[initial_extruder]==\"PLA\"}\n
    {if (bed_temperature[initial_extruder] >45)||(bed_temperature_initial_layer[initial_extruder] >45)}\n
    M106 P3 S180\n
    {elsif (bed_temperature[initial_extruder] >50)||(bed_temperature_initial_layer[initial_extruder] >50)}\n
    M106 P3 S255\n
    {endif};Prevent PLA from jamming\n
{endif}\n
```
to:
```gcode
;=============turn on fans to prevent PLA jamming=================\n
{if filament_type[initial_extruder]==\"PLA\"}\n
    {if (bed_temperature[initial_extruder] >50)||(bed_temperature_initial_layer[initial_extruder] >50)}\n
    M106 P3 S255\n
    {elsif (bed_temperature[initial_extruder] >45)||(bed_temperature_initial_layer[initial_extruder] >45)}\n
    M106 P3 S180\n
    {endif};Prevent PLA from jamming\n
{endif}\n
```
Before this change even if the bed temperature is greater than 50 the fans would not turn up as fast as they are supposed to, because the first if condition would also trigger and set the fan speed to 180 and not to 255!

The changes are difficult to visualize, because the gcode is all in one line, but by using this command locally the changes can easily be verified.
```bash
git diff --word-diff=color
```


lanewei120:
thanks for doing this

afriede:
You're welcome. I'm glad I could help!

PR #3968, opened 2024-04-28
fix: Upgrade dependencies and code to work on new macOS SDK

queengooborg:
This PR performs a number of updates to make Bambu Studio compile on the latest macOS SDK (14.4 at the time of writing).

The particular changes made are the following:

- Bump CGAL to v5.6.1
- Bump Boost to v1.85.0
  - A frozen copy of Boost will no longer need to be maintained!
  - The `filesystem/string_file.hpp` header was deprecated and had been removed entirely in v1.83.0, so code utilizing it has been replaced
  - Various other methods were deprecated in Boost and have been removed, so they have been replaced according to Boost's suggestions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,9 @@ if(SLIC3R_STATIC)
     set(Boost_USE_STATIC_LIBS ON)
     # Use boost libraries linked statically to the C++ runtime.
     # set(Boost_USE_STATIC_RUNTIME ON)
+
+    # Enable deprecated timer header (XXX replace header)
+    add_definitions(-DBOOST_TIMER_ENABLE_DEPRECATED)
 endif()
 #set(Boost_DEBUG ON)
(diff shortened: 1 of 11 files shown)


MackBambu:
HI,@queengooborg
Thank you for submitting the PR
Has this PR resolved all issues, and is it ready for validation?

emilazy:
(Uh, very sorry about the notifications spam here, I forgot that GitHub likes to do that whenever you reference an issue in a commit message. Suffice to say that downstream distributors would love to see this brought to completion!)

queengooborg:
It's alright, those messages don't create emails!

I have mostly stepped away from this PR as my work commitments have taken up the free time I originally had to work on this.  It might be best to split this up into smaller chunks and update one or two dependencies per PR.  If I remember correctly, this was tricky because newer versions of the libraries wanted C++17.

PR #3971, opened 2024-04-28
Remove redundant CMakeLists code for Linux

queengooborg:
This PR removes some redundant code in the CMakeLists for Linux compilation.  The code removed implements a workaround for older versions of CMake that did not support `CMAKE_CXX_STANDARD`; however, the very first line in the CMakeLists requires a version of CMake newer than the version that implemented the worked-around feature.  Therefore, the `if` clause will never be true!


lanewei120:
thanks for doing this

Issue #1192, opened 2023-01-27
Full Screen Camera View in Bambu Slicer


RiverRaid:
Hi,

It would be great, if the Camera View (in the Bambu Slicer => Device) could be Displayed as Fullscreen (e.g. with a double click on it, and in fullscreen-mode doubleclick or escape returns to the normal view)

BTW: Pinch to zoom in the Mobile app would be also great

Best regards

daylerees:
👍

PR #3974, opened 2024-04-28
Added fullscreen support to the camera stream view on macOS

flickerfree:
Modified wxMediaCtrl2 to use NSView::enterFullScreenMode and exitFullScreenModeWithOptions to toggle fullscreen.
You can enter/exit by clicking on the video and pressing escape. A fullscreen button was also added as an overlay to the video view that can be used to enter/exit fullscreen.

This pull request addresses issue #1192 for macOS.

### DIFF
--- a/resources/images/fullscreen.svg
+++ b/resources/images/fullscreen.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   id="图层_1"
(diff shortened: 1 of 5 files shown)


bambu123:
Is there a solution on the Windows platform?

victornpb:
I really like this

in addition to this do you know how you can prevent the camera from turning off when the printer is idle? i dont care if its hacky

Issue #3977, opened 2024-04-29
Use after free bug in LinesBucketQueue::emplace_back_bucket


flickerfree:
### Bambu Studio Version

1.9.1.58

### Where is the application from?

Built from source by yourself or others

### OS version

macOS Sonoma 14.4.1

### Additional system information

MacBook Air M1

### Printer

Bambu Lab X1C

### How to reproduce

I found a use after free bug in LinesBucketQueue::emplace_back_bucket. This was found by enabling address sanitizer. I also had undefined behavior and malloc scribble enabled, but those are not required to reproduce the problem.
1) Build with Address Sanitizer enabled.
2) Run under Xcode
3) Slice a plate that has one object and a wipe tower. The problem might reproduce even with one object, but I haven't tried that.

### Actual results

Crash in LinesBucketQueue::emplace_back_bucket (see screenshot and more details below):

<img width="1377" alt="Screenshot 2024-04-28 at 5 30 19 PM" src="https://github.com/bambulab/BambuStudio/assets/14137664/60cf8e7c-55f2-4104-8f82-7d99ff243b31">

I debugged the issue and will submit a pull request soon.

The LinesBucketQueue class has two related members:
std::vector<LinesBucket> line_buckets;
std::priority_queue<LinesBucket *, std::vector<LinesBucket *>, LinesBucketPtrComp> line_bucket_ptr_queue;
line_bucket_ptr_queue holds pointers into line_buckets. However, since items are inserted into line_buckets one at a time, existing pointers that were stored inside line_bucket_ptr_queue become invalid. Specifically:
void LinesBucketQueue::emplace_back_bucket(ExtrusionLayers &&els, const void *objPtr, Point offset)
{
    auto oldSize = line_buckets.capacity();
    line_buckets.emplace_back(std::move(els), objPtr, offset); <--- Causes a reallocation, making previous pointers invalid
    line_bucket_ptr_queue.push(&line_buckets.back()); <-- priority queue compares against old, now invalid pointers
    ...

### Expected results

It should not crash

### Project file & Debug log uploads

I was testing using this: https://makerworld.com/en/models/422168#profileId-325155

### Checklist of files to include

- [ ] Log file
- [ ] Project file

XunZhangBambu:
@flickerfree  Thank you very much for identifying this issue. Under your pull request #3978 , I have proposed a solution.

PR #3978, opened 2024-04-29
Fix use after free bug in LinesBucketQueue::emplace_back_bucket (#3977)

flickerfree:
Fix for #3977 - Fix use after free bug in LinesBucketQueue::emplace_back_bucket.
I found a use after free bug in LinesBucketQueue::emplace_back_bucket. This was found by enabling address sanitizer. The LinesBucketQueue class has two related members:
std::vector<LinesBucket> line_buckets;
std::priority_queue<LinesBucket *, std::vector<LinesBucket *>, LinesBucketPtrComp> line_bucket_ptr_queue; 

line_bucket_ptr_queue holds pointers into line_buckets. However, since items are inserted into line_buckets one at a time, existing pointers that were stored inside line_bucket_ptr_queue become invalid. Specifically:
void LinesBucketQueue::emplace_back_bucket(ExtrusionLayers &&els, const void *objPtr, Point offset) {
    auto oldSize = line_buckets.capacity();
    line_buckets.emplace_back(std::move(els), objPtr, offset); <--- Causes a reallocation, making previous pointers invalid
    line_bucket_ptr_queue.push(&line_buckets.back()); <-- priority queue compares against old, now invalid pointers
...
The code was already trying to handle the case where the capacity changes (which invalidates the pointers), but still pushed into the queue before recreating the queue from scratch.


XunZhangBambu:
```
    auto oldSize = line_buckets.capacity();
    line_buckets.emplace_back(std::move(els), objPtr, offset);

    auto newSize = line_buckets.capacity();
    if(oldSize == newSize){
        line_bucket_ptr_queue.push(&line_buckets.back());
    }
    else{ // pointers change
        decltype(line_bucket_ptr_queue) newQueue;
        for (LinesBucket &bucket : line_buckets) { newQueue.push(&bucket); }
        std::swap(line_bucket_ptr_queue, newQueue);
    }
```
I think it can be modified like this.

flickerfree:
> @flickerfree Good job !
> 
> ```
>     auto oldSize = line_buckets.capacity();
>     line_buckets.emplace_back(std::move(els), objPtr, offset);
> 
>     auto newSize = line_buckets.capacity();
>     if(oldSize == newSize){
>         line_bucket_ptr_queue.push(&line_buckets.back());
>     }
>     else{ // pointers change
>         decltype(line_bucket_ptr_queue) newQueue;
>         for (LinesBucket &bucket : line_buckets) { newQueue.push(&bucket); }
>         std::swap(line_bucket_ptr_queue, newQueue);
>     }
> ```
> 
> I think adding this piece of code will make it safer

Thanks for the review @XunZhangBambu . Do you prefer to keep my original reserve changes as well, or just use your proposed change. My original change will slightly reduce the number of allocations, but probably insignificant since the number of elements is typically small.

flickerfree:
@XunZhangBambu, I updated based on your suggestion. I removed my previous changes as it was adding additional complexity for only a small benefit.

XunZhangBambu:
> @XunZhangBambu, I updated based on your suggestion. I removed my previous changes as it was adding additional complexity for only a small benefit.

Thanks. I thin this commit can be merged now.

PR #4099, opened 2024-05-18
Added pt_br translation on web part

DvoraGames:
### Modifications:

Added brazilian portuguese translation to the web part.

Also added pt_br as a code safe language.

### Screenshot:

image

### Issues:

**Login problem (Caused by the URL of the login page, because pt_br doesn't exist.):**

image


lanewei120:
thanks for doing this

DvoraGames:
> thanks for doing this



> thanks for doing this

You're welcome.

Now you only need to add the "pt" code in the URL for the login page. Without this, we can't log in when the software is in Portuguese.

I can help translate the website too, but only in my free time.

PR #4102, opened 2024-05-18
Disable use of system eigen3 for OpenCV.cmake

astrolemonade:
This is a port of https://github.com/SoftFever/OrcaSlicer/pull/5311 which is a fix for https://github.com/SoftFever/OrcaSlicer/issues/5194 which happens to apply to Bambu Studio as well


lanewei120:
thanks for doing this

PR #4113, opened 2024-05-19
Mispellings fixes

bistory:
Fixed two mispellings : 

1. PrintingPause instead of Printing Pause
2. printetrs instead of printers


StoneLiBambu:
Hi @bistory 
    We will fix these two issues. The po file is automatically generated by the project and does not need to be submitted in the patch, so your patch will not be merged. Thank you for your understanding and feedback.

Thanks!

Stone Li

PR #4179, opened 2024-05-30
Update Korean translation

lunDreame:


lanewei120:
thanks for doing this

PR #4189, opened 2024-05-31
Complete translation of the entire Korean language

lunDreame:
The translation of the Bambu Studio Korean language file has been completed.


lanewei120:
we just released 1.9.2
could you use the newest version and update the translation files?

lunDreame:
@lanewei120 

Yes, if it's not a lot, I'll update it within a few hours

lunDreame:
@lanewei120 
is it possible to add the revised part based on the above update file? There are too many changes with all files

lanewei120:
@StoneLiBambu 
as there are too many changes, please give a solution

lunDreame:
@StoneLiBambulab 
Is it possible to work directly with Localazy? If you reflect the translation I translated, you can complete the translation added from official release 1.9.2

PR #4195, opened 2024-06-02
Update BambuStudio_tr.po

fatih5228:
Update Turkish language


lanewei120:
could you use the newest version and update the translation files?

fatih5228:
Added

PR #4196, opened 2024-06-02
PET support

hogum:
Based on PC
Print test pass time is greater than 30 hours
Unmodified id
从淘宝&拼多多购买的PET塑钢线耗材

### DIFF
--- a/resources/info/filament_info.json
+++ b/resources/info/filament_info.json
@@ -8,6 +8,7 @@
         "PA-CF",
         "PA-GF",
         "PA6-CF",
+        "PET",
         "PET-CF",
         "PPS",
         "PPS-CF",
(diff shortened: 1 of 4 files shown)


XunZhangBambu:
Sorry for taking a long time to reply to you. For the filaments purchased by yourself, you can add them as private filament preset. Since they may not be general enough, it is not recommended to add them to the public preset.

yzheka:
> Sorry for taking a long time to reply to you. For the filaments purchased by yourself, you can add them as private filament preset. Since they may not be general enough, it is not recommended to add them to the public preset.

The problem is that there is no PET filament type in the dropdown menu. I was able to add custom PET filament by creating PETG profile then exporting it to json file and manually editing it in the notepad app. After importing it again I got proper filament type. To cover all cases (for example PBT filament type is missing too) maybe it is better to allow user to input it manually, like Filament vendor value?

Paulomart:
I would also support a PET profile.

PR #4236, opened 2024-06-07
ja .po file contribution

do-gugan:
Randomly translated English displays that caught my eye into Japanese.
I apologize if this first commit doesn't adhere to the conventional expressions used in Bambu Studio or 3D printing.


lanewei120:
thanks for doing this

PR #4279, opened 2024-06-13
Update BambuStudio_pt-BR.po

ramorimdias:
As asked,
updated strings on 1.9.2
add missing text
correct wrong translations


ramorimdias:
Just updated the file, I had uploaded the same file before, my bad 😅

StoneLiBambu:
hi @ramorimdias ,
    First of all, thank you for your contribution.  refs to the wiki page https://wiki.bambulab.com/en/bambu-studio/contribute-languages. We already have a maintained platform for this language.  so we will only update and download it from our translation platform.

Thanks！
Stone Li

ramorimdias:
I noticed that the translations for Brazilian Portuguese aren't updated yet. They are very bad at the actual state. Please consider updating for the next release, please.

PR #4358, opened 2024-06-22
Copy fan min/max from Bambu PETG to fix excessive cooling in Generic PETG

StdVectorBool:
The Generic PETG profile for Bambu machines had 40-90 cooling which is even more than PLA.  Copy more sane values from the Bambu PETG profile since PETG needs lower fan for better layer adhesion.

### DIFF
--- a/resources/profiles/BBL/filament/Generic PETG @base.json
+++ b/resources/profiles/BBL/filament/Generic PETG @base.json
@@ -21,10 +21,10 @@
         "30"
     ],
     "fan_max_speed": [
-        "90"
+        "40"
     ],
     "fan_min_speed": [
-        "40"
+        "10"
     ],
     "filament_flow_ratio": [
         "0.95"


StdVectorBool:
@XunZhangBambu Can you approve?

XunZhangBambu:
@StdVectorBool  You can modify your own configuration to get better printing

StdVectorBool:
:shrug:  For the bridging and overhang concerns you described, both the Bambu PETG and Generic PETG profiles currently have identical settings:
* Force cooling for overhangs and bridges: Enabled 
* Cooling overhang threshold: 10%
* Fan speed for overhangs: 90%
 
Just wanted to let you know that I tried two different PETG brands using the "Generic PETG" profile and the excess "40-90%" fan thresholds caused very weak prints with bad layer adhesion (sometimes snapped during print) until I switched over to the Bambu PETG profile.

StdVectorBool:
> @StdVectorBool You can modify your own configuration to get better printing

I'm fine creating my own configuration from it, just saying it's likely also a bad starting point for most people.

PR #4479, opened 2024-07-12
Adding 3rd party manufacturer Two Trees Sp-5 Sapphire Plus V1.1

abdbali:
Adding 3rd party manufacturer Two Trees Sp-5 Sapphire Plus V1.1 

The Sapphire Plus v1.1 model is an old but powerful printer, I still use it with its large area and direct drive conversion, but it does not have its own basic profile in bambulab, it is still a good printer with its retrofit kits and mks motherboard.

For this reason, I have uploaded the profile and stl data of Sapphire Plus to my page. It is important that you review it.
[TwoTrees.zip](https://github.com/user-attachments/files/16192129/TwoTrees.zip)

I recommend you to add the profile given in the image and locations, I will follow and update for better profile settings.

TwoTrees_v1 1_logo
TwoTrees_v1 1_cover


abdbali:
@SaltWei What are we waiting for, let's do this! :))

QingZhangBambu:
@abdbali please check the comment and modify the files

QingZhangBambu:
@abdbali have you update the files?

PR #4489, opened 2024-07-14
Fix typo in calibration string

t3chguy:


lanewei120:
@t3chguy 
hello
we just need to modify src/slic3r/GUI/CalibrationWizardSavePage.cpp
other files are auto generated

thanks

t3chguy:
@lanewei120 done

lanewei120:
thank you~

PR #4532, opened 2024-07-21
Set default sparse infill to crosshatch

mbifulco:
Inspired by [this thread on reddit](https://www.reddit.com/r/BambuLab/comments/1e7voun/petition_to_remove_grid_infill_as_default/), this PR updates the default Sparse infill option to `CrossHatch`, rather than `Grid`.

Why this change?

> The grid infill, while simple and fast, has a notable drawback: material accumulates at the crossing points, which can cause annoying noise or even print failures due to the nozzle going over these material accumulations. The crosshatch infill mitigates this issue by distributing material more evenly, reducing the risk of these complications. 

[source: Prusa Blog](https://blog.prusa3d.com/everything-you-need-to-know-about-infills_43579/)

I had many prints fail as a result of awkwardly overlapping grid infill - it seems like an odd choice for the default option.

### DIFF
--- a/src/libslic3r/PrintConfig.cpp
+++ b/src/libslic3r/PrintConfig.cpp
@@ -1623,7 +1623,7 @@ void PrintConfigDef::init_fff_params()
     def->enum_labels.push_back(L("Support Cubic"));
     def->enum_labels.push_back(L("Lightning"));
     def->enum_labels.push_back(L("Cross Hatch"));
-    def->set_default_value(new ConfigOptionEnum<InfillPattern>(ipCubic));
+    def->set_default_value(new ConfigOptionEnum<InfillPattern>(ipCrossHatch));
 
     def = this->add("top_surface_acceleration", coFloat);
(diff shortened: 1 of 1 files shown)


mbifulco:
Any thoughts on this one? Seems like it would provide a better end user experience for Bambu customers.

MadeWithLayers posted this great video comparing some infill types yesterday - interestingly, he landed on Cubic as the infill type of choice, but that's more balancing speed/strength than print failures. As far as I can tell, cubic infill still crosses over itself on individual layers. https://www.youtube.com/watch?v=nV3GbN6hLjg

dragon-eyes-bbl:
Hi @mbifulco,
Thanks for the works. 
We've been working on improving the CrossHatch infill since we first designed it. However, there are still some issues that could affect print appearance. We're committed to optimizing it to meet our expectations. Until then, we will keep refining it and will switch it to the default once we are happy with the results.

PR #4584, opened 2024-08-04
Add Docker flag to remove containers after finishing

penguineer:
When using `DockerRun.sh` stale containers were left over after closing the application. These containers would also not be re-used.

Add the `--rm` flag to make Docker delete the containers after running the application, since they are re-created for each run anyways.


MackBambu (approved):

penguineer:
@MackBambu Thank you for the approval. It seems the workflow run is still waiting an approval.

Is there any further action I need to take for this PR?

PR #4611, opened 2024-08-09
Critical bugfix: re-add spiral z hop on layer changes when timelapse is disabled

ziehmon:
## Update 2024-08-13

The new PR is ready which entirely fixes the problem without a workaround. Please see https://github.com/bambulab/BambuStudio/pull/4631. I will close this one sooner or later of coordination with the maintainers.

## Update 2024-08-12

I want to apologize for the tone of my latest update. It was not my intention to threat or accuse contributors. I sat in front of my computer for way too long this day.

It seems I was able to track down the original cause and fix it cleanly without workarounds. I will open a new PR for this, since this would serve as just another workaround.

## Update 2024-08-10

This issue is larger then expected. My PR will just mitigate a fundamental bug in `change_layer` in  `GCode.cpp`. The spiral mode shouldn't even occur in `timelapse_gcode` but rather inside the gcode parser. It will always use normal lift, no matter how I manipulate the settings. 

It seems you were absolutely aware of this, because you added the spiral lift as a workaround in the timelapse code. If you'd implemented this correct, this would cause double Z hops. A bit disappointing guys. Sort this out, I am seriously considering doing a RMA. I'll leave this PR as it is so you can understand the issue.

## Scope

Bambu Lab A1 series (Mini/regular)

## Abstract

On layer changes, the Bambu Lab A1 is supposed to use a spiral z hop to mitigate stringing as documented in https://wiki.bambulab.com/en/software/bambu-studio/parameter/retraction#auto.

The gcode for timelapse handling has a bug since its initial release in 2023 that will only execute the spiral Z hop when timelapse is activated and hence use a normal straight Z hop instead. This is causing stringing on prints with short layer time. 

This PR moves the spiral gcode before the timelapse condition is trigged. It will still support the additional instructions inside the timelapse handling conditional.

## Details

Look at this gcode section from `time_lapse_gcode`. It also viewable in Bambu Studio ("Printer Settings" -> "Machine gcode"):

```gcode
{if !spiral_mode && print_sequence != "by object"}
; don't support timelapse gcode in spiral_mode and by object sequence for I3 structure printer
M622.1 S1 ; for prev firware, default turned on
M1002 judge_flag timelapse_record_flag
M622 J1
    [...]
    G17
    G2 Z{layer_z + 0.4} I0.86 J0.86 P1 F20000 ; spiral lift a little
    [...]
M623
[...]
G1 Z7.8 F42000
```
[...]

PR #4631, opened 2024-08-13
Bugfix: support spiral lift Z hop without timelapse enabled (fix stringing)

ziehmon:
This is the successor of https://github.com/bambulab/BambuStudio/pull/4611, which was just another workaround after detailed analysis.

## Problem
As outlined in the original PR, BambuStudio will not use a spiral Z hop for layer changes when timelapse is disabled and there is no infill. Actually, it will not use a spiral Z hop in other scenarios too (whenever timelapse gcode is injected), but they don't matter as much as the layer changes.

- Find a slow-motion video here: [imgur](https://imgur.com/a/1P2z3HZ)
- Find a .3mf / gcode file here: [missing_spiral_z.zip](https://github.com/user-attachments/files/16603440/missing_spiral_z.zip)

**Screenshot show the model's missing spiral Z hop:**

grafik

**Screenshot after successful fixing of the bug (see "Solution/PR Content"):**

grafik

## Cause

Timelapse gcode is injected into the model and Bambu Studio is technically not aware what it does - the user can change it individually. That means the position is changed without the gcode processor in Bambu Studio knowing or causing it. 

It's quite important for Bambu Studio to know the position of all axes. Most importantly to: 
1. prevent the collisions, e.g nozzle scratching when moving to next position because assumed position is wrong
2. do safety checks, e.g. if a spiral Z hop would move an axis out of its physical limits

For the Z position (2.), BambuStudio already extracts the last position from the `timelapse_gcode` and writes it to its internal state (`get_last_z_from_gcode`). But for  X and Y, it is not implemented. We don't know why and there actually was comment suggesting to do it ("TODO") in a previous commit, but it seems it was forgotten.

**Instead, a flag was implemented that marks the position as unknown/unsafe after importing `timelapse_gcode` and additionally set the Z hop type to normal. This definitely prevents collisions but ultimately causes stringing.**

## Solution/PR Content

**To make layer changes use spiral Z hop, we need to do two things:**

1. Change lift mode to spiral (the easy part)
2. Flag the position as safe, so import the axis positions after `timelapse_gcode` execution (the hard part)

So this PR does the following:

- Extent the `GCodeProcessor::get_last_z_from_gcode` method to support X and Y coordinates
- Change the usage of this method in timelapse handling to retrieve X and Y position too
- Remove the flag `m_writer.set_current_position_clear(false);` that sets the position as unknown/unsafe
- Use the optimized method/approach for other gcode injection too (`change_filament_gcode`)

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -527,16 +527,36 @@ static std::vector<Vec2d> get_path_of_change_filament(const Print& print)
             toolchange_gcode_str = toolchange_retract_str + toolchange_gcode_str;
             //BBS
             {
-                //BBS: current position and fan_speed is unclear after interting change_filament_gcode
+
                 check_add_eol(toolchange_gcode_str);
+
+                //BBS: gcode writer doesn't know fan speed after inserting tool change gcode
                 toolchange_gcode_str += ";_FORCE_RESUME_FAN_SPEED\n";
-                gcodegen.writer().set_current_position_clear(false);
-                //BBS: check whether custom gcode changes the z position. Update if changed
-                double temp_z_after_tool_change;
-                if (GCodeProcessor::get_last_z_from_gcode(toolchange_gcode_str, temp_z_after_tool_change)) {
-                    Vec3d pos = gcodegen.writer().get_position();
-                    pos(2) = temp_z_after_tool_change;
-                    gcodegen.writer().set_position(pos);
+                
+                //BBS: check whether custom gcode changes the axis positions. Update if changed.
+                bool position_changed = false;
+                Vec3d new_pos = gcodegen.writer().get_position();
+
+                double temp_x_after_toolchange_gcode;
+                if (GCodeProcessor::get_last_pos_from_gcode(toolchange_gcode_str, 0, temp_x_after_toolchange_gcode)) {
+                    new_pos(0) = temp_x_after_toolchange_gcode;
+                    position_changed = true;
+                }
+                
+                double temp_y_after_toolchange_gcode;
+                if (GCodeProcessor::get_last_pos_from_gcode(toolchange_gcode_str, 1, temp_y_after_toolchange_gcode)) {
+                    new_pos(1) = temp_y_after_toolchange_gcode;
+                    position_changed = true;
+                }
+                
+                double temp_z_after_toolchange_gcode;
+                if (GCodeProcessor::get_last_pos_from_gcode(toolchange_gcode_str, 2, temp_z_after_toolchange_gcode)) {
+                    new_pos(2) = temp_z_after_toolchange_gcode;
+                    position_changed = true;
+                }
+                
+                if (position_changed) {
+                    gcodegen.writer().set_position(new_pos);
                 }
             }
 
@@ -3073,16 +3093,37 @@ GCode::LayerResult GCode::process_layer(
     gcode += this->change_layer(print_z);  // this will increase m_layer_index
     m_layer = &layer;
     m_object_layer_over_raft = false;
+
+    // insert timelapse_gcode when traditional mode is not used (smooth mode)
     if (printer_structure == PrinterStructure::psI3 && !need_insert_timelapse_gcode_for_traditional && !m_spiral_vase && print.config().print_sequence == PrintSequence::ByLayer) {
-        std::string timepals_gcode = insert_timelapse_gcode();
-        gcode += timepals_gcode;
-        m_writer.set_current_position_clear(false);
-        //BBS: check whether custom gcode changes the z position. Update if changed
-        double temp_z_after_timepals_gcode;
-        if (GCodeProcessor::get_last_z_from_gcode(timepals_gcode, temp_z_after_timepals_gcode)) {
-            Vec3d pos = m_writer.get_position();
-            pos(2) = temp_z_after_timepals_gcode;
-            m_writer.set_position(pos);
+
+        std::string timelapse_gcode = insert_timelapse_gcode();
+        gcode += timelapse_gcode;
+        
+        //BBS: check whether custom gcode changes the axis positions. Update if changed.
+        bool position_changed = false;
+        Vec3d new_pos = m_writer.get_position();
+        
+        double temp_x_after_timelapse_gcode;
+        if (GCodeProcessor::get_last_pos_from_gcode(timelapse_gcode, 0, temp_x_after_timelapse_gcode)) {
+            new_pos(0) = temp_x_after_timelapse_gcode;
+            position_changed = true;
+        }
+        
+        double temp_y_after_timelapse_gcode;
+        if (GCodeProcessor::get_last_pos_from_gcode(timelapse_gcode, 1, temp_y_after_timelapse_gcode)) {
+            new_pos(1) = temp_y_after_timelapse_gcode;
+            position_changed = true;
+        }
+        
+        double temp_z_after_timelapse_gcode;
+        if (GCodeProcessor::get_last_pos_from_gcode(timelapse_gcode, 2, temp_z_after_timelapse_gcode)) {
+            new_pos(2) = temp_z_after_timelapse_gcode;
+            position_changed = true;
+        }
+        
+        if (position_changed) {
+            m_writer.set_position(new_pos);
         }
     }
     if (! print.config().layer_change_gcode.value.empty()) {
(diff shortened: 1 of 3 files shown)


Roemer:
Respect for your work. Hope this can be finished and merged anytime soon as I also suffer from some stringing.

ziehmon:
> Respect for your work. Hope this can be finished and merged anytime soon as I also suffer from some stringing.

I just finished it  and the compiler ran fine. I'm still doing some final test but would you be interested to test an AppImage for me, if you're able to run one?

XunZhangBambu (approved):

Roemer:
@XunZhangBambu The PRis from @ziehmon , I just volunteer to test if needed.

XunZhangBambu:
@Roemer Haha, sorry, I asked the wrong person. @ziehmon  Please look at the question I raised above.

ziehmon:
> @Roemer Good Job.I also want to know why you abandoned the previous PR. Extract the lift operation from the gcode is a simpler approch. Even if you record the position in code, the timelapse gcode will be skipped during actual printing if you disable timelapse, and the state recorded in the program will still be inaccurate.

Thanks @XunZhangBambu. The answer is simple: because the original approach in the previous PR causes a double lift when no timelapse is used!

BambuStudio used to (before this PR) **always insert a normal Z hop after `timelapse_gcode`** insertion. 

**Meaning**: if we **do a Z hop in the `timelapse_gcode` (previous PR), this adds a second hop to the existing one**. The Z hop in `timelapse_gcode` would be is a spiral, which fixes the problem, but **the normal Z hop in BambuStudio (before this PR) continued to exist**. This just adds another 0.4mm of Z hop which is unrequired and and in my tests also cause a little bit of stringing due to interrupting the retraction.

**However, even with this PR, there is a second Z hop:  when timelapse is actually enabled, the one that is written in `timelapse_gcode` after the condition to check if timelapse is enabled.**. This is not easily solvable because the gcode parser could not possibly know if it has to insert the Z hop at all - the user could disable timelapse after the gcode has beenm generated already. **Meaning: it's not perfect, but at least for disabled timelapse, its clean.** 

Also, by triggering spiral Z hop inside the gcode parser, we can utilize collision avoidance, feedrate generation etc.

Hope this makes it clear. Thank you for your review!

ziehmon:
After thinking about it again, we could remove the `get_last_pos_from_gcode` entirely, because the Z hop in `timelapse_gcode` does not do it too, right? That's you decision. I tried to keep the code the way you originally implemented, I thought you implemented the flag `m_writer.set_current_position_clear(false);` for a reason but since its not chcked in timelapse_gcode at all, maybe we can omit it.

ziehmon:
> @ziehmon could you send this pr to orcaslicer?

@lodgnewt Will look into it. The PR for Orca needs to be different because they integrate the upstream code from Bambu Lab manually with modifications to support other printers but it shouldn't differ too much. Orca definitely has the bug too, I checked that.

ziehmon:
@lanewei120 since this PR is approved, do think we can ship it into 1.9.4.59? I saw you're building this version right now.

lanewei120:
@ziehmon I am afraid it can not catch up with 1.9.4.59 for limited time

@XunZhangBambu please continue this discussion

XunZhangBambu:
@ziehmon The operation of get_position does not need to be removed, as at least when timeplase is enabled, we can obtain the correct position.In your test, does the spiral have less wire drawing compared to the normal lift?

ziehmon:
> The operation of get_position does not need to be removed, as at least when timeplase is enabled, we can obtain the correct position

Perfect, that was my idea too.

> In your test, does the spiral have less wire drawing compared to the normal lift?

Absolutely, yes. Especially with low layer times. I attached the .3mf of my test model. Stringing was gone entirely after the modification.

It is noted in the Bambu Studio UI, that spiral Z hop will prevent stringing.

Also documented in [Bambu Wiki](https://wiki.bambulab.com/en/software/bambu-studio/parameter/retraction#auto)

tsmith35:
@ziehmon did you notice that "timelapse" was misspelled as "timepals" in the old code? Just noticed yesterday while digging through OrcaSlicer code.

ziehmon:
> @ziehmon did you notice that "timelapse" was misspelled as "timepals" in the old code? Just noticed yesterday while digging through OrcaSlicer code.

Yes, noticed this and fixed it together with my other changes. 

I looked into Orca too and I think I remember it being misspelled there too. Just a small typo, I guess.

ziehmon:
> Is the modified g-code compatible with the A1 Mini as well? I’m experiencing the exact same stringing issues.

@sydude the modified G Code was not continued due to causing other problems. This MR solves the actual issue and is compatible with all kind of Bambu Lab printers. The bad news is, to test it, you need to compile. Right now, I can only provide linux executables for testing.

ziehmon:
@XunZhangBambu was my answer to your questions sufficient? Do you need any other details?

I can recommend looking at the slow-motion video I did at [imgur](https://imgur.com/a/1P2z3HZ) to see it yourself.

ZackySteel:
Just trying to come up to speed, since the issue I was experiencing was merged into the same topic. I really haven't experimented with the various types of Z-hops, as I never tended to use them even before moving to Bambu/Orca, but I was still experiencing massive surface blemishes and knocked-over/failed prints due to what I've seen others have discovered to be missing retraction moves in the G-Code generated. I believe it was 1.8.1 people mentioned being the last release before this issue began, but I haven't gotten much of a chance to test with that version yet. I would have thought the only connection between these two topics would be that Z-hops are often done when retraction moves are scheduled to happen and that, as such, the slicer not generating G-Code for retraction moves would also mean it might not generate a Z-hop there either, but I feel that can't be the reasoning behind the merge. From what I have understood thus far (and I'm sure I don't have all the information), it seems like the missing retraction-moves are the core problem, rather than the missing Z-hops, so it seems odd to merge it into the Z-hops topic rather than vice-versa or even at all.

Just trying to make sure I understand everything here, so info from anyone more in-the-know than I am would definitely be appreciated! I'm really hoping this gets solved soon, as I'm up to my eyeballs in under-performing printers at the moment and want to get this A1 working so I can actually complete some of the prints that have been piling up!

ziehmon:
> Just trying to come up to speed, since the issue I was experiencing was merged into the same topic. I really haven't experimented with the various types of Z-hops, as I never tended to use them even before moving to Bambu/Orca, but I was still experiencing massive surface blemishes and knocked-over/failed prints due to what I've seen others have discovered to be missing retraction moves in the G-Code generated. I believe it was 1.8.1 people mentioned being the last release before this issue began, but I haven't gotten much of a chance to test with that version yet. I would have thought the only connection between these two topics would be that Z-hops are often done when retraction moves are scheduled to happen and that, as such, the slicer not generating G-Code for retraction moves would also mean it might not generate a Z-hop there either, but I feel that can't be the reasoning behind the merge. From what I have understood thus far (and I'm sure I don't have all the information), it seems like the missing retraction-moves are the core problem, rather than the missing Z-hops, so it seems odd to merge it into the Z-hops topic rather than vice-versa or even at all.
> 
> Just trying to make sure I understand everything here, so info from anyone more in-the-know than I am would definitely be appreciated! I'm really hoping this gets solved soon, as I'm up to my eyeballs in under-performing printers at the moment and want to get this A1 working so I can actually complete some of the prints that have been piling up!



> Just trying to come up to speed, since the issue I was experiencing was merged into the same topic. I really haven't experimented with the various types of Z-hops, as I never tended to use them even before moving to Bambu/Orca, but I was still experiencing massive surface blemishes and knocked-over/failed prints due to what I've seen others have discovered to be missing retraction moves in the G-Code generated. I believe it was 1.8.1 people mentioned being the last release before this issue began, but I haven't gotten much of a chance to test with that version yet. I would have thought the only connection between these two topics would be that Z-hops are often done when retraction moves are scheduled to happen and that, as such, the slicer not generating G-Code for retraction moves would also mean it might not generate a Z-hop there either, but I feel that can't be the reasoning behind the merge. From what I have understood thus far (and I'm sure I don't have all the information), it seems like the missing retraction-moves are the core problem, rather than the missing Z-hops, so it seems odd to merge it into the Z-hops topic rather than vice-versa or even at all.
> 
> Just trying to make sure I understand everything here, so info from anyone more in-the-know than I am would definitely be appreciated! I'm really hoping this gets solved soon, as I'm up to my eyeballs in under-performing printers at the moment and want to get this A1 working so I can actually complete some of the prints that have been piling up!

@ZackySteel may I ask you to have look on this comment I wrote: https://github.com/bambulab/BambuStudio/issues/3423#issuecomment-2335161516.

This issue is not related to your problem. It is about stringing and will help a lot with it, but probably not with other Z hop related issues such as surface scratching and knocking over prints.

ZackySteel:
> @ZackySteel may I ask you to have look on this comment I wrote: [#3423 (comment)](https://github.com/bambulab/BambuStudio/issues/3423#issuecomment-2335161516).
> 
> This issue is not related to your problem. It is about stringing and will help a lot with it, but probably not with other Z hop related issues such as surface scratching and knocking over prints.

Ah, yes, I read that and appreciate you taking the time to provide that list! I think I may have caused some confusion though, as I followed a link from the topic I was on to this one, as it was provided by someone who said it had been merged into this topic. I may very well just be confused about something, but that is what I was trying to figure out, as the two topics don't necessarily seem to have a ton of crossover aside from the fact that z-hops are often ordered to occur during retraction moves, but that seemed a someone thin relation to completely merge the two topics for.

I've already gone through pretty much everything I can to overcome the issue I found mention of in the topic being merged into this one with, unfortunately, with no success. Now I'm just trying to make sure I understand why it is being merged into this and if the underlying missing retraction move G-Code commands have been identified as a problem (I really hope that's the case) and if we can be looking forward to a fix or at least a timeline for a fix relatively soon.

Thanks again for taking the time to check in; I appreciate it!

tsmith35:
@ziehmon 
> Edit: look like the typo does not exist in Orca. @tsmith35 thanks for porting this to Orca (looks like you are about to do it).

I finally got a chance to spend a few minutes on it. PLMK if you see any issues.
[it's over here](https://github.com/tsmith35/OrcaSlicer/pull/1)

XunZhangBambu:
@ziehmon  yes, we will merge the pr soon ！

ziehmon:
> @ziehmon  yes, we will merge the pr soon ！

Very nice, thank you for you extensive review. I am on standby if you need anything 👍🏻

tsmith35:
@ziehmon eh, I'm not the smartest git user ever. :-) It's now here:

[PR#6709](https://github.com/SoftFever/OrcaSlicer/pull/6709)

ziehmon:
> @ziehmon eh, I'm not the smartest git user ever. :-) It's now here:
> 
> 
> 
> [PR#6709](https://github.com/SoftFever/OrcaSlicer/pull/6709)

Nonetheless, you made it and this PR looks clean! Thanks for your work. I will have a look into it tomorrow.

ZackySteel:
For myself and anyone else who are still mostly baffled by the workings of Github, what's being referred to in the last few comments? XD

PR mean pull request or push request? Mainly if anyone can english-ify it so I'll have that translated in my brain for when I see other stuff like it on here! Thanks!

Also, I finally got the files onto my home computer, yet realized it doesn't allow me to upload that file type. What's the normal protocol for uploading G-Code and other print-related files?

Thanks for all the help!

Issue #4841, opened 2024-09-19
X axis offset on multiple model when changing color with AMS lite 


jmilhau:
### Bambu Studio Version

1.9.7.50

### Where is the application from?

Bambu Lab Official website

### OS version

MacOS 15.0

### Additional system information

Since today (since latest Studio update? v1.9.7.50) any print I make with several colors in the same layer gets a x-axis gap a several millimeters. Maintenance did not help.

![image](https://github.com/user-attachments/assets/f152a22f-5e69-4c62-a733-5bc2edd26ef3)
![image](https://github.com/user-attachments/assets/9091f570-814f-4a49-a8d7-45208a53c43d)

Black was printed first, then purple then cyan then orange.

### Printer

Bambu Lab A1 Combo

### How to reproduce

Print a model with 2 colors on the same layer

### Actual results

Offset of the first color on the print
![image](https://github.com/user-attachments/assets/9091f570-814f-4a49-a8d7-45208a53c43d)

### Expected results

No offset
![image](https://github.com/user-attachments/assets/f152a22f-5e69-4c62-a733-5bc2edd26ef3)

### Project file & Debug log uploads

[test v1.3mf.zip](https://github.com/user-attachments/files/17063455/test.v1.3mf.zip)

### Checklist of files to include

- [ ] Log file
- [ ] Project file

jmilhau:
Re installing 1.9.5.51 solves the issue

BastR:
Same issue on A1, had to stop print. 
![IMG_3095](https://github.com/user-attachments/assets/d78a03ab-1ab9-458d-81e6-7fd8de4b0942)

MCKanpolat:
I've just updated the latest version and same issue I have. Mac OS 15

BastR:
I did a few tests with Windows and printing went off without a problem.

ziehmon:
Are you using timelapse by any chance?

BastR:
In my case no, no timelapse on the A1

jmilhau:
No timelapse either

ziehmon:
Allright, can you do another check just to make sure: remove time timelapse_gcode in extruder settings? It's a wild guess but worth a try. The behavior was changed in the new release.

NicolasABSuarezB:
I have the same problem with mi A1 (Bambu Studio v1.9.7) on Windows
Solved with the reinstallation of Bambu Studio v1.9.5.51  

[...]

PR #4848, opened 2024-09-20
Regression of #4631: support spiral lift with timelapse gcode and multi-color prints

ziehmon:
# Context

This is a regression of the previous pull request https://github.com/bambulab/BambuStudio/pull/4631, which had to be reverted due to bugs when used with multi-color/AMS Lite prints.

- [Bug report for 1.9.7.50](https://github.com/bambulab/BambuStudio/issues/4841)
- [Root cause of bug in comment](https://github.com/bambulab/BambuStudio/issues/4841#issuecomment-2362988198)

grafik

# Cause

**At the print sequence, where the described bug was happening, there is a difference between printing multi-filament with AMS Lite and printing single-filament:**

**Without multi-filament (so, without AMS Lite)**, the position before the Z hop for the layer change is executed, is `X=0.00`. This is set in "Timelapse G-Code" and will be executed as last step after the picture is taken. Other movements in the "Timelapse G-Code" exist too, but they also use a safe position, even though not `x=0.0000` (e.g. for nozzle clumping detection).

grafik

**But with multi-color (so, with AMS Lite)**, the position before the Z hop for the layer change is executed, is `X=-48.2`. This is set in "Change Filament G-Code" and is the last movement of the "wipe and shake" mechanism. That "wipe and shake" happens either after changing filament, and/or after extrusion calibration:

grafik

# Solution

**The cause of the bug raises two questions:**
1. Why was there no check if the printer is moving out ouf the physical boundaries?
	- Answer: there is no boundary check implemented when the spiral Z hop gets executed. This is where the check is [supposed to be done](https://github.com/bambulab/BambuStudio/blob/0e686206fc8101875e5a6568a977df3d5d3b89b4/src/libslic3r/GCodeWriter.cpp#L385):
	
		```
		//BBS: todo: check the arc move all in bed area, if not, then use lazy lift 	double radius = delta(2) / (2 * PI * atan(GCodeWriter::slope_threshold));
		Vec2d ij_offset = radius * delta_no_z.normalized();
		ij_offset = { -ij_offset(1), ij_offset(0) };
		slop_move = this->_spiral_travel_to_z(target(2), ij_offset, "spiral lift 	Z");
		```

2. How can we solve the problem?
	- Answer: we need to ensure that after the execution of gcode to change the filament, the position is safe for executing a spiral Z hop. This is already done for the timelapse gcode

# Implementation

For a multi-filament/AMS Lite print, there are two scenarios where we need to set a safe position afterwards:
- After wipe the filament change wipe is done, and when no extrusion calibration will be executed
- When an extrusion calibration happens after filament change and another wipe was executed

**The change is the same for both scenarios - the safe position just needs to be set in different sections of "Change Filament G-Code":**

**Before, without a safe position** (`X=-48.2`):

```
G1 X-38.2 F18000
G1 X-48.2 F3000
G1 X-38.2 F18000 ;wipe and shake
G1 X-48.2 F3000
G1 X-38.2 F12000 ;wipe and shake
G1 X-48.2 F3000
M400
```

**Afterwards, with a safe position** (`X=0.00`):

```
G1 X-38.2 F18000
G1 X-48.2 F3000
G1 X-38.2 F18000 ;wipe and shake
G1 X-48.2 F3000
G1 X-38.2 F12000 ;wipe and shake
G1 X-48.2 F3000
G1 X0 F18000 ; move to safe pos
M400
```

Result after implementation of bugfix:

grafik

# Summary

Additionally to the previous, included PR, this MR changes the `change_filament_gcode` (in GUI: "Change filament G-Code") for A1 and A1 Mini printers to set the postion safe (`X=0.00`) after filament was changed in two different scenarios.

This bugfix does not apply to anything else than A1 or A1 Mini. The original Z hop fix did not change behavior of other printers like P1S or X1C (non I3-Structure).

# Test-cases

Lesson learned - the testing was not sufficient last time. I came up with the following test cases to make sure, we don't have another bug. Personally, I can only cover A1 test-cases without AMS Lite. Maybe Bambu Lab can test it at scale on their print farm.

| Printer  | AMS Lite | Timelapse | Result |
|----------|----------|-----------|--------|
| A1       | Not used | Used      | TODO   |
| A1       | Used     | Used      | TODO   |
| A1       | Not used | Not used  | TODO   |
| A1       | Used     | Not used  | TODO   | 
| A1 Mini  | Not used | Used      | TODO   | 
| A1 Mini  | Used     | Used      | TODO   | 
| A1 Mini  | Not used | Not used  | TODO   | 
| A1 Mini  | Used     | Not used  | TODO   | 

# Appendix

grafik


XunZhangBambu:
@ziehmon Thank you very much for continuing to work on fixing this PR. I think we could discuss a more general approach to fix the issue
1. We need to check the spiral lift to ensure that when executing the spiral lift, it does not exceed the range of the bed. If it goes beyond the range, it can be replaced with a normal lift.
2. We need to find a way to get the nozzle's position after executing custom G-code. Currently, only checking G0, G1, G2, and G3 commands is not sufficient, as other commands could also move the nozzle.

tsmith35:
@ziehmon, I submitted a PR for OrcaSlicer to revert the modifications until the corrected modification has been tested.

ziehmon:
> @ziehmon, I submitted a PR for OrcaSlicer to revert the modifications until the corrected modification has been tested.

Thank you! I think SoftFever already reverted it

ziehmon:
> @ziehmon Thank you very much for continuing to work on fixing this PR. I think we could discuss a more general approach to fix the issue
> 
>     1. We need to check the spiral lift to ensure that when executing the spiral lift, it does not exceed the range of the bed. If it goes beyond the range, it can be replaced with a normal lift.
> 
>     2. We need to find a way to get the nozzle's position after executing custom G-code. Currently, only checking G0, G1, G2, and G3 commands is not sufficient, as other commands could also move the nozzle.

My pleasure, thank you for your support and review!

I absolutely agree. There are better approaches. The question is, how to implement them? From my perspective, we should do it in three steps.

**1. Step:** keep the current approach for now, but review that all gcode (`timelapse`, `toolchange`, `machine_start/end`) ends in the safe position of `X=0.00` and `Z>={last_layer_z}`. For Z, this is already given and for X, this PR adds it.

 **2. Step:** extent `GCodeProcessor::get_last_pos_from_gcode` in such a way, we reliably can retrieve all movement commands.

**3. Step:** in `GCodeWriter::travel_to_xyz` add a validation if the printer would move out of bounds. There is a comment already that mentions it still needs to be done[ here](https://github.com/bambulab/BambuStudio/blob/0e686206fc8101875e5a6568a977df3d5d3b89b4/src/libslic3r/GCodeWriter.cpp#L385).

Step 3. requires Step 2. to work properly. I think I could implement all three of them. 

However, I might need some time. Do you think, Step 1. is safe enough to begin with, and we ensure 2. and 3. in a separate Pull Request? Or are you concerned we might not catch other cases where the position is unknown?

XunZhangBambu:
@ziehmon I think we can start from the third step. What needs to be done in the third step is to add detection of the hotbed boundary in the travel_to_xyz function. This should be the safest and simplest modification.

XunZhangBambu:
@ziehmon   Finding a safe position in all circumstances for the timelapse G-code will be quite complex and requires careful consideration.

ziehmon:
> @ziehmon I think we can start from the third step. What needs to be done in the third step is to add detection of the hotbed boundary in the travel_to_xyz function. This should be the safest and simplest modification.

I agree, but one challenge: the `travel_to_xyz` function would check if the hotbed boundary is exceeded **before** gcode for filament change and/or timelapse is injected. At this point, the position is safe for travel, even though the filament change gcode moves the head to an unsafe position. 

Implementing the detection after the `get_last_z_from_gcode` would work, because this happens while processing the injected gcode, but then other moves (like `travel_to_xyz`) could still cause problems.

Do you have any idea how to solve this?

tsmith35:
> Do you have any idea how to solve this?

Does Prusa Slicer have the spiral glitch? How do they handle this?

ziehmon:
> > Do you have any idea how to solve this?
> 
> Does Prusa Slicer have the spiral glitch? How do they handle this?

I don't think so. This is purely an implementation issue caused by user configurable gcode injections. I can think of different approaches to fix this, but of course we're looking for the easiest and most effective one. I am looking forward to the answer from @XunZhangBambu since I might be overlooking something.

ziehmon:
> > @ziehmon I think we can start from the third step. What needs to be done in the third step is to add detection of the hotbed boundary in the travel_to_xyz function. This should be the safest and simplest modification.
> 
> 
> 
> I agree, but one challenge: the `travel_to_xyz` function would check if the hotbed boundary is exceeded **before** gcode for filament change and/or timelapse is injected. When the function `travel_to_xyz` runs, the position would be safe for travel since the injected gcode (timelapse and/or filament change) is unknown at this point. 
> 
> 
> 
> Implementing the detection after the `get_last_z_from_gcode` would work, because this happens while processing the injected gcode, but then other moves (like `travel_to_xyz`) could still cause problems.
> 
> 
> 
> Do you have any idea how to solve this? 

@XunZhangBambu may I ask you to review my comment? This is a fundamental problem from my understanding.

XunZhangBambu:
@ziehmon  Of course !  The injected G-code is provided internally by us, we will ensure that the movements do not exceed the physical limits of the machine. So we only need to check the movements generated by code. By the way, we are also internally exploring ways to ensure that even if the injected G-code is skipped, the correct nozzle position can still be obtained after executing the injected G-code and this will take some time.

ziehmon:
@XunZhangBambu thank you for your reply. I think we misunderstood each other, so let me summarize:

- I raise this PR to change the injected filament change gcode to end in an safe, consistent and expected position (x=0, y=0)
- You suggest we additionally validate/check if the position is safe for a spiral Z hop in the gcode generator (reffered to as "code" in your reply)
- I ask how to implement the validation, since the problematic filament change gcode is injected, not from the generator ("code") itself 
- You suggest we don't validate the injected filament change gcode, but only the gcode from the generator ("code"), which is not the problem

Please excuse me, but that does not make any sense to me.

- In a nutshell, we need to ensure the injected filament change gcode ends in a safe, expected and consistent position (`x=0, y=0`) - just like other injected gcodes. 
- Currently, this does not happen, so this PR changes it to `x=0, y=0` to fix the problem and afterwards merge #4631  

We can build additional validations, but remember: the cause of the out of bounds move om #4631 is the unsafe, unexpected and inconsistent filament change gcode and validating it is very challenging.

ziehmon:
@tsmith35 may I ask for your opinion on this? I think you had a close look at the code too.

ziehmon:
I am also willing to jump on a discord call or whatever you would prefer, myself and a lot of other people would like to see this fixed and are willing to contribute but we need to agree on an approach. 

/cc @lanewei120

lanewei120:
> I am also willing to jump on a discord call or whatever you would prefer, myself and a lot of other people would like to see this fixed and are willing to contribute but we need to agree on an approach.
> 
> /cc @lanewei120

@ziehmon could you give your discord account to our email (lane.wei@bambulab.com)
@XunZhangBambu will contact you by discord

XunZhangBambu:
> @XunZhangBambu thank you for your reply. I think we misunderstood each other, so let me summarize:
> 
> * **I raise this PR** to change the injected filament change gcode to end in an safe, consistent and expected position (`x=0, y=0`)
> * **You suggest we instead validate/check** if the position is safe for a spiral Z hop, and disable it if not
> * **I ask how to implement the validation**, since the position is changed by the injected gcode after the decision is made in the gcode generator
> * **You suggest we don't validate** using the injected gcode position, but only validate using the gcode position from the "code", which contains invalid data since gocde is injected afterwards.
> 
> Please excuse me, but that does not make any sense to me.
> 
> * In a nutshell, we need to ensure the injected filament change gcode ends in a safe, expected and consistent position (`x=0, y=0`) - just like other injected gcodes
>   
>   * If we can't ensure this, the alternative is to validate/check the position before setting the Z hop type to spiral
>   * Implementing this validation/check is currently not possible, due to the injection happening **after** the Z hop type is selected.
> 
> The cause of the out of bounds move on #4631 is the unexpected and inconsistent filament change gcode position and validating it is very challenging since the layer processing is done when injections happen.

@ziehmon  From my perspective, checking the inserted G-code still doesn't seem to solve this issue. Because in the slicing software, you are not certain whether the inserted G-code will definitely be executed, similar to timelapse G-code. I believe that by inserting a secure movement command (that cannot be skipped) at each unskippable part of such G-code, we can ensure that we obtain the correct position of the nozzle after executing the inserted G-code.

ziehmon:
> > @XunZhangBambu thank you for your reply. I think we misunderstood each other, so let me summarize:
> > 
> > * **I raise this PR** to change the injected filament change gcode to end in an safe, consistent and expected position (`x=0, y=0`)
> > * **You suggest we instead validate/check** if the position is safe for a spiral Z hop, and disable it if not
> > * **I ask how to implement the validation**, since the position is changed by the injected gcode after the decision is made in the gcode generator
> > * **You suggest we don't validate** using the injected gcode position, but only validate using the gcode position from the "code", which contains invalid data since gocde is injected afterwards.
> > 
> > Please excuse me, but that does not make any sense to me.
> > 
> > * In a nutshell, we need to ensure the injected filament change gcode ends in a safe, expected and consistent position (`x=0, y=0`) - just like other injected gcodes
> >   
> >   * If we can't ensure this, the alternative is to validate/check the position before setting the Z hop type to spiral
> >   * Implementing this validation/check is currently not possible, due to the injection happening **after** the Z hop type is selected.
> > 
> > The cause of the out of bounds move on #4631 is the unexpected and inconsistent filament change gcode position and validating it is very challenging since the layer processing is done when injections happen.
> 
> @ziehmon From my perspective, checking the inserted G-code still doesn't seem to solve this issue. Because in the slicing software, you are not certain whether the inserted G-code will definitely be executed, similar to timelapse G-code. I believe that by inserting a secure movement command (that cannot be skipped) at each unskippable part of such G-code, we can ensure that we obtain the correct position of the nozzle after executing the inserted G-code.

Hi @XunZhangBambu! Please excuse my late reply, I was on vacation. :sun_behind_small_cloud: 

I really like your proposed solution. It is simple to implement, safe and maintainable. I will implement it in next days. We don't need a discord call then. Testing the change will probably be a lot of work but we have a great solution then. 

Thanks for thinking about this again! :)

tsmith35:
> I really like your proposed solution. It is simple to implement, safe and maintainable. I will implement it in next days. We don't need a discord call then. Testing the change will probably be a lot of work but we have a great solution then.

@ziehmon I have both an A1 (with AMS lite) and A1 mini (also with AMS lite), so that may help! Did you ever get any testing machines?

ziehmon:
@tsmith35 you're faster than light. 

I will definitely come back to your offer for testing with AMS Lite. 

Unfortunately, I didn't received an AMS Llite from Bambu, I assume it wasn't approved which is okay. It was a nice thought at least. Maybe they can test on their print farm though.

tsmith35:
@ziehmon I added a PR #5219 just a few minutes ago. It adds a "move to safe position" to each wipe instance for each of the A1 and A1 mini 0.4 machine profiles. I don't see a downside to returning to X0 after each such wipe instance during filament changes, since not returning to X0 can have unexpected consequences. Will you take a look? Thanks.

SaltWei:
> I am also willing to jump on a discord call or whatever you would prefer, myself and a lot of other people would like to see this fixed and are willing to contribute but we need to agree on an approach.

Hello,  @ziehmon Do you remember when I mentioned that I would help you with AMS Liteor a discount coupon? I sent you an email separately to discuss this matter, but it seems that your GitHub email is unable to receive emails.

SaltWei:
> > I am also willing to jump on a discord call or whatever you would prefer, myself and a lot of other people would like to see this fixed and are willing to contribute but we need to agree on an approach.
> 
> Hello, @ziehmon Do you remember when I mentioned that I would help you with AMS Lite or a discount coupon? I sent you an email separately to discuss this matter, but it seems that your email on github page is unable to receive emails.

tsmith35:
Simon is doing a world of good, so he is quite deserving of some test-bed
equipment. I applaud you for assisting. It's a win-win.

Tom

SaltWei:
Hello @ziehmon . Is this your email ? ziehmon@smtpd.de. 
We (Bambu Lab Marketing Team colleague and me) send a email to you but have no response yet. Could you please check your email and have a look?

ziehmon:
> Hello @ziehmon . Is this your email ? [ziehmon@smtpd.de](mailto:ziehmon@smtpd.de). We (Bambu Lab Marketing Team colleague and me) send a email to you but have no response yet. Could you please check your email and have a look?

Hi @SaltWei! I received the email and just answered it, forgot to put you in CC. Thank you very much for reaching out! :heart_hands:

SaltWei:
> > Hello @ziehmon . Is this your email ? [ziehmon@smtpd.de](mailto:ziehmon@smtpd.de). We (Bambu Lab Marketing Team colleague and me) send a email to you but have no response yet. Could you please check your email and have a look?
> 
> Hi @SaltWei! I received the email and just answered it, forgot to put you in CC. Thank you very much for reaching out! 🫶

My pleasure!

ziehmon:
> > @XunZhangBambu thank you for your reply. I think we misunderstood each other, so let me summarize:
> > 
> > * **I raise this PR** to change the injected filament change gcode to end in an safe, consistent and expected position (`x=0, y=0`)
> > * **You suggest we instead validate/check** if the position is safe for a spiral Z hop, and disable it if not
> > * **I ask how to implement the validation**, since the position is changed by the injected gcode after the decision is made in the gcode generator
> > * **You suggest we don't validate** using the injected gcode position, but only validate using the gcode position from the "code", which contains invalid data since gocde is injected afterwards.
> > 
> > Please excuse me, but that does not make any sense to me.
> > 
> > * In a nutshell, we need to ensure the injected filament change gcode ends in a safe, expected and consistent position (`x=0, y=0`) - just like other injected gcodes
> >   
> >   * If we can't ensure this, the alternative is to validate/check the position before setting the Z hop type to spiral
> >   * Implementing this validation/check is currently not possible, due to the injection happening **after** the Z hop type is selected.
> > 
> > The cause of the out of bounds move on #4631 is the unexpected and inconsistent filament change gcode position and validating it is very challenging since the layer processing is done when injections happen.
> 
> @ziehmon From my perspective, checking the inserted G-code still doesn't seem to solve this issue. Because in the slicing software, you are not certain whether the inserted G-code will definitely be executed, similar to timelapse G-code. I believe that by inserting a secure movement command (that cannot be skipped) at each unskippable part of such G-code, we can ensure that we obtain the correct position of the nozzle after executing the inserted G-code.

Hi @XunZhangBambu! I did implement the change you suggested: after each timelapse gcode injection, we move to the safe position from within the GCode processor. So, no matter what the end position in the `timelapse_gcode` is, we know, the position is safe for a Z-hop. 

Unfortunately, this does not work either. When timelapse is disabled by the user, the move to the safe position remains in the code and that causes unrequired extra movements (and potentially would also cause filament residue).

I highly suggest we stick to the approach I proposed with this PR: we ensure the GCode ends in the safe position `x=0, y=0`.

If you have a better idea, please let me know.

ziehmon:
@XunZhangBambu: here is a screenshot from the change. I removed the `timelapse_gcode` entirely to simulate what happens when the user disables it, and you can see the additional movements to the safe position.

grafik

XunZhangBambu:
@ziehmon 
Thank you very much for continuing to analyze this issue. Recently, I’ve been occupied with other tasks and haven’t had the chance to address this problem yet. I completely understand your thoughts, and it’s possible that my earlier comments were unclear. To ensure we can always accurately determine the current position after each timelapse, what we need to do is add a safe move in the G-code itself, rather than adding the move in the GCodeProcessor.

In simpler terms, the goal is to ensure that the GCodeProcessor always has a clear understanding of the current position under any circumstances. By inserting a fixed move into the G-code, we can break it down into three scenarios:

Timelapse G-code is empty: In this case, the position remains unchanged, so we can accurately determine the current position.
Timelapse G-code is not empty and timelapse is enabled: The fixed move is executed, and the current position is clearly defined.
Timelapse G-code is not empty but timelapse is disabled: The fixed move is still executed, ensuring the current position is clear.

The challenge, however, lies in determining what is an absolutely safe move. As you mentioned in your response, moving to (0, 0) might seem like a potential solution, but can we guarantee that position is always safe? If there is a model located at (0, 0) (in by-object mode), there would be a significant risk of collision.

ziehmon:
> The challenge, however, lies in determining what is an absolutely safe move. As you mentioned in your response, moving to (0, 0) might seem like a potential solution, but can we guarantee that position is always safe? If there is a model located at (0, 0) (in by-object mode), there would be a significant risk of collision.

You are right: (0,0) is not safe. I just tested it. I think we should pick (-10,0), or (x - spiral Z hop radius,0) - but let me check this and I will report back!

I think I understand your idea now. But just to be sure:

> Timelapse G-code is not empty but timelapse is disabled: The fixed move is still executed, ensuring the current position is clear.

But when timelapse is disabled, we will always approach that safe position, and this adds:
- Printing time
- Excess/residual filament at nozzle

ziehmon:
@XunZhangBambu  I solved it. It was a lot of hard work but I found a clean and nice solution. It's totally different from the original proposal, but I integrated all of your feedback! :) I would greatly appreciate if you can take some time to look at it in details and maybe also share it with your colleagues: https://github.com/bambulab/BambuStudio/pull/5736

@SaltWei thanks again for helping me with the AMS Lite. The new PR is because of your assistance! Maybe you can assign @XunZhangBambu as a reviewer since he is familiar with the history. I will close this.

/cc @tsmith35.

tsmith35:
@ziehmon that sounds like a perfect approach. No more doubts about the location of the nozzle when the spiral lift is executed. This is a much-needed enhancement. Great job!

Issue #4933, opened 2024-10-03
Wrongly summarized support filament usage when slicing multiple pla


emil-nasso:
### Bambu Studio Version

1.9.7.52

### Where is the application from?

Bambu Lab Official website

### OS version

Mac

### Printer

A1

### How to reproduce

Sliced all plates of this https://makerworld.com/en/models/676112#profileId-604218

i removed some models from the plates (duplicates/color variants) and sliced all plates to get an idea of total filament usage.

The support filament usage seem to be summarizing the model filament usage values.

### Actual results

970g

### Expected results

A lot less

### Project file & Debug log uploads

<img width="825" alt="image" src="https://github.com/user-attachments/assets/3f342198-ce3f-45d8-85ec-2f74d086b9bb">

### Checklist of files to include

- [x] Log file
- [x] Project file

emil-nasso:
[PUMP2000V1.0.3mf.zip](https://github.com/user-attachments/files/17241431/PUMP2000V1.0.3mf.zip)

DrOetcker:
I can confirm this error as I just wanted to create an Issue for it.

After manual rearrange & Filament change of the official A1 Top Mount Model in Studio, the estimation of the "Slice all" option is wrong. The total calcuation for the Model itself is correct, but the total calculation for the support is wrong and identical to the Model.

Bambu Studio Version: 
1.9.7.52

OS: 
Windows 10 Pro

Printer: 
A1

[AMS+Lite+Top+Mount-Final - Wrong Filament calc on slice all.zip](https://github.com/user-attachments/files/17242767/AMS%2BLite%2BTop%2BMount-Final.-.Wrong.Filament.calc.on.slice.all.zip)


Greetings
Oetcker

![Wrong slice all estimation](https://github.com/user-attachments/assets/f5a5cb58-6473-4cff-80f4-31b160b41193)

LiZ-Li-BBL:
This issue has been fixed internally, please wait for our update.

ttt888zzz:
Please update the newest version，https://github.com/bambulab/BambuStudio/releases
![image](https://github.com/user-attachments/assets/80bdeb9f-98b7-458e-9b37-c0f8b9ae9b57)

PR #4942, opened 2024-10-05
fix issue #4933

shiftrs:
issue seems to have been caused by an incorrect copy-paste. can't build in my current environment but this should fix it


lanewei120:
@shiftrs 
sorry for the late replay

thanks very much for this submission
we found it has been fixed internall before, and already in the beta release

image

PR #5103, opened 2024-10-29
linux.d/debian: use correct string match syntax

dankamongmen:
The `test(1)` command (also known as `[(1)`) does not use the `==` syntax for testing equality, which is a feature of bash's `[[` builtin. String equality is tested with '=', and numeric equality with `-eq`. This eliminates a series of warnings when running `BuildLinux.sh` on Debian-derived systems:

```
[schwarzgerat](0) $ ./BuildLinux.sh -u
./linux.d/debian: line 39: [: ==: unary operator expected
./linux.d/debian: line 39: [: ==: unary operator expected
./linux.d/debian: line 42: [: ==: unary operator expected
```


MackBambu:
Thank you for the suggestion. I need to conduct some tests.

MackBambu (commented):
@lanewei120 "=" has better compatibility and can be merged.

PR #5153, opened 2024-11-06
Add all Neptune 4 printers as supported by Elegoo generic filaments

RyanHir:
When using the Neptune 4 Plus on the 1.10 beta all the Elegoo filaments are missing and we are left with a generic filament. This PR adds all the Neptune 4 family of printers as supported by the generic Elegoo filaments.


QingZhangBambu (approved):
looks good to me. and it passed base check test

PR #5219, opened 2024-11-13
Added a move to safe position to A1 and A1 mini change filament gcode

tsmith35:
Previous versions of these two files did not return the nozzle to a safe position after performing a wipe, leaving the nozzle positioned outside of the printable area after finishing. This can cause problems if the nozzle subsequently attempts to move in a direction that may exceed the physical limits of the machine.


ziehmon:
Hey Tom, so this PR is basically an improved version of #4848. I see you spotted even more situation where we shall move to the safe position. Nice find!

Bambu suggested moving the nozzle to the save position inside the gcode processor, so it's not done by injection (as in our PRs) but rather "hard-coded".  I think, we should do both after looking at your gcode modifications actually: keep the safe position in the injected gcode AND ensure it's the final position after the injection ends. 

I will reach out in this PR again when I have draft for the "hard-coding", and we can then merge both I'd propose :)

tsmith35:
> Bambu suggested moving the nozzle to the save position inside the gcode processor, so it's not done by injection (as in our PRs) but rather "hard-coded". I think, we should do both after looking at your gcode modifications actually: keep the safe position in the injected gcode AND ensure it's the final position after the injection ends.

Yes, I believe it to be the safest approach when custom gcode injection is permitted; custom gcode may be unpredictable, therefore the final step of moving the head to a safe position following any such custom gcode will only help improve operational reliability. Trust, but verify.

tsmith35:
@XunZhangBambu does everything look okay to merge, or do I need to make any other changes? This is to prepare for the spiral lift modifications, so it would be much easier to implement that once these changes are in place. Just let me know if there is anything else that you need for me to do on this PR to get it ready. Thank you.

tsmith35:
@lanewei120 is this a reasonable PR? I haven't been able to get it reviewed yet, though the changes would provide safer return positioning for additional movement. Thank you.

XunZhangBambu:
@tsmith35 Sorry for the late reply.I’ve been busy with other things recently. From your PR, I noticed that you added G1 X0 to the change filament G-code, but this could cause a collision if a model is placed at the x=0 position. Especially in by-object printing mode, finding a safe position is indeed a challenge.

tsmith35:
@ziehmon has found a better way to avoid potential unsafe locations while using spiral lifts, so the intended purpose of this PR is unnecessary. :-)

PR #5228, opened 2024-11-13
Update BambuStudio_tr.po

fatih5228:
Update Turkish language


lanewei120:
Hello @fatih5228
we just released 1.10.1 

could you resubmit this patch based on the newest codes?

thanks very much

PR #5286, opened 2024-11-15
fix: Bambu P1P 0.4mm profile added missing MC fan gcode

thopedam:
I added the missing MC fan control to the machine start gcode for the Bambu P1P 0.4mm nozzle.
This is an exact copy of the machines 0.2, 0.6 and 0.8mm nozzle machine start code.

This is needed if you upgraded the P1P to P1S, because internally the printer is still a P1P. This way the Filament Profiles can be correctly mapped, when using the AMS.


lanewei120:
close it due to patch too old and not valid now

PR #5328, opened 2024-11-18
Fixed tooltip for cool plate initial layer temp.

ShaneQi:
See below screenshot. When hover cursor on the Cool Plate (not supertack)'s initial layer print temp setting, the tooltip is showing "Bambo Cool Plate Supertack" which is incorrect.

This PR fixes this tooltip verbiage.

bambu-bug


QingZhangBambu (approved):
looks good to me

PR #5347, opened 2024-11-20
Remove domain restrictions on URL protocol imports

isycat:
When importing a file via bambustudio://open/?file=bla, the URL is now allowed even if it does not match makerworld, amazonaws.com or aliyuncs.com

### DIFF
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -1092,12 +1092,9 @@ void GUI_App::post_init()
             std::string download_url;
 #if BBL_RELEASE_TO_PUBLIC
             for (auto input_str : input_str_arr) {
-                if (boost::starts_with(input_str, "http://makerworld") ||
-                    boost::starts_with(input_str, "https://makerworld") ||
-                    boost::starts_with(input_str, "http://public-cdn.bblmw.com") ||
-                    boost::starts_with(input_str, "https://public-cdn.bblmw.com") ||
-                    boost::algorithm::contains(input_str, "amazonaws.com") ||
-                    boost::algorithm::contains(input_str, "aliyuncs.com")) {
+                if (boost::starts_with(input_str, "http://") ||
+                    boost::starts_with(input_str, "https://") ||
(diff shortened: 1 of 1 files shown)


lanewei120:
@isycat thanks for the patch

I am afraid we can not merge it for the security reason

Without restricting the source of the URL, anyone could use the URL to pass in something like a Trojan or other types of malware.

lanewei120:
Not everyone will carefully read these warning messages, 
and even if they do, if they end up clicking "Import" and getting infected by a virus, Bambustudio would still be held responsible.

In fact, what we initially wanted was the way you’ve modified it in this patch
image

isycat:
Currently, you can just host the malware on a - random s3 bucket or literally any domain or sub domain starting with "makerworld". Examples:
`evilbucket.s3-website-us-east-1.amazonaws.com/malware.stl`
`makerworldevil.com/malware.stl`
`makerworld.evil.example.com/malware.stl`

If this really is intended to help the user it should be togglable, as suggested above, and check for whitelisted certificates instead of a partial URL check. 

Anything short of that is effectively just annoying gatekeeping. Let's not pretend it's a security feature.

lanewei120:
yes, you are right, above urls also have risks:)
so I think we should add more checks for the urls to improve the security

In fact, this security risk is already reported by our customer before on the the previous version using "http://" and "https://"

lanewei120:
@NanashiTheNameless
 thanks for the suggestion

@isycat 
we have a fast discuss internally,
and we suggest to do as follows:

keep the original url judge logic currently, let it go on as before(we will refine more these address check later)
and popup a warning dialog when the url doesn't meet the above
and if user clicks ok, bambu studio will go on to load it

could you do the modifications?

isycat:
This is a perfect idea, thanks! 
I am not sure when/if I will find the time to make the change, so if someone else wants to pick it up, please do.

chromozonechris:
hello I need a website that gives adware please

Floppy:
I'd like to add my support to this ticket, having just opened a feature request related to it at #6120.

PR #5419, opened 2024-11-27
Add LDO as a filament vendor

camerony:
Add LDO as a filament vendor


lanewei120:
@camerony 
hello
could you give a link to the filament from LDO?

camerony:
Yes,  it’s
https://ldomotion.com/filament

Issue #2731, opened 2023-10-14
Linux Dark Mode Process pane object names illegible


keithel:
### Bambu Studio Version

1.7.7.89
1.7.6.92

### Where is the application from?

Bambu Lab github releases

### OS version

Ubuntu 23.04 Lunar Lobster

### Additional system information

CPU: 13th Gen Intel(R) Core(TM) i7-13700H
Memory: 64GB
Graphics Card: RTX A1000 6GB Laptop GPU

### Printer

Bambu Lab P1P

### How to reproduce

1. Open Gnome settings
2. Click Appearance tab
3. Choose `Dark` style
4. Color is set to the default Ubuntu installs with - the leftmost choice of orange.
5. Open Bambu Studio
6. Load any project with objects.
7. In the left bottom `Process` pane, in the `Global/Objects` mode toggle, switch to `Objects` properties mode.
8. The bug should present itself.

### Actual results

After following all steps, the names of the plates and names of the objects are all in a font color that is extremely difficult to read on the dark grey background. The font color is just slightly different than the dark mode background color. The checkboxes for the object and assembly visibility in the second column are a different shade of grey, a little easier to see than the object, assembly, and plate names, but is also far too dark.

### Expected results

After following all reproduction steps, the plate, assembly, and object names in the `Process` pane should be rendered in a legible font color. I think a lighter shade of grey would probably be best, but that is up to you. All that really needs to happen is that it has enough contrast on the background to be legible.

![image](https://github.com/bambulab/BambuStudio/assets/156712/7b9699a9-0da9-43ee-9201-bd750782a170)

### Project file & Debug log uploads

Bambu Studio Project file I was using:
[Tesla J1772 Holder short.zip](https://github.com/bambulab/BambuStudio/files/12907873/Tesla.J1772.Holder.short.zip)

Log file in this case I think is unnecessary, but I include here for completeness:
[debug_Sat_Oct_14_15_19_34_411014.log.0](https://github.com/bambulab/BambuStudio/files/12907887/debug_Sat_Oct_14_15_19_34_411014.log.0.txt)

### Checklist of files to include

- [X] Log file
- [X] Project file

keithel:
Workaround for this is to just use light mode. Unfortunately, the Linux release of Bambu Studio does not contain a dark/light mode toggle in the app so that Bambu Studio can be in light mode while the rest of your desktop is in dark mode. That issue was reported as #1357.

Cass67:
[...]

PR #5487, opened 2024-12-07
Fix illegible color in 'unsaved changes dialog' on Linux

D0ot:
Some fonts are illegible in light theme on Linux:
image

The foreground color `wxSystemSettings::GetColour(wxSYS_COLOUR_INFOTEXT)` varies on different platform while the background color is hard coded, which causes the problem. Just hard code the foreground color to fix it.

After the fix:
image
image


D0ot:
Fix some problem mentioned in #2731

PR #5521, opened 2024-12-12
Tiny English translation update

ping-localhost:
Capatizle `Enable` in `Enable multi instance rendering by OpenGL`, so it matches the rest of the text.

image

PR #5559, opened 2024-12-16
Tiny English translation update (now in the proper file)

ping-localhost:
As requested https://github.com/bambulab/BambuStudio/pull/5521#issuecomment-2544226332


lanewei120:
Hi @ping-localhost 
no need to change this now, we have updated it in local code base, and will update to github soon

PR #5581, opened 2024-12-19
i18n: added special case for pt_BR folder naming on Unix

### LINKED ISSUE #5580: Brazilian Portuguese translations not properly loaded on Linux
### Bambu Studio Version

1.10.1.50

Built from source by yourself or others

### OS version

Debian 12

OS: Debian GNU/Linux 12 (bookworm) x86_64 
Kernel: 6.1.0-28-amd64 
Shell: bash 5.2.15 
DE: Plasma 5.27.5 
WM: KWin 
Theme: Breeze [Plasma], Adwaita-dark [GTK2/3] 
Icons: [Plasma], breeze-dark [GTK2/3] 
CPU: AMD Ryzen 9 5900X (24) @ 4.200GHz 
GPU: AMD ATI Radeon RX 6800/6800 XT / 6900 XT 
Memory: 6039MiB / 32000MiB 

Note: Installed from Flathub - basically the AppImage rebundled. The same behavior happens on .AppImage from Github releases.

### Printer

Bambulab A1

### How to reproduce

1. Install BambuStudio on Linux via Flatpak/Download the AppImage and make executable.
2. Switch language to Brazilian Portuguese
3. After reload, the a few words on the first page will be in Portuguese, the remainder will be in English.

This does not seems to be the case on Windows. This may not be the case if the Linux filesystem is case insensitive.

### Actual results

The program is not translated.

### Expected results

The program is expected to be translated to Portuguese.

I could not find any info in the debug output, sadly, but I could diagnose the issue and have a "workaround":

After the flatpak is installed, you can go to the app folder and look for the share/i18n subdirectory. There are two folders there, one named 'pt-BR' and another named 'pt_br'. I managed to get it translated by creating a symbolic link pointing to 'pt-BR', named 'pt_BR'. I believe that the folder name should follow the pt_BR naming convention as it would properly be recognized when loaded on Linux/Windows, instead of the 'pt_br' with the country code in lower case.

I can try to submit a patch however I'm not sure where. The simplest way would be to just rename the source folder, but I'm not sure if this will break windows builds.

- [ ] Log file
- [ ] Project file

ronoaldo:
Similar to what is done to support zh_CN, but for pt_BR.

Fixes #5580

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -262,6 +262,10 @@ else ()
             COMMAND ln -sfn "zh_cn" "${BIN_RESOURCES_DIR}/i18n/zh_CN"
             COMMENT "Symlinking zh_CN language setting to zh_cn"
             VERBATIM)
+        add_custom_command(TARGET BambuStudio POST_BUILD
+            COMMAND ln -sfn "pt-BR" "${BIN_RESOURCES_DIR}/i18n/pt_BR"
+            COMMENT "Symlinking pt_BR language setting to pt_BR"
+            VERBATIM)
     endif()
 endif ()
 


MackBambu (approved):

MackBambu:
@lanewei120  It can be merged.

Issue #4076, opened 2024-05-15
BambuStudio forgets Printer's password when disconnected


adras:
### Bambu Studio Version

1.8.2.56

### Where is the application from?

Bambu Lab Official website, Bambu Lab github releases

### OS version

Windows 11

### Printer

Bambu Lab X1 Carbon

### How to reproduce

1. Connect your lan mode only printer in bambu studio
2. Turn the printer off, wait until BambuStudio notices that it's disconnected
3. Click on the printer to connect to it again (it's still turned off)

### Actual results

4. After the "Connect 3DP-00M-185 failed! [SN:00M09A360201185, code=-1]" message BambuStudio forgot the printer's password

### Expected results

BambuStudio should not forget the printer's password just because the printer is turned off and a connection is attempted

### Project file & Debug log uploads

File size too big: 25 MB are allowed, 37 MB were attempted to upload.

### Checklist of files to include

- [ ] Log file
- [ ] Project file

walterwongbbl:
Hi. @adras 
Please use our latest version 1.9.1. This issue has been optimized. Thank you.

adras:
I just tested it in 1.9.1.66 where the issue still exists. Is there more information about this?

walterwongbbl:
> Author

Hi. You need to uninstall the old version first and delete the configuration file at the same time. Do not use overlay installation.

mrtwebdesign:
The issue is "optimized"?

I'm on Bambu Studio 1.9.2.57.

I have problems with BS forgetting my P1S password without either the printer or the app being restarted. It feels like it's on a 12 hour timeout.

Where is the configuration file on Macos?
Also, wouldn't this wipe out all of my customizations? Filament profiles? The nuke-it-from-orbit approach feels like overkill. For now I just keep the Bambu password where I can copy-paste it into Bambu studio. I may see if I can automate that process.

I switched to LAN only the last time Bambu cloud services went down for a few hours. I had nothing but time on my hands to find a way to print without the cloud. Other than the inconvenience of Bambu Studio "forgetting" my P1S password, I don't see a reason to go back for my use case.

walterwongbbl:
Hi @mrtwebdesign 
I'm very sorry for replying to you so late, you can access the configuration file directory by opening it in the studio menu.
Have you encountered this problem recently.

patrick-at-qreis-dot-com:
I can confirm the issue is still present.

[...]

PR #5673, opened 2025-01-02
Quick fix attempt for machine password erased on disconnect (issue #4076).

crashkopf:
This patch attempts to keep the printer access code from being cleared erroneously on a lost or failed connection.  This only applies to LAN mode.

Notes:
  - The setting is 'access_code'.  There is also a setting named 'user_access_code', but it's function is unclear and may be vestigal.
  - Some parts of the network connection behavior are handled by the networking plugin.  I can't seem to make the machine list refresh from outside.


walterwongbbl:
Hi. @crashkopf 
I checked your request, and everything looks good. Also, it helps avoid resending the access code when the connection is lost. Thanks a lot, we'll go ahead and use it!

lanewei120:
thanks for doing this

PR #5674, opened 2025-01-02
Added Linux Mint support to the build system.

crashkopf:
Created symlinks for `linuxmint` and `ubuntu` in `linux.d`, and removed the Ubuntu check in `BuildLinux.sh`.  This is an easy way to handle derivative distributions.

One particular caveat: symlinks don't play well
with older versions of git on Windows.  It's unlikely to be a problem unless somebody tries to commit changes to `linux.d` from windows. See: https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows

An alternative approach is to use the `ID_LIKE` tag in `/etc/os-release`.  See example 3 in the `os-release` manual.


MackBambu:
In fact, we primarily work on Windows and often need to modify files in the Linux.d directory on Windows.

crashkopf:
Fair.  I'm going to send another PR for alternative approach.

louhy:
> Fair. I'm going to send another PR for alternative approach.

Still I'm glad you did this - I may try setting this up to build on Mint so it might come in useful until there's a more permanent fix.

crashkopf:
> Still I'm glad you did this - I may try setting this up to build on Mint so it might come in useful until there's a more permanent fix.

I'm pleased to say my other [pull request](https://github.com/bambulab/BambuStudio/pull/5686) for this issue got approved.  Feedback is welcome!

PR #5686, opened 2025-01-03
Change Linux build script to recognize more supported distributions

crashkopf:
This changes the way the Linux build script recognizes which distribution script might be compatible with the host OS.  It does so by checking both the `ID` and `ID_LIKE` tags in `/etc/os-release` against available scripts in `linux.d`.  First match is used.

Note that ID_LIKE isn't a 100% compatability guarantee, but it gives the user a fighting chance that the script will work for their distribution, and should reduce the need for manual addition of more distribution targets.


MackBambu (approved):
@crashkopf Thank you for your contribution..
@lanewei120 It can be merged.

lanewei120:
thanks for doing this

Issue #4201, opened 2024-06-03
[Regression]  Printer / Filament / Process dropdown menus do not display


jmickelin:
### Bambu Studio Version

1.9.1.66

### Where is the application from?

Bambu Lab github releases

### OS version

OS: Fedora 38

### Additional system information

Window manager: i3 4.22

### Printer

Any printer

### How to reproduce

Using Bambu_Studio_linux_fedora-v01.09.01.66.AppImage.

Example:

1. Go to the "Prepare" tab
2. Click the "Plate type" dropdown list

But it happens with _any_ dropdown lists in the entire program. Including the ones in dialog windows and the ones for "Slice plate" and "Print plate" in the top bar.

It _does not_ happen with the "menu bar" dropdowns (the hamburger "File" menu and the ∨-button immediately to the right of it).

In very rare cases I _have_ managed to get certain dropdown list to open properly, namely the infill and pattern dropdowns under the "Strength" tab. There does not seem to be any consistent setup to trigger it (just clicking around a bunch in the program, opening and closing some of the preference dialogs, moving the window to different workspaces on my desktop). Once it starts working, these dropdown lists (and only those!) seem to work until I restart the program. I have not managed to trigger this intermittently working behavior with any other dropdown list.

### Actual results

Dropdown list briefly flashes on screen for a fraction of a second, then disappears.

### Expected results

Dropdown list should stay open and allow you to select an option.

### Project file & Debug log uploads

[debug_Mon_Jun_03_14_24_06_34357.log.0.txt](https://github.com/user-attachments/files/15533995/debug_Mon_Jun_03_14_24_06_34357.log.0.txt)

### Checklist of files to include

- [X] Log file
- [ ] Project file

jmickelin:
This was originally reported downstream in OrcaSlicer: https://github.com/SoftFever/OrcaSlicer/issues/2101

Other users with different combinations of distro versions and desktop environments report the same issue.

The issue was _not_ present in BambuStudio 1.4.5, as per the findings I reported in this comment, hence I have marked this as a regression in the title of this bug report: https://github.com/SoftFever/OrcaSlicer/issues/2101#issuecomment-2041036989

DanBao-Bambu:
I am very sorry for the inconvenience caused to you, and the issue has been recorded.

smcgu:
Reverting the changes and compiling resolved this bug for me in 1.9.7.52.

smcgu:
[...]

PR #5704, opened 2025-01-06
Fix PopupWindow dropdown menus do not display #4201 

gw0:
Fixes https://github.com/bambulab/BambuStudio/issues/4201 that makes Bambu Studio unusable under some window managers, but reintroduces inconvenience of PopupWindow remaining in front after alt+tab.

### DIFF
--- a/src/slic3r/GUI/Widgets/PopupWindow.cpp
+++ b/src/slic3r/GUI/Widgets/PopupWindow.cpp
@@ -16,7 +16,7 @@ bool PopupWindow::Create(wxWindow *parent, int style)
     if (!wxPopupTransientWindow::Create(parent, style))
         return false;
 #ifdef __WXGTK__
-    GetTopParent(parent)->Bind(wxEVT_ACTIVATE, &PopupWindow::topWindowActiavate, this);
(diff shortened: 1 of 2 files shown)


bambu123:
if (!event.GetActive() && IsShown()) DismissAndNotify();

Deleting this code may cause the popup to remain on the desktop, even if the main window has been switched to the background.

gw0:
> Deleting this code may cause the popup to remain on the desktop, even if the main window has been switched to the background.

Yes, this does have this consequence. But it is better to have an annoyance when pressing alt+tab than have a unusable app. Removing this line resolves the regression #4201 that makes Bambu Studio unusable.

I am not a GTK expert, but maybe the proper solution for the alt+tab issue would be to use one of the GTK window type hints instead of dismissing popup windows on events (what the problematic line does): https://docs.gtk.org/gdk3/enum.WindowTypeHint.html

bambu123:
Thank you!
OK, Please help merge @lane.wei.

lanewei120:
thanks for doing this

back2ll:
Thank you!

PR #5721, opened 2025-01-08
Update appimagetool link

Drsheppard01:
It helps to build AppImage for any linux include musl-based
This pr also raises the question of the need for additional builds for ubuntu, ubuntu-latest and fedora

### DIFF
--- a/src/platform/unix/build_appimage.sh.in
+++ b/src/platform/unix/build_appimage.sh.in
@@ -1,5 +1,5 @@
 #!/bin/sh
-APPIMAGETOOLURL="https://github.com/AppImage/AppImageKit/releases/latest/download/appimagetool-x86_64.AppImage"
+APPIMAGETOOLURL="https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage"
 
 
 APP_IMAGE="@SLIC3R_APP_KEY@_ubu64.AppImage"


MackBambu:
This does not appear to be a stable version. What issues might be resolved by using this version of the AppImage build tool?

Drsheppard01:
Firstly, this solves the libfuse problem, (on ubuntu only fuse3 is shipped by default, while the appimage kpc only work with version 2)
Secondly, it will potentially make it work for systems on musl libc, but that's not certain

Samueru-sama:
> This does not appear to be a stable version. What issues might be resolved by using this version of the AppImage build tool?

Hi, that appimagetool is stable and uses the static appimage runtime from [here](https://github.com/AppImage/type2-runtime). 

This removes the libfuse2 dependency from the AppImage, which is an EOL library that no longer comes on new distros by default. 

-----------------------------------------------------

Since the runtime is static is also allows to make AppImages that work on any linux system, [like this one](https://github.com/probonopd/PrusaSlicer/releases/tag/continuous). 

I noticed right now you have 3 different appimage releases, likely due to the webkit2gtk incompatibility issue, the only way to fix that issue is to bundle webkit2gtk in the AppImage along with opengl since the lib needs a specific version of opengl to work otherwise you will have a different white screen issue.

MackBambu:
@Drsheppard01 Thank you for your submission. I will will more tests on the new packaging tool.


> I noticed right now you have 3 different appimage releases, likely due to the webkit2gtk incompatibility issue, the only way to fix that issue is to bundle webkit2gtk in the AppImage along with opengl since the lib needs a specific version of opengl to work otherwise you will have a different white screen issue.

@Samueru-sama 
Is there a similar project that packages webkit2gtk and OpenGL for reference? I've been stuck on this for a long time.

Samueru-sama:
> Is there a similar project that packages webkit2gtk and OpenGL for reference? I've been stuck on this for a long time.

The very AppImage of PrusaSlicer that probonopd made that I just linked xd

[steps](https://github.com/probonopd/PrusaSlicer/blob/continuous/.github/workflows/build_appimage.yml#L77-L205)

Note that the tool that's being used, which is [sharun](https://github.com/VHSgunzo/sharun) recently got something called hooks that do a lot of those steps automatically, I made the PR to add the [webkit2gtk hook](https://github.com/VHSgunzo/sharun/pull/5) to avoid doing that was done originally with the Prusa appimage of doing it all manually.

Samueru-sama:
@MackBambu the deployment script finally got the refactor it needed, and [now the steps are these](https://github.com/probonopd/PrusaSlicer.AppImage/blob/main/bundle.sh)

MackBambu (approved):

lanewei120:
thanks

Issue #5614, opened 2024-12-25
Error when exporting G-Code file


imba-LALA:
### Bambu Studio Version

1.10.1.50

### Where is the application from?

Bambu Lab Official website

### OS version

Windows11 23H2

### Printer

Anycubic Kobra Plus 0.4 nozzle

### How to reproduce

1.Import Model
2.Slice Plate
3.Export G-code file

### Actual results

The pop-up window displays:
Failed processing of the filename_format template.
Parsing error at line 1: Not a variable name
{input_filename_base}_{filament_type[initial_tool]}_{print_time}.gcode

![Bug](https://github.com/user-attachments/

### Expected results

Output G-code file.

### Project file & Debug log uploads

This problem occurs regardless of the model used.

### Checklist of files to include

- [ ] Log file
- [ ] Project file

imba-LALA:
![Bug](https://github.com/user-attachments/assets/1b13a041-545e-449e-9cd9-556d2f5402d3)

imba-LALA:
![Uploading Bug.png…]()

imba-LALA:
![Bug](https://github.com/user-attachments/assets/3d148866-8f92-446d-9ce6-18125a0fc428)

kkeonline:
I have submit a pull request 
https://github.com/bambulab/BambuStudio/pull/5734

XunZhangBambu:
@kkeonline  This has been fixed in our latest verison.

Issue #5291, opened 2024-11-15
Failed processing of file name format


Zoch707:
### Bambu Studio Version

1.10.1.50

### Where is the application from?

Bambu Lab Official website

### OS version

macOS Sonoma 14.2.1

### Printer

Creality Ender 3 V3 KE

### How to reproduce

slice file and try to export g-code. 

### Actual results

the little window pops up and says failed prossesing of file name format ...

### Expected results

it should ask me where to save the .gcode file and proceed to save it there.  

### Project file & Debug log uploads

[Auto_TestGit.3mf.zip](https://github.com/user-attachments/files/17777913/Auto_TestGit.3mf.zip)

### Checklist of files to include

- [ ] Log file
- [X] Project file

dcs3473:
Ditto.  This has happened to me also.  It has to do with this: Gcode Output: File Format:{input_filename_base}_{filament_type[initial_tool]}_{print_time}.gcode coming from the Orca Profile, I think.  It looks like it was copied over and has leftovers that are not supported in BambuStudio

XunZhangBambu:
@Zoch707 We did not include initial_tool as a placeholder internally. To avoid affecting your export, you can change the format to: {input_filename_base}_{filament_type[0]}_{print_time}.gcode, which will use the first selected material type as the material type in the exported file.

dcs3473:
Found it.  Need to turn on Dev Mode to see the missing field.  And that worked.  Thanks.

kkeonline:
Just try Bambu studio with Creality Ender 3 V3 KE and got the same error.
![2025-01-09_214342](https://github.com/user-attachments/assets/52b8f0b8-04e7-45f9-aa88-ede0b28caee0)
checking the json files all ender model are wrong.

PR #5734, opened 2025-01-09
Fix: Failed processing of the filename_format template for Creality Ender 3 Printer

kkeonline:
Fix Failed processing of the filename_format template for Creality Ender 3 Printers 
Ender 3 V3 
Ender 3 V3 KE
Ender 3 V3 Plus
Ender K1C

https://github.com/bambulab/BambuStudio/issues/5614
https://github.com/bambulab/BambuStudio/issues/5291


kkeonline:
I see that the latest beta does not error with the template 
So this should not be fix.

PR #5736, opened 2025-01-09
Fix stringing: support spiral lift/Z-hop with timelapse gcode injections

ziehmon:
This is the successor of https://github.com/bambulab/BambuStudio/pull/4848 and https://github.com/bambulab/BambuStudio/pull/4631. For a detailed explanation why injection of gcode for timelapse causes stringing, please see both PRs.

## Summary

This PR improves the injection of timelapse gcode so that a lift/Z-hop operation is executed **before** the timelapse gcode is injected. When the print head returns from the timelapse macro position, no spiral Z-hop is generated in the slicer anymore - this was causing collisions. After thinking about the process (for a long time), I noticed that we do not actually require two Z-hops.

I created a diagram to show the differences.

## Previous Implementation

grafik

## Proposed Implementation

grafik

## Details

To use only one Z-hop, we need to immediately execute one **before** the timelapse gcode is injected. With the current implementation (`lift`), this is not possible, since these operations are "scheduled"/"queued" to be executed at a later point (when using `travel_to_xyz()`).  

This PR **adds the new function** `travel_lift` which immediately executed a lift operation to a given `z` height and flags existing scheduled/queued lift operations to be done (` m_lifted = delta_z; m_to_lift = 0;`). 

Furthermore, it removes the second Z-hop from the `timelapse_gcode` in the printer profile, since we now can rely it was executed.

Essentially, this means that no Z-hop will be executed **within or after the timelapse gcode** and thus, no collisions can occur because of an unknown position:

grafik

## Testing

Since I have an AMS Lite now (thanks BambuLab and thanks @SaltWei!) I was able to test all different injection scenarios/positions. With and without AMS Lite, with/without purge/wiper tower. However I assume it would be great if BambuLab test on their end too. 

Happy for your feedback!


ziehmon:
> @ziehmon Awesome work, especially the diagram, very clear. Change the implementation of lift from lazy lift to instant lift seems to be able to solve the problem. Could you give us some time for thorough testing? We are currently busy with other matters.

Thanks you for the flowers :bouquet:!  Take your time on testing and let me know if and how I could help.

XunZhangBambu:
@ziehmon Hello, we are currently incorporating the contents of this PR into our codebase and testing it internally. It seems to address this long-standing issue. If the tests pass, would you be open to us submitting the code internally in your name, rather than merging the PR directly from GitHub?

ziehmon:
> @ziehmon Hello, we are currently incorporating the contents of this PR into our codebase and testing it internally. It seems to address this long-standing issue. If the tests pass, would you be open to us submitting the code internally in your name, rather than merging the PR directly from GitHub?

Hi @XunZhangBambu, even though this is not my most appreciated workflow, if it works for you (and you need to do it), feel free.

oscareliacim:
How can I test this point to see if it solves the chafing problem, greetings, I have Windows 11

ziehmon:
@lanewei120 @XunZhangBambu any news?

XunZhangBambu:
@ziehmon  Sorry for the delay response. We have implement the eager lift logic in our 2.0 version. You can try our latest version and see if it works

ziehmon:
Closed, as change was integrated and released upstream by BambuLab.

Issue #5171, opened 2024-11-09
Problems building Bambu Studio


adaviel:
### Bambu Studio Version

01.09.07.52

### Where is the application from?

Bambu Lab github releases

### OS version

Fedora 41 6.11.5-300.fc41.x86_64

### Printer

A1 mini

### How to reproduce

$ unzip BambuStudio-01.09.07.52.zip
$ cd BambuStudio-01.09.07.52
optionally edit BuildLinux.sh to set BUILD_ARGS="-DDEP_WX_GTK3=ON -DJPEG_VERSION=6"
$ ./BuildLinux.sh -dr |tee BuildLinux2.log

### Actual results

[135/180] Performing patch step for 'dep_OCCT'
FAILED: dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch /home/andrew/bambu/BambuStudio-01.09.07.52/deps/build/dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch 
cd /home/andrew/bambu/BambuStudio-01.09.07.52/deps/build/dep_OCCT-prefix/src/dep_OCCT && /usr/bin/git apply --directory deps/build/dep_OCCT-prefix/src/dep_OCCT --verbose --ignore-space-change --whitespace=fix /home/andrew/bambu/BambuStudio-01.09.07.52/deps/OCCT/0001-OCCT-fix.patch && /usr/bin/cmake -E touch /home/andrew/bambu/BambuStudio-01.09.07.52/deps/build/dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch
Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt...
error: deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt: No such file or directory
Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx...
error: deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx: No such file or directory
ninja: build stopped: subcommand failed.

### Expected results

I expect to get a working BambuStudio.

[...]

Issue #4689, opened 2024-08-26
`--directory` parameter in cmake files for OCCT and OpenCV cause build to fail in Fedora 40


9strands:
### Bambu Studio Version

01.09.04.60

### Where is the application from?

Built from source by yourself or others

### OS version

Fedora 40

### Additional system information

x86_64

### Printer

Bambu Lab A1

### How to reproduce

Attempt to build from source with "./BuildLinux.sh -d" 

### Actual results

[3/49] Performing patch step for 'dep_OCCT'
FAILED: dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch /home/user/projects/BambuStudio/deps/build/dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch 
cd /home/user/projects/BambuStudio/deps/build/dep_OCCT-prefix/src/dep_OCCT && /usr/bin/git apply --directory deps/build/dep_OCCT-prefix/src/dep_OCCT --verbose --ignore-space-change --whitespace=fix /home/user/projects/BambuStudio/deps/OCCT/0001-OCCT-fix.patch && /usr/bin/cmake -E touch /home/user/projects/BambuStudio/deps/build/dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch
Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt...
error: deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt: No such file or directory
Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx...
error: deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx: No such file or directory
[4/49] Performing install step for 'dep_OpenCSG'
[0/1] Install the project...
-- Install configuration: "Release"

-- Build files have been written to: /home/user/projects/BambuStudio/deps/build
[1/41] Performing patch step for 'dep_OCCT'
FAILED: dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch /home/user/projects/BambuStudio/deps/build/dep_OCCT-prefix/src/dep_OCCT-stamp/dep_OCCT-patch 
[...]

Issue #4158, opened 2024-05-26
occt cant build


echo-lxq:
### Bambu Studio Version

last version

### Where is the application from?

Bambu Lab github releases

### OS version

window 10

### Additional system information

![图片](https://github.com/bambulab/BambuStudio/assets/79234901/d8238433-7cad-418b-b63a-bb53ee2ab733)

### Printer

x1 series

### How to reproduce

building the deps

### Actual results

已启动生成…
1>------ 已启动生成: 项目: dep_FREETYPE, 配置: Debug x64 ------
2>------ 已启动生成: 项目: dep_OCCT, 配置: Debug x64 ------
2>Performing patch step for 'dep_OCCT'
2>Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt...
2>CUSTOMBUILD : error : deps/build/dep_OCCT-prefix/src/dep_OCCT/CMakeLists.txt: No such file or directory
2>Checking patch deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx...
2>CUSTOMBUILD : error : deps/build/dep_OCCT-prefix/src/dep_OCCT/src/Font/Font_FTFont.cxx: No such file or directory
2>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241,5): error MSB8066: “D:\work\Projects\BambuStudio\deps\build\CMakeFiles\433f5f2e4fd2cfbe4738a5a96451353e\dep_OCCT-patch.rule;D:\work\Projects\BambuStudio\deps\build\CMakeFiles\433f5f2e4fd2cfbe4738a5a96451353e\dep_OCCT-configure.rule;D:\work\Projects\BambuStudio\deps\build\CMakeFiles\433f5f2e4fd2cfbe4738a5a96451353e\dep_OCCT-build.rule;D:\work\Projects\BambuStudio\deps\build\CMakeFiles\433f5f2e4fd2cfbe4738a5a96451353e\dep_OCCT-install.rule;D:\work\Projects\BambuStudio\deps\build\CMakeFiles\68c521bc805cc530064eef33e47b39d2\dep_OCCT-complete.rule;D:\work\Projects\BambuStudio\deps\build\CMakeFiles\258fc9b4ab6c35b7f3020ecde013b882\dep_OCCT.rule”的自定义生成已退出，代码为 1。
[...]

PR #5966, opened 2025-02-17
Fix patch step for OCCT and OpenCV dependencies

jpalazz2:
The `PATCH_COMMAND` for OCCT and OpenCV are trying to patch files that don't exist because of the addition of the `--directory` flag. `PATCH_COMMAND` runs in the working directory of the downloaded source files per my own testing. Specifying `--directory` here is causing the patch to try to apply to `dep_OpenCV-prefix/src/OpenCV` from root of the downloaded source rather than from the root of the downloaded source code. This causes `git apply` to try to patch files that don't exist which returns nonzero error codes (I can only reproduce this behavior of `git apply` when the `--directory` flag is present, without it the patch fails silently). I have a comment going into a bit more depth about this [here](https://github.com/bambulab/BambuStudio/issues/4689#issuecomment-2662200508).

This is leading to the build failures and workaround being discussed in #4158, #4689, #5171.

With these changes I'm able to fully cleanbuild on Arch Linux.


MackBambu:
@jpalazz2
Removing the --directory parameter works fine on Fedora and other Unix-like systems, but it does not work properly on Windows. You can check the github action logs . Can you make it compatible with Windows ?

PR #6155, opened 2025-03-25
Build fixes for 2.0.0

hadess:
As needed for the Flatpak build, see https://github.com/flathub/com.bambulab.BambuStudio/pull/56


lanewei120:
thank you~

PR #6314, opened 2025-03-30
fix: typo "pringing"

pfeerick:
The text for "Changed fan speed during printing may affect print quality, please choose carefully." has a typo - "pringing" instead of "printing". 

I simply did a find and replace for all instances of the typo, not sure if more is needed.


lanewei120:
@pfeerick 
thanks for doing this

could you only modify the original text please?
we will generate the translation files locally

pfeerick:
Do I drop the change to `bbl/i18n/BambuStudio.pot` also?

lanewei120:
thanks for doing this

PR #6373, opened 2025-04-01
Fix Brim paint gizmo memory leaking

Noisyfox:
Fix `render_hover_point` memory leaking by using `std::optional`


MackBambu (approved):

MackBambu:
@Noisyfox  Thanks, this is an great improvement.

@lanewei120 I think it can be merged.

PR #6468, opened 2025-04-08
Add ability to modify the print flow ratio per object

RyanHir:
This functionality is from OrcaSlicer and is used for changing the flow rate of each object on a build plate. 

This functionality is useful to me as to allow for custom flow calibration prints to be performed. Particularly, this allows for the following project on Maker World to work natively on Bambu Studio, without needing Orca Slicer. https://makerworld.com/en/models/189543-improved-flow-ratio-calibration-v3

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -5478,9 +5478,7 @@ std::string GCode::_extrude(const ExtrusionPath &path, std::string description, 
         gcode += m_writer.set_jerk_xy(jerk);
     }
     // calculate extrusion length per distance unit
-    auto _mm3_per_mm = path.mm3_per_mm * double(m_curr_print->calib_mode() == CalibMode::Calib_Flow_Rate ? this->config().print_flow_ratio.value : 1);
-
-    // calculate extrusion length per distance unit
+    auto _mm3_per_mm = path.mm3_per_mm * double(this->config().print_flow_ratio.value);
     if( path.role() == erTopSolidInfill )
         _mm3_per_mm *= m_config.top_solid_infill_flow_ratio.value;
     else if (this->on_first_layer())
--- a/src/slic3r/GUI/ConfigManipulation.cpp
+++ b/src/slic3r/GUI/ConfigManipulation.cpp
@@ -727,6 +727,7 @@ void ConfigManipulation::toggle_print_fff_options(DynamicPrintConfig *config, in
         toggle_line(el, has_overhang_speed, variant_index);
 
     toggle_line("flush_into_objects", !is_global_config);
+    toggle_line("print_flow_ratio", !is_global_config);
 
     toggle_line("support_interface_not_for_body",config->opt_int("support_interface_filament")&&!config->opt_int("support_filament"));
 
--- a/src/slic3r/GUI/Tab.cpp
+++ b/src/slic3r/GUI/Tab.cpp
@@ -2191,6 +2191,7 @@ void TabPrint::build()
         optgroup->append_single_option_line("is_infill_first","parameter/quality-advance-settings");
         optgroup->append_single_option_line("bridge_flow","parameter/bridge");
         optgroup->append_single_option_line("thick_bridges","parameter/bridge");
+        optgroup->append_single_option_line("print_flow_ratio");
         optgroup->append_single_option_line("top_solid_infill_flow_ratio","parameter/quality-advance-settings");
         optgroup->append_single_option_line("initial_layer_flow_ratio","parameter/quality-advance-settings");
         optgroup->append_single_option_line("top_one_wall_type","parameter/quality-advance-settings");


magyoshaa:
usefull .. hope they do this for release version .. the more options the better

RyanHir:
@magyoshaa Were you ever able to test this feature addition? I recall you wanting a download link about a week ago.

magyoshaa:
> @magyoshaa Were you ever able to test this feature addition? I recall you wanting a download link about a week ago.

i tried it in the slicer. and looked useful.
however was unable to test print on my H2D as it complained with an error about authentication . presume need to have printer in dev mode maybe ?
image

RyanHir:
> however was unable to test print on my H2D as it complained with an error about authentication . presume need to have printer in lan/dev mode maybe ?
I believe this to be due to the new Authorization Control System which requires the slicer to be signed by Bambu for the network plugin to work for the H2D/X1 in cloud mode.

RyanHir:
@zhimin-zeng-bambulab Is there anything that is needed from me to proceed with review? There are a few people I know in a nearby lab that would benefit from this feature being in Bambu Studio.

magyoshaa:
a great pity they do not seem to be actioning this change. its essential for calibration . 
i recently had to find the correct flow ratio for sunlu high speed petg, with standard numbers of 0.95 it was terrible, with stock bambu studio it was nearly imposible to calibrate. using this small change i was able to establish flow ratio needed to be 0.87 and now my prints are perfect.

logikal:
Has anyone been able to use the builds from this PR to print to an H2D that's in Developer Mode? My H2D never shows up for me in dev mode, unless I'm using the officially released and signed build.

I'm on MacOS 15.3.2

RyanHir:
I have had issues with LAN mode for custom builds, but cloud still works for the devices that still allow for it. Unfortunately for the H2D you have no option of a firmware build that doesn't have the authorization control. I believe the root cause to be with the Bambu network plugin but that is outside the scope of this PR. For reference all my testing has been on MacOS 15.

For now I would recommend bringing the 3MF/gcode files to the printer manually with a flash drive.

RyanHir:
@lanewei120 @zhaofengli Is it possible to have this PR reviewed? It has been sitting idle for over a month and it is a fairly simple change.

logikal:
> For now I would recommend bringing the 3MF/gcode files to the printer manually with a flash drive.

Are you testing this PR by transferring files via USB stick?

Thanks for the suggestion, I agree the rest of my line of questioning is out of scope of this PR.

zhaofengli:
@RyanHir I'm just a fellow owner, not related to Bambu. This does look useful though!

logikal:
I noticed that if you have multiple objects with individual flow ratio, then merge them (to e.g. optimize "print by object" on a bunch of small objects), the individual objects lose their flow ratio.

Not sure if this is intended behavior or not.

magyoshaa:
> Has anyone been able to use the builds from this PR to print to an H2D that's in Developer Mode? My H2D never shows up for me in dev mode, unless I'm using the officially released and signed build.
> 
> I'm on MacOS 15.3.2

yes .. works perfectly for me . 
i do find in lan/developer mode that BS seems not to remember the printer so you have to reconnect it each time. that is a sepetate issue however and this mergbe request/change works perfectly

magyoshaa:
> @lanewei120 @zhaofengli Is it possible to have this PR reviewed? It has been sitting idle for over a month and it is a fairly simple change.

@zhaofengli yes please review ! this is essential for correct flow ratio calibrations

RyanHir:
> Are you testing this PR by transferring files via USB stick?

I only have a P1S so my testing has been with cloud mode on FW 01.07.00.00, which does not have the authorization control.

RyanHir:
> > I noticed that if you have multiple objects with individual flow ratio, then merge them (to e.g. optimize "print by object" on a bunch of small objects), the individual objects lose their flow ratio.
> > Not sure if this is intended behavior or not.
> 
> if you merge them surely they cease being individual objects , therefore the individual objects become one and the flow per object setting would therefore be lost.

I would agree that they become a single object, so the flow per object option gets lost. Oddly, the flow option is present in the per part selection within the assembly, even though the option is no-op. It might be best to find a way to hide this setting. It should be noted that the other flow options already in the slicer also appear to be broken when editing each objects flow setting in an assembly.

magyoshaa:
> > Are you testing this PR by transferring files via USB stick?
> 
> I only have a P1S so my testing has been with cloud mode on FW 01.07.00.00, which does not have the authorization control.

I have tested on a H2D in lan/dev mode and it works perfectly

magyoshaa:
if there not going to approve this change perhaps they could tell us how we are meant to calibrate flow ratio correctly. Bambu's own brand filaments my well work well without change however many other need drastic changes eg sunlu petg-hs matte needed tobe calibrated to get a number of 0.87 and so does esun asa+ . there appears to be no functional way to get these figures and auto calibration does not work well

helto4real:
Please merge this, we really need better ways to calibrate flow ration on third party filaments!

lanewei120:
Hello, if the software supports flow calibration that allows users to specify their own model to replace the default one, would that address your issue?

Since object-level flow coefficients are mainly useful for calibration and not for regular printing,
we are hesitant to expose too many parameters to users, as it would increase the complexity of the parameter interface.

DoesN0tCompute:
The ideal solution for this is to put this feature behind a checkbox in configuration so "regular" users will not see it unless its enabled.

RyanHir:
> Hello, if the software supports flow calibration that allows users to specify their own model to replace the default one, would that address your issue?
> 
> Since object-level flow coefficients are mainly useful for calibration and not for regular printing, we are hesitant to expose too many parameters to users, as it would increase the complexity of the parameter interface.

Hi @lanewei120, Thank you for getting back to us.

I can understand the hesitancy for exposing new parameters in the UI, but the UI would still be needed to make new calibration models. Without it exposed, it would require external software to be used to generate the 3MF for Bambu Studio. 

As I understand, this UI option would only be exposed in Developer mode which most users would/should not need to enable. I feel like it is good to keep this exposed as a UI option but only for users with Developer mode enabled.

zhimin-zeng-bambulab:
image

@RyanHir  What is the purpose of this modification? Is it unnecessary?

RyanHir:
Hi @zhimin-zeng-bambulab,

That is to hide the option from the global flags, and to make this PR purely a per-object flag. I believe is that if the global flow settings need to be changed it can be done in the filament profile. 

I suppose it can be useful for some niche tasks such as combat robotics that require overextrusion on all parts for better layer adhesion. This could allow for the settings to be placed in a custom print profile that can be used for all filaments, instead of just one modified filament.

RyanHir:
After further thought, I think it might not be a bad idea to remove this check. The option is already a developer mode specific option, so this option being in the global list can be used to define specialized profiles that require overextrusion. But I will leave it here for now in case you decide it should remain a per-object only flag.

zhimin-zeng-bambulab:
image
 
@RyanHir  We hope to uniformly manage the display and hiding of parameters in this function (ConfigManipulation::toggle_print_fff_options). Could you please modify it as shown in the figure?

RyanHir:
I have reverted the custom check and added the check in the place you requested.

lanewei120:
Hello @RyanHir 
thanks for this work
We will merge this PR soon

Could you reduce the commit count to 1
it seems too much commits in this pr now~

RyanHir:
Hi @lanewei120,

I have went ahead and squashed the commits down to one.

lanewei120:
@RyanHir 

done, thanks

RyanHir:
Thank you!

PR #6471, opened 2025-04-08
FIX: recognition with correct ID: SUNLU PETG @BBL A1M.json

addius8:
Changed inherits from X1C to regular SUNLU PETG base and adapted the A1 preset 1 to 1, works now in Bambu Studio!


XunZhangBambu:
@addius8  Thanks for your commit., but it has been fixed in out latest version. I will close the PR

PR #6639, opened 2025-04-22
Add fusrock abs gf filament

FusRock:
Add fusrock abs gf filament


hotellonely:
Though I just print it with BBL abs-gf profile :p

FusRock:
> Though I just print it with BBL abs-gf profile :p虽然我只是用 BBL abs-gf profile 打印它:p

Hah

FusRock:
进度如何了？是否存在什么问题？ @lanewei120

PR #6854, opened 2025-05-07
Small error in message (file download)

eried:
When it fails to download a file, the msgbox says "file upload failed" instead of "file download"


FiredEgg-bambu:
Hi @eried, this is not a translation issue. It looks like our developers may have used the same source key for different dialogs. We will fix this as soon as possible, and this PR is unlikely to be merged.

Issue #6120, opened 2025-03-19
Allow URL handler to open files from sites other than makerworld.com


Floppy:
Many 3d model hosting sites exist, other than makerworld.com. Bambu Studio has a URL handler for `bambustudio:` URLs, allowing websites to tell Bambu Studio to open a file from provided URL. This is very useful functionality.

However, currently the URLs are rejected if they aren't from `makerworld.com`:

https://github.com/bambulab/BambuStudio/blob/4b48ad974389c3c20dfd605e9ddba0324787d27f/src/slic3r/GUI/GUI_App.cpp#L1095

It would be much better for the community, and allow the use of more models from more sources, to remove this restriction, or perhaps allow it to be disabled in an advanced setting somewhere.

(Full disclosure: I'm the author of https://manyfold.app, and the ability to open files directly into Bambu Studio (and others) is a long-requested feature)

Floppy:
In fact, I see there's already a PR at https://github.com/bambulab/BambuStudio/pull/5347.

rstocks:
Setting Manyfold aside, seems a reasonable request to allow you to open models from anywhere.

Issue #3798, opened 2024-03-30
Project with long project and plate name failed after sending to printer


DPattee:
### Bambu Studio Version

1.8.4.51

### Where is the application from?

Bambu Lab Official website

### OS version

macOS 13.6.3

### Printer

Bambu X1C

### How to reproduce

Open a project from MakerWorld that has multiple plates like https://makerworld.com/en/models/97518#profileId-104159
Use a Mac and save it as a new project name with space and several words like "My Gridfinity Battery Setup" .
Import a few other models so you have exactly the set of battery holders you need, deleting/adding plates as necessary.
I didn't rename the original Plate 1 which was "AAx25 2x2x3 - magnet holes, with lip"

Slice it, send it to the printer, it uploads to the cloud, it switches to web cam view, the bed starts heating, and then an error says invalid file name. 

The autogenerated file name based on "My Gridfinity Battery Setup - AAx25 2x2x3 - magnet holes, with lip" seems to have been either too long or too something. 

### Actual results

After the file was sent to the cloud, and the app switched to webcam view, an error popped up saying unsupported file name and printing was cancelled

### Expected results

Print to succeed

### Project file & Debug log uploads

[My Gridfinity Battery Setup.3mf.zip](https://github.com/bambulab/BambuStudio/files/14812510/My.Gridfinity.Battery.Setup.3mf.zip)
[Archive.zip](https://github.com/bambulab/BambuStudio/files/14812516/Archive.zip)

### Checklist of files to include

- [X] Log file
- [X] Project file

walterwongbbl:
@DPattee 
We will fix this issue

BambulabRobot:
This issue has been marked as inactive due to no response for 90 days.

PR #6856, opened 2025-05-08
Allow URL handler to open files from untrusted sites with confirm

LightDestory:
Essentially this PR is a follow up of #5347. It addresses #5347 #6120 #3798

I am not going to talk about the vague "security" approach currently deployed for the URL handler, improve that is out of scope of this PR.

The changes directly implements the decision made by @lanewei120 

> keep the original url judge logic currently, let it go on as before(we will refine more these address check later)
and popup a warning dialog when the url doesn't meet the above
and if user clicks ok, bambu studio will go on to load it

I am currently having issue building on my PC due to strange deps error (W11, every requirement installed, clean clone of repository), so I am using this PR as a github-action build check.

The changes are minimal.

### DIFF
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -1100,6 +1100,7 @@ void GUI_App::post_init()
 
             std::string download_url;
 #if BBL_RELEASE_TO_PUBLIC
+			USHORT ext_url_open_state = -1; // -1 not set, wxNO not open, wxYES open
             for (auto input_str : input_str_arr) {
                 if (boost::starts_with(input_str, "http://makerworld") ||
                     boost::starts_with(input_str, "https://makerworld") ||
@@ -1108,6 +1109,15 @@ void GUI_App::post_init()
                     boost::algorithm::contains(input_str, "amazonaws.com") ||
                     boost::algorithm::contains(input_str, "aliyuncs.com")) {
                     download_url = input_str;
+                }
+                else {
+                    if (ext_url_open_state == -1) {
+                        wxString askMsg = wxString::Format(_L("This file is not from a trusted site, do you want to open it anyway?"));
+                        ext_url_open_state = wxMessageBox(askMsg, "Bambu Studio", wxYES_NO | wxICON_EXCLAMATION);
+                    }
+                    if (ext_url_open_state == wxYES) {
+                        download_url = input_str;
+                    }
                 }
             }
 #else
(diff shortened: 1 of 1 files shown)


walterwongbbl:
Hi. @LightDestory 
USHORT ext_url_open_state = -1; // -1 not set, wxNO not open, wxYES open
Do you expect there to be a switch here to enable or disable subsequent operations? Use int or short here. Please confirm.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -1100,6 +1100,7 @@ void GUI_App::post_init()
 
             std::string download_url;
 #if BBL_RELEASE_TO_PUBLIC
+			USHORT ext_url_open_state = -1; // -1 not set, wxNO not open, wxYES open
```
walterwongbbl, line 1103: Please confirm the variable type.

LightDestory:
Thank you for your comment.
`wxYes` and `wxNo` are macros defined in the `defs.h` file and represent integer values “2” and “8” respectively.
`wxMessageBox` returns an integer value. I declared an unsigned short to use only 2 bytes instead of 4 (int). However, per your request I am going to change `ext_url_open_state` to `short`.

> Do you expect there to be a switch here to enable or disable subsequent operations? Use int or short here. Please confirm.

I am following @lanewei120 direction of unalter the existing logic. Currently:

1. You generate an array of str by splitting based on "file" separator;
2. Check every string of such array and if the "security check" is successful you update the `download_url` variable
	1. A malformed reuqest, with more than one "file=" GET params will evaluate only the last GET params

The confirm request is performed for every bambustudio: urlhandler execution because I am not saving the user preference anywhere. It was planned this way.

If this PR is going to get merged I will work on a persistent user preference for this behavior.

src/slic3r/GUI/GUI_App.cpp
LightDestory: Look at my reply. I am performing an implicit conversion int to short.

LightDestory:
Please allow the github workflow to be performed because I am unable to compile locally. (I would appreciate if you could provide me support to get rid of the compiling issue).

I am confident that such changes should not introduce any bug because I am introducing the display of a MessageBox with the evaluation of its return value.

But I would prefer to actually check if  working like excepted.

walterwongbbl:
Hi. @LightDestory 
We should consider whether to add a new option in the preferences. Thank you very much for your contribution.

LightDestory:
Hi, @walterwongbbl 
Is my explanation good enough? Is the PR ready to merge and no further changes are needed?

This PR do not aim to add a new option in the preference because it was not suggested by @lanewei120 at that time.

lanewei120:
thanks for doing this

PR #6942, opened 2025-05-16
feat(url handler): add User Preference for 3rd party model sites

LightDestory:
This PR is a direct follow up of #6856.

- Introducing a User Preference to allow the import of 3d models from 3rd party websites.
- Unified the behavior between the BBL_RELEASE_TO_PUBLIC and self-compiled versions.

The changes are minimal and aims to improve the user consistency between the different variants of BBS.


LightDestory:
Updated OP to add the `split_str` bug

PR #7084, opened 2025-06-03
Small typo in FanControl.cpp

eried:
Small typo fixed but...

image

The message itself is still quite weird `Heating mode is suitable for printing ABS/ASA/PC/PA materials and circulates filters the chamber air.` I am not sure if is references that it will recirculate and filter the air?, maybe this is more clear:

`Heating mode is suitable for printing ABS/ASA/PC/PA materials, and it circulates and filters the chamber air.`


lanewei120:
thanks

PR #7506, opened 2025-07-07
Fix spanish translation for "Ludicrous"

rastersoft:
The Spanish .po file translates "Ludicrous" as "Lúdico", which is absolutely erroneous. The correct translation is "Absurda", as it was done when Spaceballs was dubbed both in Spain and Latin American ("Ludicrous speed" -> "Velocidad absurda").

Of course, since, in Spanish, words have gender, all the other speeds must be swapped from masculine to feminine to keep the gender matching.


FiredEgg-bambu:
@rastersoft Thank you for pointing that out. We have corrected the error on our internal translation platform, so this pull request will not be merged. Thank you.
image

PR #7556, opened 2025-07-13
Update SelectMachine.cpp

BeanBagKing:
Rather than returning just 'unknown', we can pass the string of the nozzle type. At a minimum this helps the end user just as much as 'unknown'. At best it helps everyone involved figure out what kind of nozzle is causing issues when future nozzles are introduced.


XinZhangBambu:
Please check
image

PR #6765, opened 2025-04-30
Show Calibration menu for BBL printers

logikal:
Addresses #5791.

This really should just be the default. I can imagine it maybe being behind the Developer Mode flag in the settings, but let's just do it here, since while Bambu's auto-calibration is _fine_, some of us like to _really_ dial it in.

PR #7593, opened 2025-07-17
Show calibration menu for multi-extruder bbl printers

logikal:
I'm happy you've integrated [my previous contribution](ttps://github.com/bambulab/BambuStudio/pull/6765) to allow Bambu Lab customers to more finely manual tune their filaments on most of your line of printers. That's a win for the user experience, in my opinion.

Screenshot 2025-07-17 at 09 26 44

But the _entire point_ of https://github.com/bambulab/BambuStudio/pull/6765 was to enable the calibration menu for H2D printers, since Bambu's choices made it impossible to finely tune filament with other slicers.

The choice to exclude the H2D feels user-hostile, without some sort of technical reason why you'd make the user experience more confusing. It's weird and confusing that I can see the menu in Bambu Studio when connected to my X1C, but it disappears when I choose my H2D.

Yes, there are many workaround to get the calibrations to run on my H2D. 

_**None of them should be necessary**_.

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -3491,9 +3491,8 @@ void MainFrame::update_calibration_button_status()
 {
     Preset &printer_preset = wxGetApp().preset_bundle->printers.get_edited_preset();
     bool isBBL = printer_preset.is_bbl_vendor_preset(wxGetApp().preset_bundle);
-    bool is_multi_extruder = wxGetApp().preset_bundle->get_printer_extruder_count() > 1;
     // Show calibration Menu for BBL printers if Develop Mode is on.
-    bool show_calibration = (!isBBL || wxGetApp().app_config->get("developer_mode") == "true") && !is_multi_extruder;
+    bool show_calibration = (!isBBL || wxGetApp().app_config->get("developer_mode") == "true")
     wxGetApp().mainframe->show_calibration_button(show_calibration);
 }
 


DarkmoonThomas:
Why was the H2D excluded from this VERY MUCH NEEDED fix in the latest beta?!

magyoshaa:
another disappointed user here... was very much a needed feature .. at least explain why its not there for us

shan916:
What’s the technical reasons the H2D can’t support this?

remi-gelinas:
+1, I was looking forward to not having to jump through hoops to calibrate further on my H2D :/

racinrandall:
I would really like to have this feature to be able to finely tune my filament on my H2D.  I use a different slicer for all of my other printers, but since I'm currently locked in with Bambu Studio to use the H2D, I want the ability to tune the filament I use, not just use the filament that Bambu sends profiles for.

macwhiz:
The lack of H2D support needs explanation, because it seems impossible to justify. Releasing the software without calibration-menu support for the H2D would put Bambu at risk of another round of negative publicity, as users would rightly ask: why not, if not for lock-in to Bambu filament?

Or better yet, support needs to be implemented.

logikal:
Relates to https://github.com/bambulab/BambuStudio/issues/7595

zhimin-zeng-bambulab:
Because some parameters are related to the extruder or nozzle type, we cannot simply reuse the original logic. We are also considering how to support H2D and will solve this problem as soon as possible.

DarkmoonThomas:
> Because some parameters are related to the extruder or nozzle type, we cannot simply reuse the original logic. We are also considering how to support H2D and will solve this problem as soon as possible.

May we ask exactly what logic does not work? I've been using them all, with a work around and have found no issues at all on my H2D. Some clarification would be appreciated.

zhimin-zeng-bambulab:
Before slicing, it is not specified which extruder the filament is located in. When the filament is located in different extruders, some parameters need to be modified accordingly.

logikal:
> We are also considering how to support H2D and will solve this problem as soon as possible.

That's great news @zhimin-zeng-bambulab, and thank you for confirming this will eventually be enabled for the H2D!

Which parameters are necessary? What are the symptoms of them not being available? I haven't exhaustively tested the calibrations using both nozzles, or different nozzle sizes or types.

I'm happy to try to help work on this parameter problem to bring the menu to the H2D.

hotellonely:
> Because some parameters are related to the extruder or nozzle type, we cannot simply reuse the original logic. We are also considering how to support H2D and will solve this problem as soon as possible.

Zhimin saves the day again

logikal:
@zhimin-zeng-bambulab I believe this has been enabled in recent versions of studio, yes? If you can confirm that's intentional and likely to stay, I'll close this PR.

Thanks!

zhimin-zeng-bambulab:
@logikal  Yes, it has been enabled.

logikal:
Thanks so much @zhimin-zeng-bambulab!

Issue #7017, opened 2025-05-24
Wont launch due to libOSMesa.so.8 update / dependencies. Garuda linux


Supersudo2000:
### Bambu Studio Version

bambustudio-bin-02.00.03.54-1

### Where is the application from?

Built from source by yourself or others

### OS version

Garuda Linux 250308 | 6.14.7-zen2-1-zen

### Additional system information

AMD Ryzen 9 5900X
NVIDIA GeForce RTX 3080
Memory:  31.24 GiB

### Printer

Bambu lab A1

### How to reproduce

Launch Bambustudio from terminal or application directly with mesa being upgraded
pacman -Q | grep mesa
lib32-mesa 1:25.1.1-1
lib32-mesa-demos 9.0.0-3
mesa 1:25.1.1-1
mesa-demos 9.0.0-6
mesa-utils 9.0.0-6

### Actual results

$ bambu-studio 
/opt/bambustudio-bin/bin/bambu-studio: error while loading shared libraries: libOSMesa.so.8: cannot open shared object file: No such file or directory
Downgrading mesa works, but program stutters like crazy

### Expected results

program launches

### Project file & Debug log uploads

 bambu-studio 
/opt/bambustudio-bin/bin/bambu-studio: error while loading shared libraries: libOSMesa.so.8: cannot open shared object file: No such file or directory

### Checklist of files to include

- [ ] Log file
- [ ] Project file

madmalkav:
Already covered by issue 6951.

c2h5oh:
> Please update the libOSMesa library

It's a part of Mesa, not BambuStudio. Latest Mesa release (25.1) has removed it completely after it has been deprecated for quite a while.

ShiroBlank:
I had success temporarily moving over to the flatpak version + nvidia workaround until this is resolved.

notkaramel:
Got the same problem on Arch Linux after updating all packages. I downgraded `mesa` to version 1:**24.3.4-1** and BambuStudio seems to work again.

Downgrade program: https://aur.archlinux.org/packages/downgrade

```sh
sudo downgrade mesa
```

Quaditz:
> Got the same problem on Arch Linux after updating all packages. I downgraded `mesa` to version 1:**24.3.4-1** and BambuStudio seems to work again.
> 
> Downgrade program: https://aur.archlinux.org/packages/downgrade
> 
> sudo downgrade mesa

[...]

Issue #6951, opened 2025-05-16
Mesa 21.3+ change effects Bambu Studio


YoSTEALTH:
### Bambu Studio Version

2.0.3.54

### Where is the application from?

Built from source by yourself or others

### OS version

Linux 6.14.6-2-cachyos (64-bit)

### Additional system information

X11

### Printer

A1

### How to reproduce

With `Mesa 25.1.0` installed, launching `bambu-studio` will say `/usr/lib/libOSMesa.so.8` not found.

According to https://docs.mesa3d.org/amber.html "After Mesa 21.3, all non-Gallium DRI drivers were removed from the Mesa source-tree."

My current solution was to uninstall `mesa` and install `mesa-amber` for legacy support.

Bambu Studio will need to add mesa 21.3 version check and account for new changes.

### Actual results

none

### Expected results

none

### Project file & Debug log uploads

none

### Checklist of files to include

- [ ] Log file
- [ ] Project file

jipiboily:
Having the same issue!

Bambu printing on Linux has been a pretty bad experience.

Hope this can be fixed quickly!

jipiboily:
The flatpack version seems to work, btw!

MackBambu:
Thanks for the reminder ， I’ll try adding some additional prompts.

c2h5oh:
Mesa 25.1 removes osmesa completely - this will have to be fixed sooner than later.

fuzunspm:
same issue on arch linux after latest mesa update, flatpak version works but this should be fixed

princess-wawa:
// copied because i posted it on a duplicated issue

basically, libOSMesa.so.8 is no longer included in the Mesa package since version 1:25.1.0-1
see the merge request: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836

this means that libOSMesa, which bambustudio relies on, has just been removed from the Mesa package for no apparent reason other than "redundency"

note that my graphical interface is gnome, and since it has been updated to use the new Mesa version, downgrading won't work, and the mesa-amber doesn't work with it either

the fix is fortunately very simple, either switch to the flatpak version since it still uses the old mesa version that includes osmesa
or replace your mesa package with the amber branch since it still has the osmesa lib

fuzunspm:
This is not a fix but a temporary workaround.

Knifa:
`mesa-amber` is missing `zink` for the NVIDIA workaround #4669 so the only option is to downgrade in that case.

[...]

PR #7899, opened 2025-08-15
FIX: Fix empty Prepare/Preview content in Flatpak

hadess:
By removing the OSMesa dependency.

The "offscreen" rendering code was supposed to use OSMesa's offscreen
buffers so as to avoid an X11 dependency, but the CLI::run() function
called XInitThreads() every time, so there's very little chance that
this ever worked.

Remove the dependency so that we don't depend on functionality that's
been removed from newer mesa:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836

Closes: https://github.com/bambulab/BambuStudio/issues/7017
Closes: https://github.com/bambulab/BambuStudio/issues/6951

See also https://github.com/SoftFever/OrcaSlicer/pull/9708

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -44,7 +44,6 @@ RUN apt-get update && apt-get install  -y \
     libgtk-3-dev libwebkit2gtk-4.0-dev \
     libsoup2.4-dev \
     libgstreamer1.0-dev libgstreamer-plugins-good1.0-dev libgstreamer-plugins-base1.0-dev libgstreamerd-3-dev \
-    libosmesa6-dev \
     libssl3 libssl-dev libcurl4-openssl-dev libsecret-1-dev \
     libudev-dev \
     curl \
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,6 @@ RUN apt-get update && apt-get install  -y \
     libgstreamer-plugins-good1.0-dev \
     libgtk-3-dev \
     libgtk-3-dev \
-    libosmesa6-dev \
     libsecret-1-dev \
     libsoup2.4-dev \
     libssl3 \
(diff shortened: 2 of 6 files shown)


hadess:
Looks like you still want to support some quite old distributions, let me know which parts of this MR you can merge, and I'll split it off.

hadess:
> Looks like you still want to support some quite old distributions, let me know which parts of this MR you can merge, and I'll split it off.

I've removed the X11 related bits.

hadess:
> I've removed the X11 related bits.

And I reworded the commit message.

hadess:
Not sure why @GuDuJian-J-Zhang's comment was removed, but building GLEW without EGL support is what causes the Prepare/Preview content not to show up.

But the problem is ultimately that wxWidgets is built with EGL support, tries to use EGL support and that GLFW isn't built with EGL support.

I'll close this, as both patches still need work, and I've opened this one to explicitly take care of the OpenGL/EGL mismatch in the Flatpak:
https://github.com/flathub/com.bambulab.BambuStudio/pull/69

guanyun-gudujian:
> Not sure why @GuDuJian-J-Zhang's comment was removed, but building GLEW without EGL support is what causes the Prepare/Preview content not to show up.
> 
> But the problem is ultimately that wxWidgets is built with EGL support, tries to use EGL support and that GLFW isn't built with EGL support.
> 
> I'll close this, as both patches still need work, and I've opened this one to explicitly take care of the OpenGL/EGL mismatch in the Flatpak: [flathub/com.bambulab.BambuStudio#69](https://github.com/flathub/com.bambulab.BambuStudio/pull/69)

GuDuJian-J-Zhang is my personal github id
and guanyun-gudujian is my offical id at BambuLab

thant's it

PR #8124, opened 2025-09-07
Fix crash when retrieving timelapse list 

### LINKED ISSUE #8079: Crash when retrieving timelapse list from printer
### Bambu Studio Version

2.2.1.60

Bambu Lab github releases

### OS version

Ubuntu 24.04.2 LTS

CPU: 11th Gen Intel® Core™ i7-1165G7 × 8
memory: 16 GiB

### Printer

Bambu Lab A1 (no AMS)

### How to reproduce

1. Turn on printer
2. Start Bambu Studio (SSL certificate popup: dont remember, Yes)
3. Go to Devices, Store, Timelapses
 Check attached video

### Actual results

App crashes

### Expected results

To be able to nagivate through the list of timelapses

[bambu-studio_crash-log.zip](https://github.com/user-attachments/files/22095640/bambu-studio_crash-log.zip)

https://github.com/user-attachments/assets/e7918c69-c4e8-41e8-b8c0-4b8df489ae4b

- [x] Log file
- [ ] Project file

bytedream:
Fixes #8079. When a thumbnail is loaded, the call to `wxImage` segfaults. In my debug builds the message "Wrong JPEG library version: library is 62, caller expects 80" got printed before the segfault. So it seems that wxImage called jpeg functions from the jpeg package that opencv built (which is always libjpeg version 6.2), and not from the jpeg dependency.
I've changed the cmake file to disable the jpeg usage in opencv.

### DIFF
--- a/deps/OpenCV/OpenCV.cmake
+++ b/deps/OpenCV/OpenCV.cmake
@@ -53,6 +53,7 @@ bambustudio_add_cmake_project(OpenCV
        -DWITH_OPENJPEG=OFF
        -DWITH_QUIRC=OFF
        -DWITH_VTK=OFF
+       -DWITH_JPEG=OFF
        -DWITH_WEBP=OFF
        -DENABLE_PRECOMPILED_HEADERS=OFF
        -DINSTALL_TESTS=OFF


lanewei120:
@bytedream 
seems  the log is related with jpeg version
how about only disable jpeg when building opencv?

we uses png from opencv somewhere

bytedream:
This works too, done.
As far as I can tell, the only time opencv loads an image is in `SkipPartCanvas::LoadPickImage`, which always is an png as you said.

lanewei120:
thanks

PR #8289, opened 2025-09-24
Update BambuStudio_tr.po

fatih5228:
Update Turkish language pls


lanewei120:
hello @fatih5228 
we have released the new version
could you update this file based on the new version?

we will merge it as soon as you updated

thanks

PR #8471, opened 2025-10-19
Fix typos in translations and original English

kunningKing11:


lanewei120:
Hi @kunningKing11 

You need to change the texts in the source file, not modify the translation files directly

thanks

kunningKing11:
where is the source file?

lanewei120:
you can search the texts in the src directory
for this patch, the original text is in src/libslic3r/PrintConfig.cpp

kunningKing11:
thanks

PR #8555, opened 2025-10-27
Include stdint.h for uint32_t definition

3vi1:
Fixes compilation failure on Linux (tested with gcc-14) due to missing uint32_t definition.


lanewei120:
thanks

PR #8680, opened 2025-11-04
Fix filament sorting by vendor and type in AMSMaterialsSetting

ansonl:
`_collect_filament_info()` in `AMSMaterialsSetting.cpp` was assigning both filament vendor and type to `query_filament_vendors` map so that the filament was only being sorted by type which led to a case where Bambu options where shown before and after other vendors' options.

Old incorrectly sorted

image

Corrected sorting screenshot - it's not entirely alphabetically because Bambu has promoted certain vendor+type combinations to the top

image


XinZhangBambu:
@ansonl Thank you for your support。we will review it

XinZhangBambu:
@ansonl  After reviewing the changes, the last three are actually modifications without practical significance. If that's the case, it's recommended that you only submit a single commit.

ansonl:
@XinZhangBambu Sure, I have removed the last 3 commits so only the first commit remains.

XinZhangBambu (approved):
Nice

Issue #8710, opened 2025-11-07
BuildLinux.sh fails on fedora


nicvagn:
### Bambu Studio Version

1.9

### Where is the application from?

Built from source by yourself or others

### OS version

Fedora 43

### Additional system information

The mesa-libOSMesa-devel package was removed from Fedora 43

### Printer

Bamboo H2D

### Printer Firmware Version

NA

### How to reproduce

./BuildLinux.sh -u
OS distribution is 'fedora'.  Using package dependencies for 'fedora'.
Updating and loading repositories:
Repositories loaded.
 https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-43.noarch.rpm                                      100% |  16.7 KiB/s |  11.3 KiB |  00m01s
Package "rpmfusion-free-release-43-1.noarch" is already installed.

Nothing to do.
Updating and loading repositories:
Repositories loaded.
Nothing to do.
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
No match for argument: mesa-libOSMesa-devel
Package "wget2-wget-2.2.0-6.fc43.x86_64" is already installed.
You can try to add to command line:

No match for argument: mesa-libOSMesa-devel

### Actual results

does not build deps

### Expected results

builds deps

### Project file & Debug log uploads

NA

### Checklist of files to include

- [ ] Log file
- [ ] Project file

nicvagn:
openvdb
 and 
 openvdb-devel 

Where also not installed by the dependancy script.

rabol:
Workaround: 
```
sed -i 's/mesa-libOSMesa-devel/mesa-compat-libOSMesa-devel/g' linux.d/fedora
```

PR #8732, opened 2025-11-08
update the linux.d/fedora file for Fedora 43

nicvagn:
The mesa-libOSMesa-devel is no longer avalible. The mesa-compat-libOSMesa-devel provides the needed files.

I also found I needed:     openvdb-devel

addresses: [https://github.com/bambulab/BambuStudio/issues/8710]


nicvagn:
CMake Error at CMakeLists.txt:590 (message):
  OpenVDB could not be found with the bundled find module.  You can try to
  specify the find module location of your OpenVDB installation with the
  OPENVDB_FIND_MODULE_PATH cache variable.

still complains

PR #8916, opened 2025-11-25
Prevent older cloud presets from overwriting newer local files

L0rdS474n:
- Guard preset downloads by combining local metadata timestamps with file modification times, rejecting cloud profiles that are not fresher.
- Improve logging to show the effective local freshness value used during sync decisions, aiding debugging of timestamp-related issues.

### DIFF
--- a/resources/web/guide/3/index.html
+++ b/resources/web/guide/3/index.html
@@ -1,41 +1,3 @@
-<<<<<<< HEAD
-<!doctype html>
-<html>
-<head>
-<meta charset="utf-8">
-<meta http-equiv="Cache-Control" content="max-age=7200" />
-<title>引导_P1</title>
-<link rel="stylesheet" type="text/css" href="../css/common.css" />	
-<link rel="stylesheet" type="text/css" href="3.css" />
-<link rel="stylesheet" type="text/css" href="../css/dark.css" />
-<script type="text/javascript" src="../js/jquery-2.1.1.min.js"></script>
-<script type="text/javascript" src="../js/json2.js"></script>	
-<script type="text/javascript" src="../../data/text.js"></script>	
-<script type="text/javascript" src="../js/globalapi.js"></script>	
-<script type="text/javascript" src="../js/common.js"></script>
-<script type="text/javascript" src="./3.js"></script>	
-</head>
-<body onLoad="OnInit()">
-<div id="Title">
-	<div class="trans" tid="t6">We kindly request your help to improve everyone's printing.<br/>Come and Join our Customer Experience Improvement Program</div>
-</div>
-<div id="Content">
-	<div id="PolicyTxt" class="TextS2 TextArea1 ZScrol">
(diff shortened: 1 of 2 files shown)


XinZhangBambu:
Thanks for your help. Is there any problem here that requires the time zone to be modified?

image

L0rdS474n:
I think so. No matter what I do, the old profiles reappear after a few hours if I have cloud sync enabled.
Regardless of when I create or update them, my profiles are treated as outdated until we pass the date that the server has stored. When you provide a synchronization service, it is crucial to use UTC for timestamps, because you cannot reliably know the user’s time zone before the files are synchronized. I, for example, travel with my laptop and cross multiple time zones – what happens if some of my profiles are synchronized during such a trip?

I have not yet had time to fully verify that this is exactly what happens, but using a fixed time zone in any synchronization logic feels fundamentally wrong.

XinZhangBambu:
Understood. We will consider the correctness and feasibility of this change and need to conduct actual tests. This may take some time.

https://github.com/bambulab/BambuStudio/pull/8916#issuecomment-3615383769

L0rdS474n:
> Understood. We will consider the correctness and feasibility of this change and need to conduct actual tests. This may take some time.
> 
> [#8916 (comment)](https://github.com/bambulab/BambuStudio/pull/8916#issuecomment-3615383769)

Thanks,

This might also refere to when you have to many profiles according to the server. If I read the Wiki I should be able to have 200 filament profiles and 100 process settings. I'm well under that limit but I get a warning that to many profiles are cached in cloud and I have to have cloud syncronization off for now.

tonghao-bbl:
Hi @L0rdS474n does this issue still persist?

L0rdS474n:
Yes it does, 100%
Tried on Linux and Windows.

tonghao-bbl:
> Yes it does, 100% Tried on Linux and Windows.

Why this PR is closed, can you make it again

L0rdS474n:
I’ve given up on trying to help Bambu Lab. Being constantly attacked by the “Bambu Army” on the forum simply isn’t worth my time or effort.

I’m sorry to say it, but at this point, Bambu Lab as a company simply doesn’t deserve the knowledge and expertise I was willing to contribute.

tonghao-bbl:
Sad to hear this, I rarely browse the forum. I thought we could just do some coding and make tools for everyone.
Anyway, wish you all the best!

L0rdS474n:
That is what I thought as well. Unfortunately, after dealing with so much hostility—including death threats directed at my family—it inevitably makes you question whether this is a company and community you still want to support.

I have never had any problem helping or contributing. However, Bambu Lab has treated me very poorly, declared an agreement we had invalid, and now claims that my warranty no longer applies, despite the fact that I have written confirmation from them stating otherwise. Under those circumstances, my motivation to spend my time improving their software has simply disappeared.

I have even switched to another printer brand because of this. I will continue using the two Bambu Lab printers I already own until they eventually stop working, but after that, no more Bambu Lab products will be coming into my home.

Fortunately, the other company genuinely appreciates the time and effort I put into improving their software, even though the entire platform is closed-source.

Thank you for your kind words, and I wish you all the best as well!

tonghao-bbl:
> including death threats directed at my family

That sounds scary, can you provide more info about this, it is definitely not what Bambu Lab wanted, Maybe there's been some misunderstanding here.

PR #9005, opened 2025-12-03
Add 'Direct Drive High Flow' to common extruder variants

michaelr0:
Adds `Direct Drive High Flow` to `fdm_filament_common.json` in `resources/profiles/BBL/filament` so that Filament profiles created via the Custom Filaments GUI, will be able to toggle between Standard and High Flow.

image

### DIFF
--- a/resources/profiles/BBL/filament/fdm_filament_common.json
+++ b/resources/profiles/BBL/filament/fdm_filament_common.json
@@ -169,7 +169,8 @@
         "0"
     ],
     "filament_extruder_variant": [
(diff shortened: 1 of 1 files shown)


lanewei120:
this file is shared by single extruder and multi-extruders
can not be modified like this

michaelr0:
> this file is shared by single extruder and multi-extruders
> 
> can not be modified like this

This works on an A1 and a P1S, does this create a problem for multi extruder machines? This change is the same changes as found in the Bambu Filament presets which enable the high speed option.

PR #9013, opened 2025-12-04
Precise outer wall is applying to all wall sequences

michaelr0:
Precise wall was not checking for the wall sequence to be Inner/Outer, as a result having Precise wall enabled on Outer/Inner or Inner/Outer/Inner would incorrectly try to compensate the outer wall width, checking I noticed the code comment which suggested that the logic originally came from Orca Slicer.

However it looks like it wasn't completely implemented here in Bambu Studio, I found the following commits on OrcaSlicer and have approximated the changes needed to check that the wall sequence is set correctly.

- https://github.com/OrcaSlicer/OrcaSlicer/commit/d2eb007d5711bc1ffa061952faa0fde2f672dc59
- https://github.com/OrcaSlicer/OrcaSlicer/commit/2f2018f9ee93b4c79fca42bfd51c04f35ab28ed7

### DIFF
--- a/bbl/i18n/en/BambuStudio_en.po
+++ b/bbl/i18n/en/BambuStudio_en.po
@@ -11220,7 +11220,7 @@ msgid "Precise wall"
 msgstr ""
 
 msgid "Improve shell precision by adjusting outer wall spacing. This also improves layer consistency."
-msgstr ""
+msgstr "Improve shell precision by adjusting outer wall spacing. This also improves layer consistency. NOTE: This option will be ignored unless the order of walls is set to inner/outer"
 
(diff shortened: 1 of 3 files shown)


zhimin-zeng-bambulab:
@michaelr0  Thanks a lot for the patch and for taking the time to work on this! we will consider it!

lanewei120:
Hello @michaelr0 

thanks for this patch
could you combine these commits into one?
image

michaelr0:
> Hello @michaelr0
> 
> thanks for this patch could you combine these commits into one? image

Done

zhimin-zeng-bambulab (approved):

lanewei120:
thanks @michaelr0

PR #9024, opened 2025-12-05
Get "other layers" bed temperature from filaments used on the first layer

### LINKED ISSUE #6790: Bed get too hot with PLA and PETG as support
### Bambu Studio Version

2.0.3.54

Bambu Lab Official website

### OS version

Windows 11

_No response_

### Printer

H2D

### How to reproduce

settings:
Textured bed
Pla on AMS
Petg on external spool

reproduce:
Set material to PLA
Set support interface to PETG

### Actual results

When printing initial layer the bed temp is set to 55 degrees as expected, but it raises at the second layer and up to 70 degrees,  the default for PETG. 70 is too high for PLA. Filament change to PETG can be way up the model and PETG does not touch the bed. 

(Tried it also the other way around, using PETG with PLA as support. The bed temp for initial and second layer and up are both 70 degrees as expected.) 

### Expected results

The bed temperature stays at 55 degrees, or the standard bed temp for PLA.

[bedtoohot1.gcode.3mf.zip](https://github.com/user-attachments/files/20006802/bedtoohot1.gcode.3mf.zip)

- [ ] Log file
- [x] Project file

twhitt14:
Fixes #6790 

After several multimaterial prints, I realized that the bed temperature was going far higher than I expected after the first layer. The materials on the first layer did not need such a high temperature.
This happens because a higher bed temperature material was used later - the bed temp rose to the hottest "other layers" temperature even though that material was not touching the bed. The goal here is to make sure that the bed temperature never goes above the highest "other layers" temperature of ONLY the materials that are on the first layer.

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -4031,7 +4031,9 @@ GCode::LayerResult GCode::process_layer(
         // BBS
         int bed_temp = 0;
         if (m_config.bed_temperature_formula == BedTempFormula::btfHighestTemp)
-            bed_temp = get_highest_bed_temperature(false,print);
+            // Keep using the hottest requirement from the first layer so later filaments that never touch the bed
+            // don't raise the bed temperature unnecessarily.
+            bed_temp = get_highest_bed_temperature(true, print);
         else
             bed_temp = get_bed_temperature(first_extruder_id, false, m_config.curr_bed_type);
         gcode += m_writer.set_bed_temperature(bed_temp);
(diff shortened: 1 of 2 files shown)


raulp:
i've learned this the hard way :) were i forgot to set PETG first layer and other layers to 55 and my prints were bad. This is a great addition. Would it be possible to maybe add a warning when previewing?

twhitt14:
@lanewei120 Could you read over this and see what you think?

PR #9025, opened 2025-12-05
Add ability to disable Power Loss Recovery on BBL machines

michaelr0:
# Description

This PR adds the ability to disable [Power Loss Recovery](https://wiki.bambulab.com/en/knowledge-sharing/power-loss-recovery) on Bambu Lab machines.

There is chatter online that the feature causes excessive wear on Micro SD storage, while the feature sounds great, I personally would rather disable it and reduce any additional wear on the storage.

I have added a tick box/toggle to the machine section, will require develop mode to be enabled and is an opt in to disable, so it shouldn't impact any existing logic unless specifically toggled to disable the feature.

I haven't yet been able to get the toolchain setup and compiling locally, but given the minor nature of the changes, I'll be more than happy to test it via the compiled version from GitHub action (once finished) and update the PR with screenshots.

### DIFF
--- a/bbl/i18n/BambuStudio.pot
+++ b/bbl/i18n/BambuStudio.pot
@@ -11095,6 +11095,12 @@ msgstr ""
 msgid "Enable this to enable the camera on printer to check the quality of first layer"
 msgstr ""
 
+msgid "Disable Power Loss Recovery"
+msgstr ""
+
+msgid "Enable this to disable power loss recovery commands in generated G-code"
+msgstr ""
+
 msgid "Enable clumping detection"
 msgstr ""
 
--- a/bbl/i18n/en/BambuStudio_en.po
+++ b/bbl/i18n/en/BambuStudio_en.po
@@ -11249,6 +11249,12 @@ msgstr "Scan first layer"
 msgid "Enable this to enable the camera on printer to check the quality of first layer"
 msgstr "Enable this to allow the camera on the printer to check the quality of the first layer."
 
+msgid "Disable Power Loss Recovery"
+msgstr "Disable Power Loss Recovery"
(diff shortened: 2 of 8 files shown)


michaelr0:
Also opened this as a PR to https://github.com/OrcaSlicer/OrcaSlicer/pull/11582

JohnSmithCYM:
Thansks for your pr and this option in printer preset is not suitable for some cases, especially for those who want to use Power Loss Recovery but he downloaded 3mf from MW or other place. I have submitted a requirement request to fullfill the similar function on printer.

michaelr0:
> Thansks for your pr and this option in printer preset is not suitable for some cases, especially for those who want to use Power Loss Recovery but he downloaded 3mf from MW or other place. I have submitted a requirement request to fullfill the similar function on printer.

This doesn't disable Power Loss Recovery, unless the user is using a custom printer profile with the feature toggled to disable the power loss recovery.

mpaperno:
I think the point is that this setting could end up in a published print profile and then inadvertently get set for someone else using that print (w/out changing printers, presumably)?

Though if that's the case, seems there are other settings which may get carried over that could be even more concerning.  Like physical machine limits.

michaelr0:
> I think the point is that this setting could end up in a published print profile and then inadvertently get set for someone else using that print (w/out changing printers, presumably)?
> 
> 
> 
> Though if that's the case, seems there are other settings which may get carried over that could be even more concerning.  Like physical machine limits.

I think that is covered by a prompt already, if the user ignores that or has it disabled, that would be them accepting the risk? But yeah that's a valid point, much more damage could be done with the other settings, z height limits etc

Issue #9028, opened 2025-12-05
Open Folder from Timelapse doesn't do anything


fcrozat:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab Official website

### OS version

openSUSE Tumbleweed

### Printer

Bambu Lab P2S

### Printer Firmware Version

01.01.00.00

### How to reproduce

Go to Device
Select printer
Select Storage
Select a timelapse
Press Download
Press "open Folder"

### Actual results

Nothing happens

### Expected results

File browser (Nautilus) should open

### Project file & Debug log uploads

No need

### Checklist of files to include

- [ ] Log file
- [ ] Project file

fcrozat:
Issue already reported at https://github.com/flathub/com.bambulab.BambuStudio/issues/80 and on Orca: SoftFever/OrcaSlicer#6449

I'm backporting the Orca fix which works fine

PR #9029, opened 2025-12-05
Fix "Open Containing Folder" on Linux (#9028, SoftFever/OrcaSlicer#6449)

fcrozat:
Backport OrcaSlicer fix for issue #9028

### DIFF
--- a/src/slic3r/GUI/GUI.cpp
+++ b/src/slic3r/GUI/GUI.cpp
@@ -566,7 +566,7 @@ void desktop_open_datadir_folder()
 #endif
 }
 
-void desktop_open_any_folder( const std::string path )
+void desktop_open_any_folder( const std::string& path )
 {
     // Execute command to open a file explorer, platform dependent.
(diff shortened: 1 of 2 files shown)


hadess:
LGTM

hadess:
I just wanted to mention that for both Bambu Studio and Orca Slicer in a Flatpak container, using dbus to call out the file manager would probably be better:
https://github.com/flathub/com.bambulab.BambuStudio/issues/80#issuecomment-3558262635

This code happens to work because the app has full access to the Downloads folder, but it wouldn't work if that wasn't the case, and it doesn't highlight the file in the file manager (like on Windows and macOS) or raise the newly opened window correctly.

MackBambu (approved):

MackBambu:
Thank you, I think this can be merged.
Of course, it would be even better if the Flathub issue could be resolved at the same time.

fcrozat:
The fix is already in flathub, the issue is fixed there (the patch will be dropped with next release of Bambu Studio).

lanewei120:
@fcrozat 
thanks~

PR #9083, opened 2025-12-09
Allow zero key to select filament 10

jomixlaf:
Hi,

This PR allows the use of the digit 0 on the keyboard to map filament 10 to the object choosen. Nothing big but I use 10 color every day! 

Thanks


Haidiye00 (commented):
ok

lanewei120:
thanks for doing this

PR #9084, opened 2025-12-10
Enable Reload from disk in multi-selection menu

jomixlaf:
As mentioned in the title, we can now reload from disk with multi-selection. 

Tested on Mac only (I don't have Windows) with multiple steps and STL. Works like a charm, extremely useful!

SCR-20251209-qsn-2


Haidiye00 (commented):
ok

lanewei120:
thanks

PR #9085, opened 2025-12-10
Add expand/collapse all for plate context menu

jomixlaf:
Another useful feature. Right-click the plate name or the bed plate to expand or collapse items in the object list. 

SCR-20251209-rmo-2

Thanks


lanewei120:
thanks very much
@jomixlaf

Issue #8238, opened 2025-09-20
Adopt "Consistent Surface" and "Short Specific Travel Acceleration" from PrusaSlicer


lorenzofanchi:
PrusaSlicer recently introduced two features that aim to significantly improve the quality of printed models: Consistent Surface and Short Specific Travel Acceleration. The former reduces/solves one of the main issues with inconsistent gloss and matte surfaces on the same model.

The latter should also drastically improve outer surface quality, as it reduces ringing on outer walls.

Top - OrcaSlicer/Bambu Studio default profiles.
Bottom - the new cooling logic and the short specific travel acceleration in PrusaSlicer 2.9.3.

<img width="922" height="779" alt="Image" src="https://github.com/user-attachments/assets/07a147a0-0501-4178-bee0-936e3e54c613" />

Read more about both [here](https://blog.prusa3d.com/new-in-prusaslicer-consistent-surface-finish-and-nerfing-vfas_120400/)

raulp:
This would be such a good feature! @JohnSmithCYM any idea if this is even on the roadmap or considered?

lorenzofanchi:
I was a bit disappointed with VFA's after receiving my H2D, so I hope Bambu can optimize Studio to minimize them 🙏🏼

raulp:
With the help of AI, i was able to build a version of it on my computer. But i need to set my printer in LAN mode to test. Will try to do this over the weekend, might submit the PR afterwards. The preview in the slicer looked amazing. But gotta print it ;)

lorenzofanchi:
Very curious! Would be awesome to see a side by side comparison of the same printed model; one with the current version of Studio and one with your changes!

raulp:
Do you have a model in mind? Share a 3mf and i'll do it. As long as it's not a cube with 100% infill and full plate HAHAHA

lorenzofanchi:
VFA's are usually prominent after high-speed, sharp corners, so I think a 8 x 4 x 1 cm rectangle (like the one in the picture above) should work well. I attached a 3MF that has that primitive and uses the same 0.16 mm Balanced Quality as referred to in the photo above.

[Cube.3mf.zip](https://github.com/user-attachments/files/23787337/Cube.3mf.zip)

raulp:
@lorenzofanchi thanks! Let me check, i might have a problem with LAN mode https://github.com/bambulab/BambuStudio/issues/8887. Let me sort that one first then i can try. Appreciate you sending the model.

[...]

PR #9141, opened 2025-12-17
Port Prusa features

raulp:
As discussed in #8238, ported a few prusa features:
- Consistent Surface
- Short Specific Travel Acceleration

Enabled only if developer mode is enabled.

The way the Consistent Surface works is pretty great, because the outer walls are not slowed down, yet all the other parts are. This is different from "don't slow down outer walls" feature.

Tested for the few weeks on my own and looks great. Would love to, if approved, to figure out a decent support article about all these features that Bambu has in regard to speed, smoothing, cooling.

### DIFF
--- a/resources/profiles/BBL/process/fdm_process_dual_common.json
+++ b/resources/profiles/BBL/process/fdm_process_dual_common.json
@@ -211,6 +211,12 @@
         "10000",
         "10000"
     ],
+    "travel_short_distance_acceleration": [
+        "250",
+        "250",
+        "250",
+        "250"
+    ],
     "top_surface_acceleration": [
         "2000",
         "2000",


zhimin-zeng-bambulab:
@raulp 
Thanks a lot for the patch and for taking the time to work on this!
We think this is a good direction, but there are still a few issues that need to be addressed before we can merge it.
Could you please take another look at the following points and update the patch accordingly?
1. Default values ​​for parameters are not handled in this way; they are still configured via a JSON file.
image

This is the link to the JSON file ： https://github.com/bambulab/BambuStudio/tree/master/resources/profiles/BBL/process

2.  Parameter "short_travel_acceleration " is a nozzle-related parameter. It needs to be added to print_options_with_variant.
image

PS：If you have any questions about the JSON file configuration, you can list the short_travel_acceleration parameter values ​​for all your printer models in a table, and we can help you configure them in the JSON file later.

raulp:
@zhimin-zeng-bambulab thank you for your feedback. Still new to the codebase, so i really appreciate your help pointing me in the right direction.

I made the changes you requested:
- moved cooling_slowdown_logic and cooling_perimeter_transition_distance to fdm_filament_common.json
- moved travel_short_distance_acceleration to fdm_process_common.json and adjusted PrintConfig.cpp

I only added them to the BL profiles, let me know if you think they should be added to all others. Shouldn't be too hard.

Thank you

zhimin-zeng-bambulab:
@raulp 
You also need to configure the default parameters for the multi-extruder printer nozzles in this file(https://github.com/bambulab/BambuStudio/blob/master/resources/profiles/BBL/process/fdm_process_dual_common.json).
 You can refer to `travel_acceleration` and configure all four values ​​of `travel_short_distance_acceleration` to 250.

Tanks you!

raulp:
@zhimin-zeng-bambulab thanks, not sure how i missed one :) Lesson learned. Pushed changes, thank you for the guidance.

lanewei120:
thanks, @raulp

Issue #7250, opened 2025-06-12
Mouse input lost


nijaba:
### Bambu Studio Version

2.1.0.59

### Where is the application from?

Bambu Lab Official website

### OS version

MacOS 15.5

### Additional system information

Projects for the H2D often loose mouse input after a few switch between prepare and preview.  The filament saving mode "dialog" seems to be interfering.  Only solution is to force quit Studio.

### Printer

H2D

### How to reproduce

- Create a project for the H2D
- Switch between prepare and preview adjusting some parameters
-  At some point the filament saving mode will appear and will block any mouse input into studio

### Actual results

Studio can't be used anymore, need to be force quit

### Expected results

Continuing on working

### Project file & Debug log uploads

Can't share that on a public server

### Checklist of files to include

- [ ] Log file
- [ ] Project file

Haidiye00:
Hello, can you download QuickRecorder software and record a GIF of the operation video and upload it? Thank you                         [lihaoyun6/QuickRecorder: A lightweight screen recorder based on ScreenCapture Kit for macOS / 基于 ScreenCapture Kit 的轻量化多功能 macOS 录屏工具](https://github.com/lihaoyun6/QuickRecorder/releases/tag/1.6.5)

nijaba:
https://github.com/user-attachments/assets/ad0dc73f-0c1b-448b-a615-1c664a10b768

Haidiye00:
Hello, thank you for recording. In the video, did you open two Bambustudio files using Bambustudio?

nijaba:
Yes

hotellonely:
Hi it's still here on newest version of BB Studio on macOS. Quite annoying... @Haidiye00  Have you managed to replicate the issue? On the 2.2 Beta 2 it's soo easy to trigger this bug making the software almost unusable on macOS with H2D.

<img width="464" height="390" alt="Image" src="https://github.com/user-attachments/assets/d3bee89e-30c5-4cb9-a96b-35ecfc933509" />

It's mostly because of this menu pop up. Could you please remove it and change it to some other form of UI controller?

[...]

Issue #8940, opened 2025-11-26
Slice Plate tooltip gets stuck & blocks UI


lorenzofanchi:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab github releases

### OS version

macOS 26.1

### Printer

Bambu Lab H2D

### Printer Firmware Version

NA

### How to reproduce

1. It seems to have to do with the timing of pressing Slice Place either just before or just after the tooltip appears.
2. The UI gets stuck and I can't click anywhere anymore. I can press Escape to close the "Warning" modal, but never get past.
3. Only option is to quit and reopen Bambu Studio.

### Actual results

Seems the tooltip steals focus, even though it's not visible anymore.

https://github.com/user-attachments/assets/8a0f0c46-022a-4c66-8a91-28f954fe3934

### Expected results

The UI should not get blocked.

### Project file & Debug log uploads

NA

### Checklist of files to include

- [ ] Log file
- [ ] Project file

qusheng-bambu:
Thank you very much for your feedback. I have reproduced your issue locally and have synchronized it with the relevant developers. I've noticed that this phenomenon might be more likely to occur when the software is maximized. You can temporarily use the software in the default window.

lorenzofanchi:
Hi! I'm running macOS and I'm already using the main window in the default mode. Unfortunately, it happens all the time with the default macOS window size. Hope you can get it fixed soon 🙏🏼

PR #9188, opened 2025-12-24
Fixed issue on macOS where app drops all mouse events and becomes non responsive (#7250)

eldade:
When slicing for H2D/H2C, the filament-saver popup (FilamentGroupPopup) causes a mouse capture issue that occasionally renders the app non-responsive. The issue is with the "Slice Plate" SideButton object.

Because on macOS it doesn't always receive clean pairs of mouseDown and mouseReleased, it can get into a state where the mouse capture is stuck on that button, which means the button now eats up all mouse events for the app.

The fix (which should be VERY safe for all platforms) is to ensure that whenever we receive a mouseReleased, we always check if the button currently has a mouse capture, and then release it if it does (without checking the pressedDown state). I considered limiting the change to just macOS, but I really don't think it is necessary.

### DIFF
--- a/src/slic3r/GUI/Widgets/SideButton.cpp
+++ b/src/slic3r/GUI/Widgets/SideButton.cpp
@@ -329,9 +329,15 @@ void SideButton::mouseDown(wxMouseEvent& event)
 void SideButton::mouseReleased(wxMouseEvent& event)
 {
     event.Skip();
+
+    if (HasCapture())
+    {
+        // Release mouse capture regardless of pressedDown state, to avoid cases where capture may be stuck permanently
+        ReleaseMouse();
+    }
+
     if (pressedDown) {
         pressedDown = false;
-        ReleaseMouse();
         if (wxRect({0, 0}, GetSize()).Contains(event.GetPosition()))
             sendButtonEvent();
     }


raulp:
hope this makes it to the next release soon

eldade:
Thanks! Who are the reviewers on this repo? Are they basically Bambu employees? Do they tend to accept external contributions?

v0idp:
@eldade 
I was looking into the issue #8940 and I think this is very related to the issue of your fix. I thought the root cause was due to the race condition by the FilamentGroupPopup timer mechanism which dismisses the window after 300ms.

- the tryPopup method sets the m_active = true and calls PopupWindow::Popup()
- the tryClose() method starts a 300ms timer to dismiss the popup
- if the slicing button is pressed while the popup is opened and before the 300ms timer triggered for dismissal, the window blocks UI interactions from what I gathered (but not confirmed).

So looking deeper into it the popup modal window captures the mouse and the timer might get reset while the popup is in that inconsistent state which means it never properly dismisses which leaves the modal state in an neverending blocking state.

If I am wrong please correct me. it is the first time looking at the code for a couple of minutes. Just wanted to throw in my thoughts tho.

EDIT: my approach would have been to add a check into FilamentGroupPopup::tryPopup by adding this check before the first m_active check:

`if (m_active && m_timer->IsRunning()) {
    ResetTimer();
    return;
}`

this makes sure that the timer gets reset if the popup is somehow opened again and cancel the rest of the function.

The next part would be to make sure the popup is properly closed when the slice button is clicked in the MainFrame.cpp in the lambda function by adding this check at the very beginning (not tested and not sure if correct):

`if (m_filament_group_popup->IsActive()) {
    m_filament_group_popup->TryClose();
}`

Looking at your solution I do like it and I hope it fixes the issues users have in general with MacOS UI being blocked.

eldade:
@v0idp I'll admit I don't fully understand that timer mechanism, but from my debug session my impression was that the core issue is a wxWidgets/macOS compatibility issue. The Popup class seems to rely on these mouse enter/leave notifications which just don't seem to work that well on macOS (at least not on macOS 26). They just don't fire correctly, so any mechanism that's relying on them is going to fail, one way or another. 

I suspect they work way better on Windows (generally speaking wxWidgets seems to be getting far more love on Windows than on macOS). 

The bug I fixed is really related to that issue -- in a well-designed UI framework the system should enforce clean pairs of mouse down and mouse up, right? Think of the edge cases of like they click the button (mouse capture is grabbed), then while it's pressed they hit cmd+tab and end up in another up. At that simple level, things don't work well with Bambu Studio, from what I can tell. Perhaps they're using an ancient version of wxWidgets? Perhaps wxWidgets needs to be updated for macOS 26? 

Sorry, this is a non-answer... I guess I'm saying my solution seems to address a real problem and I'm not sure if what you're describing is an actual problem at that level or just another manifestation of poor wxWidgets/OS integration? 🤷‍♂️

eldade:
By the way, BS is currently on wxWidgets 3.1.5, a version from April 2021... The latest release, 3.3.1 was released earlier this year and there have been MANY fixes related to macOS mouse handling and many other topics.

I suspect upgrading would take a massive effort, but that is probably the path to truly improving this thing... 

In lieu of that, the next best thing is to hack our way around these Mac mouse handling issues somehow, just please keep in mind that any time you're relying on mouse enter/leave pairs, they won't be reliable (and probably many other OS/platform issues like that).

eldade:
@bambu123 and @wtji-bbl It's now been three weeks since I put up this PR -- this bug is extremely disruptive for macOS customers. Any chance you could make this a priority? Thank you!

jomixlaf:
@eldade i'm with you on this one, being a H2D user I'be been experienced this for quite a while now! I've been very lucky all my 4 PR were push to main the same week they so I'm hoping yours is also part of it. It's been more than 2 months since the last BS update.

walterwongbbl:
@eldade Thank you very much for your contribution.

lanewei120:
thanks

eldade:
Awesome, thanks for the approval. I will try to continue making contributions to this repo as time allows. One question I had for the Bambu maintainers: If I found the time to try and upgrade to the latest version of wxWidgets, would you accept such a large PR? I think it would significantly upgrade the user experience for Mac customers and possibly for others platforms as well. 

cc: @lanewei120 @walterwongbbl @bambu123

lanewei120:
> Awesome, thanks for the approval. I will try to continue making contributions to this repo as time allows. One question I had for the Bambu maintainers: If I found the time to try and upgrade to the latest version of wxWidgets, would you accept such a large PR? I think it would significantly upgrade the user experience for Mac customers and possibly for others platforms as well.
> 
> cc: @lanewei120 @walterwongbbl @bambu123

Sure, it will be a great help to us

Currently we use wxwidgets from https://github.com/bambulab/wxWidgets

Issue #9246, opened 2025-12-31
Filter out tiny gaps doesn't work on top surface


lin-xianming:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab github releases

### OS version

Windows 10

### Printer

X1C

### Printer Firmware Version

01.10

### How to reproduce

Set filter out tiny gaps to a high value but gap fills still exist. Tried various values like 10, 100, 100 but nothing worked.

### Actual results

<img width="2318" height="1156" alt="Image" src="https://github.com/user-attachments/assets/c24e6b19-4353-4450-9e32-7adf1014027a" />

### Expected results

No gap fills

### Project file & Debug log uploads

[filter out gap fill does not work.3mf.zip](https://github.com/user-attachments/files/24397918/filter.out.gap.fill.does.not.work.3mf.zip)

### Checklist of files to include

- [ ] Log file
- [ ] Project file

Scootworks:
I also tried the Arachne wall generator, but it had no effect.

<img width="1186" height="412" alt="Image" src="https://github.com/user-attachments/assets/0be91665-3c76-45e2-930d-cb3e28cb1088" />

hotellonely:
I think you misunderstood the meaning of Filter out tiny gaps. It means to not generate sparse infill in those regions. And supposedly has nothing to do with top surface.

Scootworks:
> I think you misunderstood the meaning of Filter out tiny gaps. It means to not generate sparse infill in those regions. And supposedly has nothing to do with top surface.

maybe I didn't but according to the tooltip, it should filter out gap_infill and gap_infill are the white lines in the model:

<img width="1108" height="531" alt="Image" src="https://github.com/user-attachments/assets/30bb8c3d-c2f8-4444-a470-a63cf1f9f023" />

For example Orca Slicer has this, where there is also no gap_infill is applied as a standard. the model and the slicing profiles are the same:

<img width="1044" height="533" alt="Image" src="https://github.com/user-attachments/assets/95762b1f-1d0b-4529-adae-cefbc4d6d62a" />

xing-mantian:
Thank you for your feedback. We appreciate you taking the time to share your observations.

[...]

PR #9263, opened 2026-01-03
Fix top surface gap fill filter

lin-xianming:
Fix https://github.com/bambulab/BambuStudio/issues/9246

### DIFF
--- a/resources/profiles/BBL/process/fdm_process_dual_common.json
+++ b/resources/profiles/BBL/process/fdm_process_dual_common.json
@@ -211,6 +211,12 @@
         "10000",
         "10000"
     ],
+    "travel_short_distance_acceleration": [
+        "250",
+        "250",
+        "250",
+        "250"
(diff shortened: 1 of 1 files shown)


hotellonely:
This shouldn't be mixed with the settings made for Sparse Infill. It should have its own settings option and its own states variable.

Scootworks:
> Fix #9246

I appreciate your quick fix.

I believe @hotellonely  means that this setting might be confusing because it's listed under the sparse infill section. It would be clearer to place this option in a different section, such as 'Advanced' or a new one.

lin-xianming:
There are enough settings in the wrong category that I didn't even notice. But I'll leave it up to Bambu devs to decide on UI changes. It's not really clear where it should go, because gap fill is used to both ensure the strength of perimeters in classic mode and also the quality of the top surface when using monotonic lines.

PR #9300, opened 2026-01-07
main: Require libx11 1.8

hadess:
Require a version of libX11 new enough that it will initialise threads support itself, and not require X11 specific code in our startup.

See https://gitlab.freedesktop.org/xorg/lib/libx11/-/commit/afcdb6fb0045c6186aa83d9298f327a7ec1b2cb9

### DIFF
--- a/src/BambuStudio.cpp
+++ b/src/BambuStudio.cpp
@@ -85,10 +85,6 @@ using namespace nlohmann;
 #include "slic3r/GUI/GuiColor.hpp"
 #include <GLFW/glfw3.h>
 
-#ifdef __WXGTK__
-#include <X11/Xlib.h>
-#endif
-
 #ifdef SLIC3R_GUI
     #include "slic3r/GUI/GUI_Init.hpp"
 #endif /* SLIC3R_GUI */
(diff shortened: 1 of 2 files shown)


MackBambu:
@hadess Ubuntu 22.04 only provides libX11 1.7.5. Is it possible to detect the libX11 version at runtime and provide backward compatibility?

```
Package 'x11' has version '1.7.5', required version is '>= 1.8'
```

hadess:
No, there's no version API available at runtime to fallback to calling `XInitThreads()` manually.

MackBambu:
> No, there's no version API available at runtime to fallback to calling `XInitThreads()` manually.

@hadess Add conditional compilation checks？
```
pkg_check_modules(LIBX11 REQUIRED x11)
if (LIBX11_VERSION VERSION_LESS "1.8")
    add_compile_definitions(NEED_XINITTHREADS=1)
endif()
```

```
#if NEED_XINITTHREADS
    XInitThreads();
#endif
```

hadess:
> @hadess Add conditional compilation checks？

That won't work, because the API/ABI is the exact same between releases, so something built with a newer libX11 but run against an old libX11 won't run XInitThreads() which will break things, and something built against an older libX11 will always run XInitThreads() in its startup path.

MackBambu:
> That won't work, because the API/ABI is the exact same between releases, so something built with a newer libX11 but run against an old libX11 won't run XInitThreads() which will break things, and something built against an older libX11 will always run XInitThreads() in its startup path.

In practice, we provide two different AppImage builds. The AppImage built on Ubuntu 22 is recommended for use on systems earlier than Ubuntu 24, while the AppImage built on Ubuntu 24 is intended for systems running Ubuntu 24 or newer.

hadess:
> In practice, we provide two different AppImage builds. The AppImage built on Ubuntu 22 is recommended for use on systems earlier than Ubuntu 24, while the AppImage built on Ubuntu 24 is intended for systems running Ubuntu 24 or newer.

OK, that could work then, I'll give it a try.

PR #9328, opened 2026-01-09
Fix path used for filesystem operation incorrectly sanitized

lin-xianming:


wtji-bbl (commented):
+1

lin-xianming:
@lanewei120 Seems like this is ok can you merge it?

lanewei120:
thanks

PR #9367, opened 2026-01-12
Remember print/filament presets per printer when switching

### LINKED ISSUE #9105: settings change when switching printer, ruining print quality
### Bambu Studio Version

2.4.0.70

Bambu Lab Official website

### OS version

Mac OS

_No response_

### Printer

any

### Printer Firmware Version

NA

### How to reproduce

Open the profile, switch from X1 to H2D and the global "speed" settings will change. 

Switch to A1 and the filament "settings overrides" and "flow" will change as well 

see the video showing the problem: https://storage.googleapis.com/riverfamily/videos/2.mov

This was not happening in the past as all the settings where transferring 

This is a serius problem ruining the quality of all my shoes design. Speed, flow, retraction length, Z hop, wipe might not be crucial for some models but they are for mine. For example I need big "normal" Z hop with no wipe to avoid grabbing the soft material while moving.

Please fix it fast as people are already leaving bad reviews as my reputation is being ruined

### Actual results

settings change instead of transfer

### Expected results

all filament and print settings should stay the same when switching printer

[Tora_shoes.3mf.zip](https://github.com/user-attachments/files/24130396/Tora_shoes.3mf.zip)

- [ ] Log file
- [x] Project file

y3dltd:
## Summary

This PR adds per-printer preset memory, similar to how OrcaSlicer handles printer switching. When switching between different printer models (e.g., X1C to P1P to A1), the last used print profile and filament presets are now saved and restored automatically.

**Problem:** Currently, when changing printers, BambuStudio doesn't remember the print profile you were using with each printer. Users have to manually re-select their preferred print profile every time they switch printers.

**Solution:** Store print/filament presets per printer in the app config, and restore them when switching back to that printer.

## Changes

- **AppConfig.hpp/cpp**: Added `m_printer_settings` map and helper functions to store/retrieve printer-specific settings
- **PresetBundle.cpp**: 
  - Modified `export_selections()` to save presets per printer
  - Modified `load_selections()` to restore presets per printer
  - Maintains backwards compatibility with global presets

## Test Plan

- [ ] Switch from Printer A to Printer B, note the print profiles
- [ ] Change the print profile on Printer B
- [ ] Switch back to Printer A - verify the original print profile is restored
- [ ] Switch to Printer B - verify the changed print profile is remembered
- [ ] Test with multiple filament presets (multi-color prints)
- [ ] Verify backwards compatibility (fresh install still works)

Fixes #9105

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -699,6 +699,13 @@ std::string AppConfig::load()
                     }
                     m_printer_cali_infos.emplace_back(cali_info);
                 }
+            } else if (it.key() == "printer_presets") {
+                // Load printer-specific preset settings (remembers presets per printer)
+                for (auto& j_model : it.value()) {
+                    if (j_model.contains("machine")) {
+                        m_printer_settings[j_model["machine"].get<std::string>()] = j_model;
+                    }
+                }
             } else {
                 if (it.value().is_object()) {
                     for (auto iter = it.value().begin(); iter != it.value().end(); iter++) {
@@ -884,6 +891,11 @@ void AppConfig::save()
             j_model["nozzle_diameter"] = escaped;
             j[MODELS_STR].push_back(j_model);
         }
+    }
+
+    // Write printer-specific preset settings (remembers presets per printer)
+    for (const auto& preset : m_printer_settings) {
+        j["printer_presets"].push_back(preset.second);
     }
 
     boost::nowide::ofstream c;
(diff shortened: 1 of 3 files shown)


y3dltd (commented):
LGTM. The per‑printer preset persistence looks correct and keeps backward compatibility via the global presets. I didn’t spot regressions in the diff.

tonghao-bbl:
Thanks for you contributions, but the issue is about settings migrating: should we keep user modified settings to a another printer?
This PR can't fix this issue and we will ship a fix next release.

PR #9387, opened 2026-01-13
BUG: Fix "Cannot register URI scheme ... more than once"

hadess:
Call RegisterHandler only once per run to avoid loads of warnings on startup.


hadess:
This needs some changes so it only runs with the GTK backend.

MackBambu:
@lanewei120 On Linux, the URI scheme only needs to be registered once, so it can be merged.

hadess:
> @lanewei120 On Linux, the URI scheme only needs to be registered once, so it can be merged.

The error message comes from WebKitGTK which is only used on Linux:
https://github.com/WebKit/WebKit/blob/fdb282b50fa810129e6d14c95010a1d30e246fe9/Source/WebKit/UIProcess/API/glib/WebKitWebContext.cpp#L1401

PR #9477, opened 2026-01-24
Fix user settings not saved when Bambu Studio is used for the first time

lin-xianming:
This patch creates the log directory if it does not exist before chdir.

This problem along with a bug where incorrect path was used (fixed in #9328), causes a problem on Windows where if `%APPDATA%\BambuStudio\log` did not already exist before the application is started, `%APPDATA%\BambuStudio\BambuStudio.conf` will be unable to be saved, causing loss of user settings.


mpaperno:
I was just affected by this issue when setting up a new instance of Studio for a dev environment. On 2nd start all my previous settings were gone and the wizard started up again.  Thought it was something I had messed up during setup, actually.  Good fix.  👍🏼

wtji:
Hi, we have already done this fix and will publish it in next release. Thank you for pointing out the problem.

PR #9485, opened 2026-01-25
FIX: Fix unreadable overlay elements in 3D view on XWayland/Intel

hadess:
Backported from:
https://github.com/OrcaSlicer/OrcaSlicer/pull/11557

Tested in:
https://github.com/flathub/com.bambulab.BambuStudio/pull/90


MackBambu (commented):
```#ifdef __linux__```
The potential impact is unclear, so I'd like to limit this change to Linux only to keep the impact as small as possible.

hadess:
> The potential impact is unclear, so I'd like to limit this change to Linux only to keep the impact as small as possible.

Done.

MackBambu (approved):

MackBambu:
@lanewei120 Can be merged

lanewei120:
thanks~

Issue #8659, opened 2025-11-03
Add an Elegoo Centauri Carbon Printer Profile


PrinterMcgee:
**Is your feature request related to a problem? Please describe.**
The Elegoo Centauri Carbon is missing from the available printer profiles

**Describe the solution you'd like**
It's available in Orca Slicer, the profile would just need to be adapted to Bambu Studio

**Describe alternatives you've considered**

**Additional context**

<img width="793" height="636" alt="Image" src="https://github.com/user-attachments/assets/faf96d10-54f9-435d-beab-f24d4b324382" />

yelnesr-2962:
Checking for progress on this request, happy to do beta testing

adjbuzz:
> Checking for progress on this request, happy to do beta testing

Seconded.

ki5axg:
Upvote!

echosintherain:
Would love this f you can

garrett-brennan:
Upvoted!

tranhoanvo:
Up!

gemmapreece85-prog:
Please make this happen

tghastings:
Pull Request opened https://github.com/bambulab/BambuStudio/pull/9489

dbyy:
Would be nice if that would become official soon.

arkanturk77-blip:
am going to buy the Centauri Carbon 2 today, plz keep this slicer works for the printer

PR #9489, opened 2026-01-25
Added Elegoo Centauri Carbon Profiles

tghastings:
**Summary**

  - Adds complete printer profile for the Elegoo Centauri Carbon 3D printer
  - Includes support for 0.2mm, 0.4mm, 0.6mm, and 0.8mm nozzle sizes
  - Adds corresponding print process profiles for each nozzle size

  **Files Added**

  - Machine profiles: Base printer config + 4 nozzle variants
  - Process profiles: 4 layer height presets (0.10mm, 0.20mm, 0.30mm, 0.40mm)
  - Build plate: STL model and texture for bed visualization
  - Printer image: Cover image for printer selection UI

  **Printer Specs**

  - Build volume: 256 x 256 x 256 mm
  - Klipper firmware
  - Hardened steel nozzle
  - Based on OrcaSlicer Elegoo Centauri Carbon profiles

  **Test Plan**

  - Verify printer appears in printer selection dropdown
  - Verify printer cover image displays correctly
  - Verify build plate model and texture render correctly
  - Verify all nozzle sizes are selectable
  - Verify process profiles load for each nozzle size
  - Test slicing a model with the new profile

In support of https://github.com/bambulab/BambuStudio/issues/8659

CC 1
CC 2


murilobast:
Just tested it, it builds, runs, slices, but there's no way to send to the printer other than using a flash drive. To make it fully work, we need to support Elegoo Link.

tghastings:
> Just tested it, it builds, runs, slices, but there's no way to send to the printer other than using a flash drive. To make it fully work, we need to support Elegoo Link.

@murilobast Try again. I added ocotoprint (the same thing the Neptune printer uses).

murilobast:
> > Just tested it, it builds, runs, slices, but there's no way to send to the printer other than using a flash drive. To make it fully work, we need to support Elegoo Link.
> 
> @murilobast Try again. I added Octoprint (the same thing the Neptune printer uses).

Sadly, the Centauri Carbon does not support OctoPrint; the only way to send prints to the printer is through Elegoo Link. Neptune printers are Klipper or Klipper-based with native Octoprint support; the CC firmware is not Klipper (at least not entirely)

murilobast:
I've created a simple OctoPrint bridge you can run on a Raspberry Pi or any device with Python installed, which pretends to be Octoprint and sends the g-code to the printer using Elegoo Link, dunno if this can be integrated here.

Here's the repo: https://github.com/murilobast/centauri-carbon-octoshim

bhoot1234567890:
any update on this?

T3chGuy77:
apparently devs move slower than glaciers. This should have been reviewed and approved by now.

PR #9493, opened 2026-01-27
Fix P2S max z acceleration for high flow nozzle

northbymidwest:
This change corrects the maximum z-acceleration for the high-flow nozzle variant. The value for the standard nozzle is 500, while the high-flow nozzle is set to 200. I cannot see any reason the value should differ between these nozzles since z motion has nothing to do with the nozzle (especially on the P2S where the gantry is fixed in the z-direction...)


lanewei120:
thanks @northbymidwest 
we have already fixed it locally, and will release it soon

PR #9504, opened 2026-01-27
2.5.0 Build Fixes

hadess:


hadess:
I removed the commit for the fix that was handled in https://github.com/bambulab/BambuStudio/pull/9498

PR #9664, opened 2026-02-04
Fix window drag lag on Windows

### LINKED ISSUE #934: Application lags for multiple seconds when window is dragged
**Describe the bug**
The application lags when dragging the window around from the title bar.

The application appears to be doing non-trivial work while it's being dragged. With a high polling rate mouse the application can't keep up and begins to lag behind the mouse cursor. This isn't terribly noticeable at 500 Hz or below, but with a 1000 Hz mouse it's very apparent. When I stop dragging the window it can take multiple seconds for it to catch up. It's more apparent with small windows that fit entirely on-screen. It doesn't matter which tab I'm on (Prepare, Preview, etc). I see the main thread maxing out one CPU core, but I'm not entirely sure what it's doing. The top few stack frames when profiling aren't particularly illuminating.

I'm assuming either the main thread is repainting the window while dragging, the window is processing a single message per frame, or the message loop is doing something heavy for each received message.

image

**3mf File for This Bug**
N/A

**To Reproduce**
1. Open Bambu Studio
2. Click on the title bar
3. Drag the window around
4. Notice the delay

**Expected behavior**
The application should be able to paint fast enough to 

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10 22H2
 - Bambu Studio: 1.4.0.18

4RH1T3CT0R7:
- Suppress expensive Refresh() / Layout() / GL re-render during interactive window drag and resize on Windows
  - Use WM_ENTERSIZEMOVE / WM_EXITSIZEMOVE messages to detect the operation, then do a single refresh when it ends

  Problem

  Dragging the window by the title bar causes multi-second lag and pegs one CPU core at 100%. The custom titlebar handling fires WM_NCCALCSIZE hundreds of times
   per second during drag, each triggering wxEVT_SIZE → Refresh() → full OpenGL render with no throttling.

  Fixes #934

  Changes

  - MainFrame.hpp: Add m_is_in_move_or_resize flag (#ifdef _WIN32)
  - MainFrame.cpp: Handle WM_ENTERSIZEMOVE / WM_EXITSIZEMOVE in MSWWindowProc()
  - MainFrame.cpp: Guard Refresh() / Layout() / wxQueueEvent() in wxEVT_SIZE handler

### DIFF
--- a/src/slic3r/CMakeLists.txt
+++ b/src/slic3r/CMakeLists.txt
@@ -689,7 +689,7 @@ target_link_libraries(libslic3r_gui libslic3r cereal imgui imguizmo minilzo GLEW
 #target_link_libraries(libslic3r_gui libslic3r cereal imgui imguizmo minilzo GLEW::GLEW OpenGL::GL hidapi libcurl OpenSSL::SSL OpenSSL::Crypto ${wxWidgets_LIBRARIES} glfw)
 
 if (MSVC)
-    target_link_libraries(libslic3r_gui Setupapi.lib)
+    target_link_libraries(libslic3r_gui Setupapi.lib dwmapi.lib)
 elseif (CMAKE_SYSTEM_NAME STREQUAL "Linux")
     FIND_LIBRARY(WAYLAND_SERVER_LIBRARIES NAMES wayland-server)
     FIND_LIBRARY(WAYLAND_EGL_LIBRARIES    NAMES wayland-egl)
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -69,6 +69,7 @@
 #include <dbt.h>
 #include <shlobj.h>
 #include <shellapi.h>
+#include <dwmapi.h>
 #endif // _WIN32
 #include <slic3r/GUI/CreatePresetsDialog.hpp>
 
@@ -183,6 +184,14 @@ DPIFrame(NULL, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, BORDERLESS_FRAME_
 {
 #ifdef __WXOSX__
     set_miniaturizable(GetHandle());
+#endif
+
+#ifdef _WIN32
+    // Enable DWM hardware-accelerated compositing for this borderless window.
+    // Without this, dragging the window by the title bar causes lag because
+    // DWM can't efficiently composite a frameless window during drag.
+    MARGINS margins = {0, 0, 0, 1};
+    DwmExtendFrameIntoClientArea((HWND)GetHandle(), &margins);
 #endif
 
     if (!wxGetApp().app_config->has("user_mode")) {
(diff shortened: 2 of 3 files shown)


ianalexis:
I've tested it on Orca but the same issue. Didint see any change.

mpaperno:
Didn't make any apparent difference during dragging here either, I'm afraid.  Window 10, dark mode.  Applied to latest `main` branch head as of this writing.  

It does change the window resize behavior, but the delayed repaint is a bit odd IMHO, and at least on my system it doesn't seem to make it more efficient/quicker (though I could see how that could maybe vary by hardware).

No visible effect on window drag performance, still lags. Doesn't seem like the code in the `wxEVT_SIZE` is being hit at all during window drags, since I can't get it to log anything (except when actually resizing).

mpaperno:
@4RH1T3CT0R7  Any feedback or update on this? It would be nice to get the lag fixed and seems like you've already done some research into the matter.  Thanks!

4RH1T3CT0R7:
@mpaperno Thanks for the detailed feedback and for testing!

You were right - `wxEVT_SIZE` doesn't fire during window drags at all, so guarding it had no effect on drag performance.

I've pushed an updated fix that addresses the issues you raised:

1. **Drag lag fix**: Added `DwmExtendFrameIntoClientArea` with a 1-pixel bottom margin to enable DWM hardware-accelerated compositing for the borderless window. This is the standard approach used by Chrome, Electron, and other frameless window apps to eliminate drag lag on Windows.

2. **Resize visual fix**: Replaced the complete layout suppression (which caused the "odd" frozen content) with throttling to ~30fps using `GetTickCount64()`. The window content now updates smoothly during resize instead of freezing and jumping at the end.

3. **Also fixed**: `WM_EXITSIZEMOVE` now fires `EVT_NOTICE_CHILDE_SIZE_CHANGED` for proper child widget updates, and fixed a switch fall-through in the `WM_GETMINMAXINFO` case.

Would appreciate if you could test again when you get some time

mpaperno:
Thanks Artem!

Didn't know about the pixel trick... neat.

Well, over here the dragging performance is an improvement, if not an all-out fix.  At least it's controllable now, if dragging slow enough, whereas before I wouldn't want to even touch the window except to maximize/restore it.  It still lags and if moving too fast I have to wait for it to "catch up" to current mouse location, but it does so quicker than before, and at least it's usable.  👍🏼 

And the resize behavior is much improved!

BTW my test system is Win 10 with all updates, NVIDIA RTX 2080Ti w/latest studio drivers, 3 monitors. i7-7820X 8 core @3.6Ghz, 64GB RAM.

Hopefully this could get merged in!

Thanks again,
-Max

lanewei120:
thanks~
@4RH1T3CT0R7 @mpaperno

Readun:
This Window lag/delay seems to be still a thing in Version 2.8.2.61
Windows 10 22H2 19045.7663
NVDriver 581.80

PR #9708, opened 2026-02-08
Fix and reject some warning-classes when building

rawler:
Building Bambu Studio generates _thousands_ of warnings. A few of these represents real bugs, I.E. using "null" for a reference is illegal C++, and causes undefined behavior. Others are just noise, and contributing to a bad state of [broken windows](https://blog.codinghorror.com/the-broken-window-theory/).

Trying to improve the situation in this codebase, the first two commits of this PR fixes a few of the "simple & safe" classes of warnings, such as `pessimizing-move`, `misleading-indentation` and `catch-value`, and turns these warnings into errors - to avoid future reappearance.

The second two commits continues with warnings that require _slightly_ more complex analysis, but should still be fairly safe to fix without introducing new problems. Here are a few cases where the warnings represent _actual_ bugs (illegal C++), such as returning a reference to a stack-local null.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,15 @@ if (NOT MSVC AND ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMP
     # On GCC and Clang, no return from a non-void function is a warning only. Here, we make it an error.
     add_compile_options(-Werror=return-type)
 
+    # Prevent common bugs that can be caught at compile time:
+    # - misleading-indentation: code that looks like it's in a block but isn't
+    add_compile_options(-Werror=misleading-indentation)
+
+    # C++-only warnings (these flags are not valid for C):
+    # - pessimizing-move: std::move() preventing copy elision (RVO/NRVO)
+    # - catch-value: catching exceptions by value instead of reference (slicing)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Werror=pessimizing-move> $<$<COMPILE_LANGUAGE:CXX>:-Werror=catch-value>)
+
     add_compile_options(-Wno-unused-function -Wno-unused-variable -Wno-unused-but-set-variable -Wno-unused-label -Wno-unused-local-typedefs)
 
     # removes LOTS of extraneous Eigen warnings (GCC only supports it since 6.1)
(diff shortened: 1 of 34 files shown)


rawler:
Full Disclosure; These commits were created with the help of Claude, but I have gone over them all myself, trying to avoid contributing to low-quality AI slop PR:s.

src/libslic3r/Config.cpp
```diff
@@ -1810,7 +1810,8 @@ const double& DynamicConfig::opt_float(const t_config_option_key &opt_key, unsig
         return opt_floats_nullable->get_at(idx);
     } else {
         assert(false);
-        return 0;
+        static const double s_zero = 0.0;
+        return s_zero;
```
rawler: This is an example of an actual bug being fixed. Returning 0 here, AFAIU means returning a reference to a stack-allocated variable containing 0, which is illegal and can break if caller calls something else before dereferencing this functions return value. (It triggered return-local-addr warning)

src/libslic3r/GCode/GCodeProcessor.cpp
```diff
@@ -958,8 +958,8 @@ void GCodeProcessor::TimeProcessor::post_process(const std::string& filename, st
 
     auto handle_filament_change = [&filament_blocks,&machine_start_gcode_end_line_id,&machine_end_gcode_start_line_id](int filament_id,int line_id){
         // skip the filaments change in machine start/end gcode
-        if (machine_start_gcode_end_line_id == (unsigned int)(-1) && (unsigned int)(line_id)<machine_start_gcode_end_line_id ||
-            machine_end_gcode_start_line_id != (unsigned int)(-1) && (unsigned int)(line_id)>machine_end_gcode_start_line_id)
+        if ((machine_start_gcode_end_line_id != (unsigned int)(-1) && (unsigned int)(line_id) < machine_start_gcode_end_line_id) ||
```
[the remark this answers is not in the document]
XunZhangBambu: The intention is to skip all filament-change G-code included inside the machine’s start and end G-code sections.
The variables start_gcode_end_line and end_gcode_start_line denote the line indices where the start G-code ends and the end G-code begins, respectively, and are initialized to -1.

A line should be skipped if:

- start_gcode_end_line is still -1 (we haven’t passed the start G-code yet), or
- end_gcode_start_line has been set (not -1) and the current line_id is greater than end_gcode_start_line (we are inside the end G-code).

CMakeLists.txt
MackBambu, line 1: @lanewei120 这里的warning作为error可以提高代码质量，但是会导致错误变多，不确定要不要合入

PR #9727, opened 2026-02-10
feat: Z Anti-Aliasing (ZAA) non-planar contouring

mnott:
## Summary

Adds Z Anti-Aliasing (ZAA) — a non-planar contouring post-processing step that smooths staircase artifacts on curved surfaces by micro-adjusting Z heights during extrusion. Based on the [BambuStudio-ZAA](https://github.com/AdoC3/BambuStudio-ZAA) implementation by AdoC3.

- Integrates ZAA as a toggleable per-object setting (disabled by default)
- Adds contouring parameters: max angle, smoothing radius, transition width
- Post-processes G-code to replace discrete layer boundaries with smooth Z transitions
- Verified output: 1,306 unique Z heights vs 85 without ZAA, with only 3 redundant Z moves

## Changes

- **81 files** modified, +2,447/-214 lines
- Core ZAA engine in `src/libslic3r/NonplanarSurface/`
- G-code post-processing in `GCode.cpp` and `GCodeWriter.cpp`
- UI integration: print settings panel, preset handling
- Build system: `build.sh` for macOS, `BUILD.md` walkthrough
- Documentation: `docs/ZAA.md`
- Network plugin compatibility fix for third-party builds

## Test plan

- [x] Verified ZAA output quality with test model (quadlock)
- [x] Compared with/without ZAA: correct Z interpolation, minimal redundant moves
- [x] macOS arm64 build, sign, notarize — full release tested
- [x] Network plugin downloads and loads correctly
- [ ] Windows / Linux build verification (untested)


digitalresistor:
\o/ this would be great to have in a released version of BambuStudio! Thanks for doing the work of getting an MR put together.

samik1994r:
Great work, tried this with cursor 0claude 4.6 spend 50$ on api and it didnt manage, Kudos to the developers !!!!

alidorauxier:
Oh this could be interesting! I've wondered about the possibility of anti-aliasing before just using a series of 'substeps' in the perimeters.
Adob's approach is vastly superior - there's no time penalty and it works with thick layers, all the infill patterns and even ironing.
It's really the best use of non-planar printing - Sort of a hybrid/win-win approach, thank you so much for bringing this to the Bambu community - I can't wait until this is the hands of users.

I actually downloaded the flatpak artifact from the pr but I'm not seeing the effect in the slicer.
The Z contour option is checked but it doesn't seem to produce any non-planar lines in the preview when sliced.

I'm dying to see it in action - good luck with the merge

PR #9747, opened 2026-02-11
CI build: replace macos-13 with macos-15-intel

raulp:
According the the latest build messages, macos-13 x86_64 were deprecated and removed: https://github.com/actions/runner-images/issues/13046
image
image

This PR brings back the x86_64 builds via macos-15-intel images. Built and ran ok with the new image: https://github.com/raulp/BambuStudio/actions/runs/21788141244


MackBambu:
Thank you for your help！

PR #9762, opened 2026-02-14
batch pressure advance calibration

raulp:
This PR adds the ability to test multiple speed/acceleration combinations in a single PA Pattern calibration print, like OrcaSlicer. Original work here: https://github.com/OrcaSlicer/OrcaSlicer/pull/7199/changes#diff-eca08a5705cbb985a1c40eec4d872104395681026980e05ce808dbd5445c4691

Instead of running one calibration at a time, you can enter comma-separated speeds (e.g. 60,200) and accelerations (e.g. 2000,5000) in the calibration dialog. Each combination gets its own labeled pattern, automatically arranged across plates when they don't all fit on one.

Added validation for min and max speed and accelerations.
Right side prints the speed/acceleration as well, to easily distinguish between tests.

Tested on my H2C.

image
image
image

### DIFF
--- a/src/libslic3r/Calib.cpp
+++ b/src/libslic3r/Calib.cpp
@@ -3,6 +3,8 @@
 #include "Model.hpp"
 #include "GCode.hpp"
 #include <cmath>
+#include <sstream>
+#include <iomanip>
 
 
 namespace Slic3r {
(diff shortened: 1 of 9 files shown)


philipp-lampert:
Amazing!

PR #9809, opened 2026-02-21
Implement per-object top,bottom,initial flow ratios

logikal:
Building locally on osx breaks with every release, it feels like, so pushing to GH just to get the action to build for me.


zhimin-zeng-bambulab (approved):

tangketan (changes requested):
1. There is no role called "bottom solid infill"; it's a bit strange to add the bottom_solid_infill_flow_ratio parameter. Please provide a use-case where the new parameter is necessary.
2. For the first layer of an object, the flow rate is multiplied twice (initial_layer_flow_ratio and bottom_solid_infill_flow_ratio), which is confusing for many people. Is it better to only apply one of the parameters?

logikal:
Apologies, this was meant to still be a draft, and I'm still experimenting.

PR #9858, opened 2026-02-27
[Windows] Attach GUI to current console when started from a command line

mpaperno:
Currently there is no output to the console when running Studio from a command line on Windows. For example `--help` output or CLI parameter parsing errors.  The program just appears to not do anything at all in these cases.  This is typical for Windows GUI programs since they don't usually need console interaction.

This fix will attach to the current console, _if there is one_, meaning only if it was started from a console in the first place.  It will **not** open a new console window or start a new process.  
[I use this technique on "all" my Win GUI programs for a couple decades now.]

Of course `--help` is the only way to get a real listing of currently supported switches, since the Wiki article is rather dated/missing a lot of them.

Thanks,
-Max

### DIFF
--- a/src/BambuStudio_app_msvc.cpp
+++ b/src/BambuStudio_app_msvc.cpp
@@ -190,7 +190,15 @@ extern "C" {
 }
 extern "C" {
 #ifdef SLIC3R_WRAPPER_NOCONSOLE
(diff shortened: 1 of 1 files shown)


tonghao-bbl:
LGTM @lanewei120

lanewei120:
thanks for doing this

mpaperno:
Thanks for the review & merge!

PR #9919, opened 2026-03-08
feat(presets): reload user presets from disk without restarting

adele-with-a-b:
## Reload Presets (Cmd+R)

Reload user presets from disk without restarting the app.

- **Menu:** File → Reload Presets (Cmd+R)
- **Button:** Refresh icon next to the preset dropdown on all tabs (process, filament, printer)
- **Auto-select:** When new presets are detected after reload, the most recent one is automatically selected

This enables external tools (scripts, AI assistants, CLI workflows) to generate Bambu Studio preset JSON files that appear in the slicer immediately — no restart required.

### Preset Directory

User presets are stored at:

| OS | Path |
|---|---|
| macOS | `~/Library/Application Support/BambuStudio/user/<user_id>/process/` |
| Windows | `%APPDATA%\BambuStudio\user\<user_id>\process\` |
| Linux | `~/.config/BambuStudio/user/<user_id>/process/` |

Replace `<user_id>` with your Bambu account ID (visible in the directory), or `default` if not logged in. Filament presets go in `filament/` instead of `process/`.

Each preset needs two files: a `.json` and a `.info`.

### Process Preset Example

**Structural PETG-CF 0.6mm @BBL H2C.json** — a structural profile with 4 walls and 40% infill:
```json
{
    "from": "User",
    "inherits": "0.24mm Balanced Strength @BBL H2C 0.6 nozzle",
    "name": "Structural PETG-CF 0.6mm @BBL H2C",
    "print_extruder_id": ["1", "1", "2", "2"],
    "print_extruder_variant": ["Direct Drive Standard", "Direct Drive High Flow", "Direct Drive Standard", "Direct Drive High Flow"],
    "print_settings_id": "Structural PETG-CF 0.6mm @BBL H2C",
    "version": "2.5.0.7",
    "wall_loops": "4",
    "sparse_infill_density": "40%"
}
```

**Structural PETG-CF 0.6mm @BBL H2C.info**:
```
sync_info = create
user_id = <your_user_id>
setting_id = 
base_id = <parent_profile_setting_id>
updated_time = 0
```

The `base_id` is the `setting_id` from the parent system profile JSON (found in the `system/BBL/process/` directory). Only settings that differ from the inherited base need to be included.

Drop both files into the preset directory and hit Cmd+R.

## Implementation

- `GUI_App::reload_user_presets_from_disk()` — snapshots existing user presets, calls `load_user_presets()`, detects new presets, and selects them via `Tab::select_preset()` for proper UI update
- `MainFrame::init_menubar_as_editor()` — adds File → Reload Presets menu item with Cmd+R shortcut
- `Tab` — adds refresh button between preset dropdown and save/delete buttons on all tabs
- Custom `reload_preset.svg` icon matching the existing monochrome style

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -144,6 +144,7 @@ function(bambustudio_add_cmake_project projectname)
             -DCMAKE_C_COMPILER:STRING=${CMAKE_C_COMPILER}
             -DCMAKE_CXX_COMPILER:STRING=${CMAKE_CXX_COMPILER}
             -DCMAKE_TOOLCHAIN_FILE:STRING=${CMAKE_TOOLCHAIN_FILE}
+            -DCMAKE_POLICY_VERSION_MINIMUM=3.5
             -DBUILD_SHARED_LIBS:BOOL=OFF
             ${_cmake_osx_arch}
             "${_configs_line}"
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -144,7 +144,6 @@ function(bambustudio_add_cmake_project projectname)
             -DCMAKE_C_COMPILER:STRING=${CMAKE_C_COMPILER}
             -DCMAKE_CXX_COMPILER:STRING=${CMAKE_CXX_COMPILER}
             -DCMAKE_TOOLCHAIN_FILE:STRING=${CMAKE_TOOLCHAIN_FILE}
-            -DCMAKE_POLICY_VERSION_MINIMUM=3.5
             -DBUILD_SHARED_LIBS:BOOL=OFF
             ${_cmake_osx_arch}
             "${_configs_line}"
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -147,7 +147,6 @@ function(bambustudio_add_cmake_project projectname)
             -DCMAKE_C_COMPILER:STRING=${CMAKE_C_COMPILER}
             -DCMAKE_CXX_COMPILER:STRING=${CMAKE_CXX_COMPILER}
             -DCMAKE_TOOLCHAIN_FILE:STRING=${CMAKE_TOOLCHAIN_FILE}
-            -DCMAKE_POLICY_VERSION_MINIMUM=3.5
             -DBUILD_SHARED_LIBS:BOOL=OFF
             ${_cmake_osx_arch}
             "${_configs_line}"
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -147,6 +147,7 @@ function(bambustudio_add_cmake_project projectname)
             -DCMAKE_C_COMPILER:STRING=${CMAKE_C_COMPILER}
             -DCMAKE_CXX_COMPILER:STRING=${CMAKE_CXX_COMPILER}
             -DCMAKE_TOOLCHAIN_FILE:STRING=${CMAKE_TOOLCHAIN_FILE}
+            -DCMAKE_POLICY_VERSION_MINIMUM=3.5
             -DBUILD_SHARED_LIBS:BOOL=OFF
             ${_cmake_osx_arch}
             "${_configs_line}"
--- a/resources/images/reload_preset.svg
+++ b/resources/images/reload_preset.svg
@@ -1,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M13.5 8C13.5 11.04 11.04 13.5 8 13.5C4.96 13.5 2.5 11.04 2.5 8C2.5 4.96 4.96 2.5 8 2.5C9.82 2.5 11.41 3.39 12.4 4.75L10.5 4.75" stroke="#262E30" stroke-linecap="round" stroke-linejoin="round"/>
+<polyline points="12.5,2.5 12.5,4.75 10.25,4.75" fill="none" stroke="#262E30" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
--- a/resources/images/reload_preset.svg
+++ b/resources/images/reload_preset.svg
@@ -1,4 +1,0 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M13.5 8C13.5 11.04 11.04 13.5 8 13.5C4.96 13.5 2.5 11.04 2.5 8C2.5 4.96 4.96 2.5 8 2.5C9.82 2.5 11.41 3.39 12.4 4.75L10.5 4.75" stroke="#262E30" stroke-linecap="round" stroke-linejoin="round"/>
-<polyline points="12.5,2.5 12.5,4.75 10.25,4.75" fill="none" stroke="#262E30" stroke-linecap="round" stroke-linejoin="round"/>
-</svg>
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -5887,17 +5887,22 @@ void GUI_App::reload_settings()
     }
 }
 
-// Reload user presets from disk without restart
-void GUI_App::reload_user_presets_from_disk()
+// Reload user presets from disk without restart. Returns counts of newly
+// added presets; the caller decides whether/how to notify the user. The
+// side UI is always refreshed because an in-place edit to an existing
+// preset is invisible to our name-diff, and the user just manually
+// triggered this action so a refresh is the expected behavior.
+PresetReloadResult GUI_App::reload_user_presets_from_disk()
 {
+    PresetReloadResult result;
     if (!preset_bundle || !mainframe)
-        return;
+        return result;
 
     std::string user_id = (m_agent && m_agent->is_user_login())
         ? m_agent->get_user_id()
         : DEFAULT_USER_FOLDER_NAME;
 
-    // Snapshot existing preset names before reload, to count new ones for the toast
+    // Snapshot existing user-preset names before reload, to count new ones.
     std::set<std::string> old_prints, old_filaments;
     for (const auto& p : preset_bundle->prints)
         if (p.is_user()) old_prints.insert(p.name);
@@ -5907,33 +5912,35 @@ void GUI_App::reload_user_presets_from_disk()
     BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
     preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
 
-    // Always refresh the side UI. We can't cheaply detect every kind of change
-    // (edit-in-place of an existing preset is invisible to the name-diff below),
-    // and the user has just manually triggered this action, so a refresh is the
-    // expected behavior regardless of what actually changed.
+    // Always refresh the side UI (see function comment).
     mainframe->update_side_preset_ui();
 
-    // Count newly added presets for the toast notification
-    int new_prints = 0, new_filaments = 0;
+    // Count newly added presets.
     for (const auto& p : preset_bundle->prints)
         if (p.is_user() && old_prints.find(p.name) == old_prints.end())
-            new_prints++;
+            ++result.new_prints;
     for (const auto& p : preset_bundle->filaments)
         if (p.is_user() && old_filaments.find(p.name) == old_filaments.end())
-            new_filaments++;
+            ++result.new_filaments;
 
+    BOOST_LOG_TRIVIAL(info) << build_reload_toast_message(result);
+    return result;
+}
+
+// Reproduce the toast message logic that used to live inline in
+// reload_user_presets_from_disk(). Exported so each call site can decide
+// whether to push a notification.
+std::string build_reload_toast_message(const PresetReloadResult& r)
+{
     std::string msg = "Presets reloaded";
     std::vector<std::string> parts;
-    if (new_prints > 0)
-        parts.push_back(std::to_string(new_prints) + " new process");
-    if (new_filaments > 0)
-        parts.push_back(std::to_string(new_filaments) + " new filament");
+    if (r.new_prints > 0)
+        parts.push_back(std::to_string(r.new_prints) + " new process");
+    if (r.new_filaments > 0)
+        parts.push_back(std::to_string(r.new_filaments) + " new filament");
     if (!parts.empty())
         msg += ": " + boost::algorithm::join(parts, ", ");
-
-    BOOST_LOG_TRIVIAL(info) << msg;
-    if (plater())
-        plater()->get_notification_manager()->push_notification(msg);
+    return msg;
 }
 
 //BBS reload when logout
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -5886,6 +5886,48 @@ void GUI_App::reload_settings()
     }
 }
 
+// Reload user presets from disk without restart
+void GUI_App::reload_user_presets_from_disk()
+{
+    if (!preset_bundle || !mainframe)
+        return;
+
+    std::string user_id = (m_agent && m_agent->is_user_login())
+        ? m_agent->get_user_id()
+        : DEFAULT_USER_FOLDER_NAME;
+
+    // Snapshot existing preset names before reload
+    std::set<std::string> old_prints, old_filaments;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user()) old_prints.insert(p.name);
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user()) old_filaments.insert(p.name);
+
+    BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
+    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
+    mainframe->update_side_preset_ui();
+
+    // Find new presets and select the last one found
+    std::string new_print, new_filament;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user() && old_prints.find(p.name) == old_prints.end())
+            new_print = p.name;
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user() && old_filaments.find(p.name) == old_filaments.end())
+            new_filament = p.name;
+
+    if (!new_print.empty()) {
+        BOOST_LOG_TRIVIAL(info) << "New process preset found, selecting: " << new_print;
+        auto tab = get_tab(Preset::TYPE_PRINT);
+        if (tab) tab->select_preset(new_print);
+    }
+    if (!new_filament.empty()) {
+        BOOST_LOG_TRIVIAL(info) << "New filament preset found, selecting: " << new_filament;
+        auto tab = get_tab(Preset::TYPE_FILAMENT);
+        if (tab) tab->select_preset(new_filament);
+    }
+}
+
 //BBS reload when logout
 void GUI_App::remove_user_presets()
 {
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -24,6 +24,7 @@
 #include <cstdlib>
 #include <chrono>
 #include <regex>
+#include <set>
 #include <thread>
 #include <string_view>
 #include <boost/algorithm/string/predicate.hpp>
@@ -5896,7 +5897,7 @@ void GUI_App::reload_user_presets_from_disk()
         ? m_agent->get_user_id()
         : DEFAULT_USER_FOLDER_NAME;
 
-    // Snapshot existing preset names before reload
+    // Snapshot existing preset names before reload, to count new ones for the toast
     std::set<std::string> old_prints, old_filaments;
     for (const auto& p : preset_bundle->prints)
         if (p.is_user()) old_prints.insert(p.name);
@@ -5905,27 +5906,34 @@ void GUI_App::reload_user_presets_from_disk()
 
     BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
     preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
+
+    // Always refresh the side UI. We can't cheaply detect every kind of change
+    // (edit-in-place of an existing preset is invisible to the name-diff below),
+    // and the user has just manually triggered this action, so a refresh is the
+    // expected behavior regardless of what actually changed.
     mainframe->update_side_preset_ui();
 
-    // Find new presets and select the last one found
-    std::string new_print, new_filament;
+    // Count newly added presets for the toast notification
+    int new_prints = 0, new_filaments = 0;
     for (const auto& p : preset_bundle->prints)
         if (p.is_user() && old_prints.find(p.name) == old_prints.end())
-            new_print = p.name;
+            new_prints++;
     for (const auto& p : preset_bundle->filaments)
         if (p.is_user() && old_filaments.find(p.name) == old_filaments.end())
-            new_filament = p.name;
+            new_filaments++;
 
-    if (!new_print.empty()) {
-        BOOST_LOG_TRIVIAL(info) << "New process preset found, selecting: " << new_print;
-        auto tab = get_tab(Preset::TYPE_PRINT);
-        if (tab) tab->select_preset(new_print);
-    }
-    if (!new_filament.empty()) {
-        BOOST_LOG_TRIVIAL(info) << "New filament preset found, selecting: " << new_filament;
-        auto tab = get_tab(Preset::TYPE_FILAMENT);
-        if (tab) tab->select_preset(new_filament);
-    }
+    std::string msg = "Presets reloaded";
+    std::vector<std::string> parts;
+    if (new_prints > 0)
+        parts.push_back(std::to_string(new_prints) + " new process");
+    if (new_filaments > 0)
+        parts.push_back(std::to_string(new_filaments) + " new filament");
+    if (!parts.empty())
+        msg += ": " + boost::algorithm::join(parts, ", ");
+
+    BOOST_LOG_TRIVIAL(info) << msg;
+    if (plater())
+        plater()->get_notification_manager()->push_notification(msg);
 }
 
 //BBS reload when logout
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -6255,6 +6255,48 @@ void GUI_App::reload_settings()
     }
 }
 
+// Reload user presets from disk without restart
+void GUI_App::reload_user_presets_from_disk()
+{
+    if (!preset_bundle || !mainframe)
+        return;
+
+    std::string user_id = (m_agent && m_agent->is_user_login())
+        ? m_agent->get_user_id()
+        : DEFAULT_USER_FOLDER_NAME;
+
+    // Snapshot existing preset names before reload
+    std::set<std::string> old_prints, old_filaments;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user()) old_prints.insert(p.name);
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user()) old_filaments.insert(p.name);
+
+    BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
+    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
+    mainframe->update_side_preset_ui();
+
+    // Find new presets and select the last one found
+    std::string new_print, new_filament;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user() && old_prints.find(p.name) == old_prints.end())
+            new_print = p.name;
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user() && old_filaments.find(p.name) == old_filaments.end())
+            new_filament = p.name;
+
+    if (!new_print.empty()) {
+        BOOST_LOG_TRIVIAL(info) << "New process preset found, selecting: " << new_print;
+        auto tab = get_tab(Preset::TYPE_PRINT);
+        if (tab) tab->select_preset(new_print);
+    }
+    if (!new_filament.empty()) {
+        BOOST_LOG_TRIVIAL(info) << "New filament preset found, selecting: " << new_filament;
+        auto tab = get_tab(Preset::TYPE_FILAMENT);
+        if (tab) tab->select_preset(new_filament);
+    }
+}
+
 //BBS reload when logout
 void GUI_App::remove_user_presets()
 {
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -6283,7 +6283,12 @@ PresetReloadResult GUI_App::reload_user_presets_from_disk()
     const std::set<std::string> old_printers  = snapshot_user_names(preset_bundle->printers);
 
     BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
-    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
+    // Match the other programmatic load sites (ConfigWizard, WebGuideDialog,
+    // PresetUpdater): substitute forward-compat values silently. This overload
+    // discards the substitution report anyway, so there is no user-facing
+    // difference vs Enable here; EnableSilentDisableSystem is just the
+    // conventional rule for a non-interactive load.
+    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::EnableSilentDisableSystem);
 
     // Always refresh the side UI (see function comment).
     mainframe->update_side_preset_ui();
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -6271,12 +6271,16 @@ PresetReloadResult GUI_App::reload_user_presets_from_disk()
         ? m_agent->get_user_id()
         : DEFAULT_USER_FOLDER_NAME;
 
-    // Snapshot existing user-preset names before reload, to count new ones.
-    std::set<std::string> old_prints, old_filaments;
-    for (const auto& p : preset_bundle->prints)
-        if (p.is_user()) old_prints.insert(p.name);
-    for (const auto& p : preset_bundle->filaments)
-        if (p.is_user()) old_filaments.insert(p.name);
+    // Snapshot existing user-preset names before reload, to diff afterwards.
+    auto snapshot_user_names = [](const auto& collection) {
+        std::set<std::string> names;
+        for (const auto& p : collection)
+            if (p.is_user()) names.insert(p.name);
+        return names;
+    };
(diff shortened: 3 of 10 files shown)


mpaperno:
Hey this is great, exactly what I wanted to add as well!  Thanks for working on it.

I'm not sure I understand the part with selecting the last newly found preset after a reload, and haven't had a chance test (sorry!).  This may change what is currently selected by the user?  If I'm understanding correctly... personally I wouldn't want that at all.  If the currently selected preset was updated, then the UI should reflect those changes, but I think it's weird if it arbitrarily selects some newly added preset... would it even be compatible? And "last" in this case is somewhat arbitrary (sorted alphabetically by name IIRC).  But maybe I misunderstand the intention?

Thanks again,
-Max

adele-with-a-b:
Hey Max!

The original intent of updating the dropdown selection was a convenience for my own workflow: I generate preset JSON files externally (via scripts/LLM), hit Cmd+R, and wanted to immediately see the new preset, plus it gives a visual feedback that something actually happened. But
you're correct that this might not be a widespread use case:

- It silently changes the user's active selection. If there are unsaved changes in the active process it won't get rid of all of them but the ones that intersect with the new preset would be overwritten
- The "last" new preset is arbitrary (depends on iteration order)
- If the new preset is not compatible with the current printer/nozzle config it won't show in the dropdown until you change, for instance, the nozzle size. It inherits the configuration compatibility from the system preset set in the "inherits" key.

I think I'd be ok with striping out the auto-select logic and keep the reload to just: reload from disk → refresh UI → keep current selection. If the currently selected preset was modified on disk,
the UI will reflect those changes. If it was deleted, it can fall back to the default. I might just add a quick pop up with maybe a count of loaded presets or some sort of indicator that something happened. Open to suggestions if you give it a try and have some extra feedback.

Best,
Abdel.

mpaperno:
Hi Abdel, thanks for your response!

Both of them  :-)  And that screenshot was helpful.  I'm also very interested in improving the preset management situation, and using external tools (even a text editor) is a quick step forward w/out needing any changes in Studio. And this reload function is key to making that more practical.  👍🏼   

I even just recently started a thread on the BL forums about editing files manually and using inheritance features: 
https://forum.bambulab.com/t/bambu-studio-presets-management-utilizing-inheritance-features/239148

> It silently changes the user's active selection. 

Yea that was my first thought.  Kinda one of my "golden rules" not to do.  Not that I like rules usually.  (Have you ever read the book _Don't Make Me Think_?  Great short little book about UI design.)

And your other points all speak against that mechanism as well, I agree. Although I get your use case (adding a preset externally and you want to use it right away, IIUC), there would be other usage scenarios, like adding a bunch of presets at once (something I'm probably more likely to do, myself).  And maybe the user doesn't even remember that they had added those (also likely scenario for me personally 😃 , or some script went haywire or whatever).

Visual feedback would be good, indeed, and some stats would be a nice bonus too. The "newly added is selected" only works when a preset is added anyway (vs. edited), so that doesn't really work in its favor either.  I'm not a big fan of popups though (plus I think that would work against one of your goals here to make the process as seamless as possible).  What about the "toast" notifications that show up at bottom right?  Nice and unobtrusive.  (I haven't looked at that code at all yet, so I have no idea how simple of a suggestion this is, or not.)

I also have a question about that reload button, next to the process profile selector.  I'm not sure it makes sense there...?  As a user, visually I'm not really sure what it will do at first -- reload the current profile, it looks like, but that's weird since you're there editing it why would it need a reload? The other buttons on the same row act on the current profile, so one may expect reload to do the same (and not expect it to reload everything).  And then it also reloads filaments and machines, so that's entirely unexpected (until/unless one reads the tooltip, but "no one reads" is my usual assumption).

IMHO the _File_ menu entry is enough, especially with the shortcut.  But I'd move it down below "Batch Preset Management" instead of in the middle of the "save" options. And also get it to show up for Mac users. (The current Mac vs. rest preprocessor distinctions in that code are ugly, just for one difference in a variable(!), but that's another matter and I think you can avoid duplicating it anyway.)

As a future addition it could be cool if a preset reload could be triggered externally, for example by whatever script or app is being used to edit them.  And maybe even granular reload (a specific profile) to make it more efficient.

OK that got long, sorry!  LMK what you think!

Thanks,
-Max

adele-with-a-b:
Hey Max, great feedback, really appreciate the thoroughness!

I'm on board with pretty much everything you've laid out. Here's the plan:

**Auto-select removal** — Agreed, that needs to go. Silently changing the user's active selection violates the principle of least surprise. I'll strip that logic out and just reload in place. When I thought of this feature and started implementing it I had no intent on posting a PR 😬

**Toast notification** — I used a misnomer but that's what I had pictured. Something like "Reloaded presets: 2 new, 1 modified" in the bottom-right toast is the right level of feedback without being intrusive. I'll look into the existing toast system and wire it up.

**Remove the reload button from the Tab toolbar** — You're right that it's misleading next to the profile-specific buttons. The File menu entry + shortcut is sufficient. I was just lazy

**Move the menu entry** — I'll relocate it below Batch Preset Management where it fits better semantically.

**Cross-platform visibility** — So funny enough, it's actually the opposite of what you might expect — the menu item currently only exists in the Mac code path. I'll pull it out of the platform-specific `#ifdef` block so it shows up everywhere. I have no way of testing it on Windows, or Linux for the time being (my instance has no GUI but maybe I'll set it up just to test this out)

**External trigger** — Interesting idea for the future. Could see a file watcher or even a simple IPC mechanism working well there.

---

~One more thing I'm thinking about as a follow-up: an **Import Presets** menu item. The idea is you select one or more preset JSON files, and the app handles the rest — detects whether each is a filament or process preset, validates it, generates the companion `.info` file with the correct `base_id`, and drops everything into the right user preset directory. Then it triggers a reload automatically. This would make the external editing workflow you're describing in your forum thread a lot more accessible — users wouldn't need to know the directory structure or `.info` file format at all. Thinking of that as a separate PR that builds on this one.~ 😂 OMG that's already there!!!!! (Goes to show how little I know about this app)

I'll get started on the changes. Thanks again for the detailed review!

PS: Is there a magical way of circumventing that very annoying network plugin so I can actually print from my fork? 😭

mpaperno:
Sounds awesome, thanks!

Happy to test it out on Windows here. Haven't thrown it on any Linux system yet, but that's in the plan.  For me the MacOS versions are usually the pain point for testing... and this one in particular really needs serious video resources to be usable.  😐 

Yea noticed the logic reversal on those macros once I looked in the editor... OMG, the smell.  Definitely need to stay on one's toes in this code base.  🙄  The more I look the more surprised I am that it works at all (missing nullptr and index bounds checks in critical areas, for example).

IIRC I got the network to work by using the latest beta version plugin (or for the latest beta version of Studio, I mean).  It's tagged as v. 02.05.01.52 in the downloads (so the Windows filename is `Studio-connector-win_02.05.01.52.zip`), which I think matches the current build version in the source code.  IIRC I "hacked" the download URL to get it 'cuz installing the beta just for that would have been a pain... though now I can't remember where I got the URL in the first place. From the error in the log maybe, when it failed to do it the automatic way.  The URL formatter is in `GUI_App.cpp @ 1545: get_plugin_url()`, FWIW.

And yea... it's a massive program, lots to discover!

Ooops.. speaking of which... did you know there's a _Help -> Keyboard Shortcuts_ window?  I didn't until I noticed it in the code... lol.  But it claims in there that <kbd>CTRL+R</kbd> is used for "Slice plate" in the "Global" group.  I haven't checked if that's true (or even where it gets that list from)... but that would be good to resolve probably.

Thanks again, and looking forward to the update!
-Max

PS. I only use the printer/Studio in "LAN Mode" so I can't speak to the whole cloud thing with the network plugin.  Don't even get me started on the cloud thing... LOL.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5686,48 @@ void GUI_App::reload_settings()
...
+{
+    if (!preset_bundle || !mainframe)
+        return;
+
+    std::string user_id = (m_agent && m_agent->is_user_login())
+        ? m_agent->get_user_id()
+        : DEFAULT_USER_FOLDER_NAME;
+
+    // Snapshot existing preset names before reload
+    std::set<std::string> old_prints, old_filaments;
```
weishibambu: The "set" is not included. The repo may build failed in Mac, it's better to include the "set" in this cpp file.

mpaperno:
@weishibambu You might wish to read our previous conversation here which covers some of the comments you're making.  I believe @adele-with-a-b  is going to be working on updating this PR.

https://github.com/bambulab/BambuStudio/pull/9919#issuecomment-4034596619

Thank you,
-Max

mpaperno:
Hi Abdel,

The first of your new commits looks great, just as discussed, nice!

But did you mean to push all the following ones here also?

Thanks!
-Max

adele-with-a-b:
Hey Max,

Was just playing around with some improvements and got carried away: they are here https://github.com/adele-with-a-b/BambuStudio/tree/preset-explorer if you wanna take a look/contribute some ideas.

Best!
Abdel.

adele-with-a-b:
Hey @weishibambu, thanks for the review! All points addressed in 835e2a1:

1. **`deps/CMakeLists.txt`** — Removed. That was a workaround for building with cmake 4.x (which errors on old `cmake_minimum_required` calls in dependency source trees). Agreed it doesn't belong in the repo — it's a local build environment issue.

2. **Missing `#include <set>`** — Added. It was compiling on my end via transitive includes but that's not portable. Good catch.

3. **`update_side_preset_ui()` called unconditionally** — Moved after the new-preset detection loop and now only called when `new_prints > 0 || new_filaments > 0`.

4. **Auto-selecting new presets** — Already removed in the previous commit (2a1709e).

5. **Menu item Mac-only** — Already fixed cross-platform in the previous commit (2a1709e). It now shows on all platforms with Ctrl+Shift+R / Cmd+Shift+R.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5687,58 @@ void GUI_App::reload_settings()
...
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user() && old_prints.find(p.name) == old_prints.end())
+            new_prints++;
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user() && old_filaments.find(p.name) == old_filaments.end())
+            new_filaments++;
+
+    // Only update UI if presets changed
+    if (new_prints > 0 || new_filaments > 0)
+        mainframe->update_side_preset_ui();
```
mpaperno: I think I would always reload the UI, w/out any condition, because it would be complex to check for all types of change that could have happened with the profiles.  Some could have been removed, or just edited. If anyone is manually reloading configs, I think it is likely because they know something changed in there, so a refresh seems more likely to be needed than not.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5687,58 @@ void GUI_App::reload_settings()
...
+
+    // Show toast notification with reload stats
+    std::string msg = "Presets reloaded";
+    if (new_prints > 0 || new_filaments > 0) {
```
mpaperno: Suggestion:  

```cpp
std::string msg = "Presets reloaded";
std::vector<std::string> parts;
if (new_prints > 0)
    parts.push_back(std::to_string(new_prints) + " new process");
if (new_filaments > 0)
    parts.push_back(std::to_string(new_filaments) + " new filament");
if (!parts.empty())
    msg += ": " + boost::algorithm::join(parts, ", ");
```

`boost::algorithm::join` should already be included via `boost/algorithm/string.hpp`

JohnSmithCYM:
Thanks for the improvement. Reloading presets without restarting Studio is useful in some cases but I have a few suggestions:
1. It targets an advanced workflows and it might be better to treat it as an advanced feature and place it better in locations like Batch Preset Management.
2. 'Reload Presets from Disk' seems better than 'Reload Local Presets' in this case.

mpaperno:
I do like it in the menu, or with a shortcut at least, but I agree it's "advanced" and would simplify the "basic" UI a bit to exclude it. Not sure if the menu or shortcuts config reloads when switching modes, but having to restart would be fine I think.

Placing it in some other place as well, like the batch preset management is also a good idea, a logical place for that.  I haven't look into what that means in terms of reloading the batch preset UI, but OTOH it would provide immediate feedback about the reloaded profiles, which is really nice.

Agree on the title change also, it is clearer.

Thanks!

mpaperno:
I just realized (while working on another feature) that this is missing the actual shortcut handler, right?  (I'm too used to Qt where a lot of this is handled in one call.)

Something like this seems to capture the shortcut at the top level (adding to `MainFrame`'s `wxEVT_CHAR_HOOK` handler):

```diff
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -664,6 +664,10 @@ MainFrame::MainFrame()
             return;}
 #endif
         if (evt.CmdDown() && evt.GetKeyCode() == 'R') { if (m_slice_enable) { wxGetApp().plater()->update(true, true); wxPostEvent(m_plater, SimpleEvent(EVT_GLTOOLBAR_SLICE_PLATE)); this->m_tabpanel->SetSelection(tpPreview); } return; }
+        if (evt.CmdDown() && evt.ShiftDown() && evt.GetKeyCode() == 'R') {
+            wxGetApp().reload_user_presets_from_disk();
+            return;
+        }
         if (evt.CmdDown() && evt.ShiftDown() && evt.GetKeyCode() == 'G') {
             m_plater->apply_background_progress();
             m_print_enable = get_enable_print_status();
```

mpaperno:
@adele-with-a-b  Hi, are you planning on revisiting this PR?  Would still be a great feature.  I can take over if you don't have time or inclination to continue it (will try to keep your contributions via cherry-pick).  Just let me know.

Thanks,
-Max

adele-with-a-b:
> @adele-with-a-b Hi, are you planning on revisiting this PR? Would still be a great feature. I can take over if you don't have time or inclination to continue it (will try to keep your contributions via cherry-pick). Just let me know.
> 
> Thanks, -Max

Hey Max,

Yes! I had added a ton of related feature work but then got really busy with my 9-5. Just got back from a conference in SF but I've addressed all the comments locally. Just trying to figure out whether I should make the rest of the features related to presets that I implemented part of this PR or open a new one. I'll update it by EOW.

Thanks,
Abdel.

adele-with-a-b:
Pushed an update addressing all open feedback. Current state:

**From @mpaperno:**
- `update_side_preset_ui()` now runs unconditionally after reload — edit-in-place changes aren't visible to a name-diff, and the user manually triggered the action, so refresh is the expected behavior.
- Toast message refactored with `boost::algorithm::join` per your suggestion.
- `Cmd+Shift+R` / `Ctrl+Shift+R` handler wired up in `MainFrame::wxEVT_CHAR_HOOK` (was only a menu accelerator label before). Placed before the plain `Cmd+R` slice handler so the more specific chord wins.
- Deleted the orphaned `resources/images/reload_preset.svg`.

**From @JohnSmithCYM / @mpaperno:**
- Removed the File menu entry entirely. Reload now lives in Batch Preset Management (advanced UI) with the shortcut still available.

**New Reload button in `UserPresetsDialog`:**
- Labeled "Reload from Disk", placed in the bottom bar next to Delete. Calls `reload_user_presets_from_disk()` then rebuilds the dialog state via new `reload_presets_ui()` (clears cached sizer maps, re-runs `init_preset_list` + `layout_preset_list(true)`, re-applies `UpdateDarkUIWin(m_scrolled)` to style the newly-created Labels — same pattern as existing `on_collection_changed`).
- Rebuilt the button's `background_color` with `setTakeFocusedAsHovered(false)` so the button doesn't render in the Hovered gray when focused-but-unhovered after click.

Rebased on current master. Verified end-to-end on H2C.

lanewei120:
Hi @adele-with-a-b 

thanks for the great job~

Hi @mpaperno 

do you have any other suggestions?
It looks good to us now

mpaperno:
Hi Abdele, thanks for all the updates! Looking good!

Sorry for the delay.  I'd like to pull this into my dev tree and build to give it proper test.  Some comments follow...  and I don't mean to be pedantic about any of this, just my honest/humble opinions.  I really appreciate the time and effort it takes to prepare these PRs!

> Removed the File menu entry entirely. Reload now lives in Batch Preset Management (advanced UI) with the shortcut still available.

My only issue with this is that the shortcut is now entirely "secret," no way to know about it (as far as I can tell). It could be added to the "Shortcuts" help dialog (`GUI/KBShortcutsDialog.cpp`) which would help.  And maybe to the tooltip on the new reload button (assuming the `MainFrame` shortcut handler will work when that window is focused... might not).

> reload_user_presets_from_disk()

I see this shows the toast notification... does that work OK when using the presets dialog?  (Like I said I need to try it.)  I saw in an iteration of your notes something about that, but can't find it now.  If it's not an issue, great.  But another suggestion I'd have would be to remove any UI updates from `reload_user_presets_from_disk()` and have it return a status... like a simple structure with result counts. Then whatever triggered the reload could decide how to handle UI notification (possibly with a separate shared method for assembling the message... or have the message returned from `reload_user_presets_from_disk()` as part of the result structure).

> Rebuilt the button's background_color

Will leave a suggestion on that in the code, perhaps it will be simpler (but untested).

Thanks again!
-Max

src/slic3r/GUI/UserPresetsDialog.cpp
```diff
@@ -71,6 +71,24 @@ UserPresetsDialog::UserPresetsDialog(wxWindow *parent)
     auto label = new Label(this, _L("Select All"));
     m_label_check_count = new Label(this);
     m_label_check_count->SetForegroundColour("#6B6B6B");
+    m_button_reload     = new Button(this, _L("Reload from Disk"));
+    m_button_reload->SetToolTip(_L("Reload user presets from disk without restarting Bambu Studio"));
+    // Bambu's Button defaults to takeFocusedAsHovered=true on its background
```
mpaperno: Would this be simpler handled by forcing a focus loss on the button, in the click event handler (below)? Something like `this->SetFocus();` to shift the focus to the button's parent (or some other widget if that makes more sense). Before preforming the reload and UI update.

adele-with-a-b: I'm embarrassed to even talk about the amount of time that I spent trying to debug this. Nothing I tried worked (and yes I had tried setting the focus on a different widget but I can't remember exactly how I was doing it and it wasn't working). The only way I got it to work was setting the focus on the search bar. Anyway, I took another pass at it and it's now working as it should

adele-with-a-b:
> Hi Abdel, thanks for all the updates! Looking good!
> 
> Sorry for the delay. I'd like to pull this into my dev tree and build to give it proper test. Some comments follow... and I don't mean to be pedantic about any of this, just my honest/humble opinions. I really appreciate the time and effort it takes to prepare these PRs!
> 
> > Removed the File menu entry entirely. Reload now lives in Batch Preset Management (advanced UI) with the shortcut still available.
> 
> My only issue with this is that the shortcut is now entirely "secret," no way to know about it (as far as I can tell). It could be added to the "Shortcuts" help dialog (`GUI/KBShortcutsDialog.cpp`) which would help. And maybe to the tooltip on the new reload button (assuming the `MainFrame` shortcut handler will work when that window is focused... might not).

Added to the help dialog and the tooltip. It reads: "Reload user presets from disk without restarting [cmd+shift+R]"

While trying to look into this I also found out that the cmd+shift+R shortcut conflicts on Mac with "reload g-code from disk". We have to come up with a different one. Some ideas:
* Cmd+Shift+P (presets)
* Cmd+Shift+L (load)
* Cmd+Option+R  (kinda unconventional and also uncomfortable)

> 
> > reload_user_presets_from_disk()
> 
> I see this shows the toast notification... does that work OK when using the presets dialog? (Like I said I need to try it.) I saw in an iteration of your notes something about that, but can't find it now. If it's not an issue, great. But another suggestion I'd have would be to remove any UI updates from `reload_user_presets_from_disk()` and have it return a status... like a simple structure with result counts. Then whatever triggered the reload could decide how to handle UI notification (possibly with a separate shared method for assembling the message... or have the message returned from `reload_user_presets_from_disk()` as part of the result structure).

Yes the toast notification works perfectly fine while using the presets dialog. The one caveat is that the toast won't dismiss until after you exit the dialog. If there are other notifications they will stack.

> 
> > Rebuilt the button's background_color
> 
> Will leave a suggestion on that in the code, perhaps it will be simpler (but untested).
> 
> Thanks again! -Max

Just finishing up some smoke testing and will publish shortly. Let me know if there's anything else before I commit



---

**UPDATE 2026-05-10:** Sorry all, this PR was accidentally closed on 2026-05-07 — I clicked "Comment and close" instead of "Comment" when posting the reply above. Reopened now; all four review items have been pushed as separate commits on top of `3b4fa7355` for easy per-concern review:

| Commit | Review concern addressed |
|---|---|
| [`cf3aeaa`](https://github.com/bambulab/BambuStudio/pull/9919/commits/cf3aeaa4f371e19e9e35ca4ad99f8eef69bf7f45) — "Address review feedback: use SetCanFocus(false) for Reload button" | @mpaperno's suggestion to force focus loss in the click handler. Went one step further — `SetCanFocus(false)` at construction, matching the established Bambu pattern (`Plater::btn_sync`, `MediaPlayCtrl::m_button_play`, `CalibrationWizardPresetPage::m_btn_sync`, etc.). Net -8 lines, removes the duplicated StateColor rebuild. |
| [`f4d9416`](https://github.com/bambulab/BambuStudio/pull/9919/commits/f4d94169ba30f3bf344eea6b98ad2274a4a3f14c) — "Surface Reload Presets shortcut in KBShortcutsDialog and tooltip" | @mpaperno's "shortcut is now entirely secret" concern. Added the entry to `KBShortcutsDialog.cpp` Global shortcuts page next to the File-group shortcuts, and appended the shortcut to the Reload button tooltip using the same pattern `GLCanvas3D` uses for its toolbar tooltips. |
| [`6b9ad5b`](https://github.com/bambulab/BambuStudio/pull/9919/commits/6b9ad5be10d483b69223ae28f11b12338081cb98) — "Refactor reload_user_presets_from_disk to return a result struct" | @mpaperno's "remove UI updates from reload_user_presets_from_disk, return status" suggestion. Introduced `PresetReloadResult { new_prints, new_filaments }` with `any_change()` helper in `GUI_App.hpp`, and extracted the toast-message assembly into a free helper `build_reload_toast_message()` shared by both call sites. |
| [`510ef8e`](https://github.com/bambulab/BambuStudio/pull/9919/commits/510ef8eab2cc0adbca2c243fabad9d008c4f37b7) — "Change Reload Presets shortcut from Cmd+Shift+R to Cmd+Shift+P" | The `Cmd+Shift+R` / `Ctrl+Shift+R` chord I'd picked conflicts with the existing `File > Reload from Disk` (G-code) accelerator at `MainFrame.cpp:3678`. Switched to `Cmd+Shift+P` — unused in BambuStudio, mnemonic for **P**reset, parallels the `Cmd+Shift+S` Save-As pattern of the neighboring File-menu shortcuts, doesn't collide with any macOS system-reserved chord. Updated all three reference sites in lockstep (`MainFrame.cpp` handler, `KBShortcutsDialog.cpp` entry, `UserPresetsDialog.cpp` tooltip). |

Happy to squash these down on your end at merge time, or to rebase/fixup locally if you'd prefer — let me know the house style. Thanks again @mpaperno and @lanewei120 for the patience and the detailed review.

mpaperno:
Looks great to me, thanks!  

👍🏼

adele-with-a-b:
Hi @weishibambu — friendly nudge on this one.

It's been ~10 weeks since your initial review. I've addressed all the comments and pushed several other updates, and got sign off from the other people that had previously engaged with the work. All `Build All` CI jobs (Linux, macOS arm64 + intel, Windows) are green; the only red check is  `Linux AppImage Build`, which was cancelled after the 24h timeout and appears unrelated to this PR's diff (same pattern hits other PRs).

The PR is `MERGEABLE` but `BLOCKED` waiting on an approving review. Could you take another look when you have a chance, or if you're no longer the right reviewer for this, point me to whoever is?

Thanks!

weishibambu:
> Hi @weishibambu — friendly nudge on this one.
> 
> It's been ~10 weeks since your initial review. I've addressed all the comments and pushed several other updates, and got sign off from the other people that had previously engaged with the work. All `Build All` CI jobs (Linux, macOS arm64 + intel, Windows) are green; the only red check is `Linux AppImage Build`, which was cancelled after the 24h timeout and appears unrelated to this PR's diff (same pattern hits other PRs).
> 
> The PR is `MERGEABLE` but `BLOCKED` waiting on an approving review. Could you take another look when you have a chance, or if you're no longer the right reviewer for this, point me to whoever is?
> 
> Thanks!

Okay, I'll review it as soon as possible.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5886,6 +5887,62 @@ void GUI_App::reload_settings()
...
+
+    // Snapshot existing user-preset names before reload, to count new ones.
+    std::set<std::string> old_prints, old_filaments;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user()) old_prints.insert(p.name);
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user()) old_filaments.insert(p.name);
+
+    BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
+    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
```
tonghao-bbl, line 5913: is `EnableSilentDisableSystem` more suiteble here?

src/slic3r/GUI/GUI_App.hpp
```diff
@@ -253,6 +253,20 @@ struct TryLoadLastMachine
     boost::thread local_bind_thread;
 };
 
+// Result of GUI_App::reload_user_presets_from_disk(). Exposes counts of
+// newly added presets so callers can decide whether/how to surface a
+// notification, instead of hardcoding the UI side-effect in the reload.
+struct PresetReloadResult {
```
tonghao-bbl, line 259: * only `filament` and `process/print` are recorded, is `machine/printers` ignored by intension?
* the removed presets are not counted, would be good to have if its trivial to get them

tonghao-bbl (commented):
Hi @adele-with-a-b !
can you check these small issues when you are available? Thanks!

adele-with-a-b:
@tonghao-bbl thanks for the review — both points are addressed, and I've also rebased onto the latest `master` (the branch had gone stale), so it's mergeable again.

1. **Machine/printers + removed presets** ([`41c6bb86d`](https://github.com/bambulab/BambuStudio/pull/9919/commits/41c6bb86d33c8dcfe0db0465be79bee2ae2ac13d)): `PresetReloadResult` now counts printer presets in addition to process and filament, and for all three kinds it also counts presets **removed** on disk since the last load. The reload toast reports removals alongside additions. In-place edits to an existing preset remain invisible to the name-diff by design — the side UI is refreshed unconditionally, so edits are still reflected.

2. **Substitution rule** ([`be3822a8b`](https://github.com/bambulab/BambuStudio/pull/9919/commits/be3822a8be241438912ec6974dbed334d8d6f195)): switched from `Enable` to `EnableSilentDisableSystem` to match the other non-interactive load sites (`ConfigWizard`, `WebGuideDialog`, `PresetUpdater`). This `load_user_presets` overload discards the returned substitution report, so there's no user-facing difference vs `Enable` here — it's simply the conventional rule for a programmatic load.

deps/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ function(bambustudio_add_cmake_project projectname)
             -DCMAKE_C_COMPILER:STRING=${CMAKE_C_COMPILER}
             -DCMAKE_CXX_COMPILER:STRING=${CMAKE_CXX_COMPILER}
             -DCMAKE_TOOLCHAIN_FILE:STRING=${CMAKE_TOOLCHAIN_FILE}
+            -DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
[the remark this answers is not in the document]
adele-with-a-b: Agreed, and this file is no longer part of the PR — the diff is now confined to `src/slic3r/GUI/`. Thanks for catching it.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5686,48 @@ void GUI_App::reload_settings()
...
+    // Snapshot existing preset names before reload
+    std::set<std::string> old_prints, old_filaments;
+    for (const auto& p : preset_bundle->prints)
+        if (p.is_user()) old_prints.insert(p.name);
+    for (const auto& p : preset_bundle->filaments)
+        if (p.is_user()) old_filaments.insert(p.name);
+
+    BOOST_LOG_TRIVIAL(info) << "Reloading user presets from disk for user: " << user_id;
+    preset_bundle->load_user_presets(user_id, ForwardCompatibilitySubstitutionRule::Enable);
+    mainframe->update_side_preset_ui();
```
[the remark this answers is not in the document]
adele-with-a-b: I went the other way here, and I want to flag it as a deliberate disagreement rather than an oversight: @mpaperno argued the opposite in this same review, that the UI should always be refreshed.

The reason I followed that: the diff this function computes is a **name** diff, so it only sees presets added or removed. A preset edited in place on disk keeps its name and is invisible to it. If the side-UI refresh were gated on the counts being non-zero, then editing an existing preset on disk and hitting reload would leave the side panel showing stale values — which is one of the main things this feature would get used for. So `update_side_preset_ui()` runs unconditionally, and the counts are used only to decide whether to show a notification.

Happy to gate it if you'd rather. That would mean replacing the name diff with real change detection (mtime or content hash) so edits are counted too.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5686,48 @@ void GUI_App::reload_settings()
...
+
+    // Find new presets and select the last one found
+    std::string new_print, new_filament;
+    for (const auto& p : preset_bundle->prints)
```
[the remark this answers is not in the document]
adele-with-a-b: Agreed, and nothing is selected. The reload loads the bundle and refreshes the side UI; there are no `select_preset` or `load_current_presets` calls anywhere in the PR, so whatever the user had selected stays selected.

src/slic3r/GUI/MainFrame.cpp
```diff
@@ -2658,6 +2658,11 @@ void MainFrame::init_menubar_as_editor()
         append_menu_item(fileMenu, wxID_ANY, _L("Save Project") + "\t" + ctrl + "S", _L("Save current project to file"),
             [this](wxCommandEvent&) { if (m_plater) m_plater->save_project(); }, "", nullptr,
             [this](){return m_plater != nullptr && can_save(); }, this);
+
+        // Reload user presets from disk
+        append_menu_item(fileMenu, wxID_ANY, _L("Reload Presets") + "\t" + ctrl + "R", _L("Reload user presets from disk"),
```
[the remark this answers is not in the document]
adele-with-a-b: Fixed — the Mac-only menu entry is gone. The feature is now reachable two cross-platform ways: a **Reload from Disk** button in the Batch Preset Management dialog, and a shortcut registered in `KBShortcutsDialog` as `ctrl + "Shift+P"`, which renders as Ctrl on Windows/Linux and Cmd on macOS. The handler tests `evt.CmdDown()`, which wx maps the same way. There is no `__WXMAC__` or `__APPLE__` guard left anywhere in the PR.

src/slic3r/GUI/GUI_App.cpp
adele-with-a-b: Adopted — `update_side_preset_ui()` is now called unconditionally, for exactly the reason you give. The counts are a name diff, so an in-place edit is invisible to them, and gating on the counts would leave the panel stale in precisely the case a user is most likely to hit. The counts now only decide whether to show a notification.

Noting for the record that this is the opposite of what @weishibambu suggested in a nearby thread; I've replied there laying out the tradeoff so the choice is visible.

src/slic3r/GUI/GUI_App.cpp
adele-with-a-b: Adopted essentially verbatim, thank you — `build_reload_toast_message()` builds a `parts` vector and joins with `boost::algorithm::join(parts, ", ")`. It goes a little past your version to also cover printers and removed presets, so the six labels come from a small `add_part` lambda rather than repeated `if` blocks.

src/slic3r/GUI/GUI_App.cpp
```diff
@@ -5686,6 +5687,58 @@ void GUI_App::reload_settings()
     }
 }
 
+// Reload user presets from disk without restart
+void GUI_App::reload_user_presets_from_disk()
```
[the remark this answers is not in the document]
adele-with-a-b: It would compile as `const`, but I'd rather not, and I think the reason is worth stating. Everything it mutates goes through member *pointers* — `preset_bundle->load_user_presets(...)`, `mainframe->update_side_preset_ui()` — and constness of the enclosing method only protects the pointer, not the pointee, so the compiler wouldn't complain. It would just be a `const` method that reloads the entire preset bundle and rebuilds UI, which reads as more misleading than helpful. Happy to add it if you disagree.

resources/images/reload_preset.svg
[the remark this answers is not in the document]
adele-with-a-b: Correct, and it's out of the PR now — the Tab-toolbar button that used it was dropped (it was confusing sitting next to the profile-specific buttons), so both the button and the resource are gone from the diff.

src/slic3r/GUI/GUI_App.cpp
adele-with-a-b: Yes — changed to `EnableSilentDisableSystem`, which matches the other non-interactive load sites (`ConfigWizard`, `WebGuideDialog`, `PresetUpdater`). There's no user-facing difference versus `Enable` here, because this `load_user_presets` overload discards the substitution report either way; it's simply the conventional rule for a programmatic load.

src/slic3r/GUI/GUI_App.hpp
adele-with-a-b: Both added. `PresetReloadResult` now counts printer presets alongside process and filament, and for all three kinds it also counts presets **removed** on disk since the last load; the notification reports removals next to additions. In-place edits to an existing preset remain invisible to the name diff by design — the side UI is refreshed unconditionally, so edits are still reflected.

Haidiye00 (commented):
Hello, I'd like to ask, what are the keyboard shortcuts for Windows? Currently, Ctrl+R is occupied by the slicing function

adele-with-a-b:
@Haidiye00 good catch, and the confusion was mine — the title still said "(Cmd+R)" and was out of date. I've just corrected it. It is no longer Cmd+R, for exactly the reason you give.

The shortcut is now **Ctrl+Shift+P** on Windows and Linux, **⌘+Shift+P** on macOS. The handler tests `evt.CmdDown() && evt.ShiftDown()` (`MainFrame.cpp:706`), and it's registered in the shortcuts dialog as `shortkey_ctrl_prefix() + "Shift+P"` so it renders per platform. There's also a **Reload from Disk** button in the Batch Preset Management dialog, so the feature isn't keyboard-only.

One thing worth flagging for Windows specifically. `Ctrl+P` there opens the Preferences dialog (`MainFrame.cpp:747`, the non-`__APPLE__` branch), and that condition tests only `CmdDown()` — it does not check `ShiftDown()`. So `Ctrl+Shift+P` is unambiguous today only because the reload handler sits earlier in the same key handler and returns before Preferences ever sees the event. That's an ordering dependency rather than an explicit one: if those two blocks were reordered, `Ctrl+Shift+P` would quietly start opening Preferences instead.

I'd be happy to add `&& !evt.ShiftDown()` to the Preferences condition so the pairing is explicit and order-independent rather than relying on handler order. Tell me which you'd prefer and I'll push it.

If Ctrl+Shift+P is itself unwelcome for any reason, I'm flexible on the binding — the only constraint I found was avoiding Ctrl+R (slice) and plain Ctrl+P (Preferences on Windows).

PR #9941, opened 2026-03-10
Fix CLI segfault (SIGSEGV) when using --info, --slice, or --export-3mf

niccolodevries:
## Summary

The BambuStudio CLI crashes with a segmentation fault (exit code 139) on **every** command that loads a model (`--info`, `--slice`, `--export-3mf`). This makes the CLI completely unusable.

**Root cause:** In CLI mode, `PartPlateList` is constructed with a `NULL` plater pointer (`BambuStudio.cpp:3917`). When `set_shapes()` is subsequently called, it unconditionally executes render data preparation code that dereferences the null plater through calls like `generate_print_polygon()` → `wxGetApp().plater()`.

**Fix:** Add `if (m_plater != nullptr)` guards around two render-only code blocks that are not needed in CLI mode:

- **`PartPlateList::set_shapes()`** — skip polygon generation, triangle calculation, gridlines, and icon vertex computation
- **`PartPlate::set_shape()`** — skip logo triangles, bed bounding box, and plate name texture computation

This has **zero impact on GUI mode** where `m_plater` is always a valid pointer. Only CLI mode (where `m_plater` is `NULL`) is affected.

## Steps to reproduce

```bash
# Any of these crash with exit code 139 (SIGSEGV):
BambuStudio --info model.stl
BambuStudio --slice 0 model.stl
BambuStudio --export-3mf output.3mf model.stl
```

## After fix

```
$ BambuStudio --info 3DBenchy.stl
[3DBenchy.stl]
size_x = 60.000999
size_y = 31.004000
size_z = 48.000000
...
number_of_facets = 225154
manifold = yes
volume = 15550.382812
```

### DIFF
--- a/src/slic3r/GUI/PartPlate.cpp
+++ b/src/slic3r/GUI/PartPlate.cpp
@@ -2885,24 +2885,26 @@ bool PartPlate::set_shape(const Pointfs& shape, const Pointfs& exclude_areas, co
 
 		calc_bounding_boxes();
 
-		ExPolygon logo_poly;
-		generate_logo_polygon(logo_poly);
-        auto triangles = triangulate_expolygon_2f(logo_poly, NORMALS_UP);
-        m_logo_triangles.reset();
-		if (!m_logo_triangles.init_model_from_poly(triangles, GROUND_Z + 0.01f))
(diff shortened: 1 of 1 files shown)


Haidiye00:
agree

lanewei120:
thanks for doing this~
@niccolodevries

PR #9955, opened 2026-03-12
Fix memory leak from allocated but uncleared object instance in `ConfigOptionDef::create_default_option()`

mpaperno:
The cloned instance wasn't deleted before the method returned, and not really needed in the first place.

Thanks,
-Max


lanewei120:
thanks~

Issue #9716, opened 2026-02-09
Crashing when setting Top Surface Flow


flessaaco:
### Bambu Studio Version

2.5.0.66

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11 (Build 26200)

### Printer

H Series (S and D)

### Printer Firmware Version

01.01.00.00

### How to reproduce

Model opens normally. Once I try to set the object specific top surface flow, Bambu Studio just closes. No crash warning.
When reopen Bambu offers to restore the file.

### Actual results

Bambu Studio just closes. No crash warning.
When reopen Bambu offers to restore the file. Flow is still the default (1)

### Expected results

Simply have the flow value changed.

### Project file & Debug log uploads

[Tramming 2mm.zip](https://github.com/user-attachments/files/25173017/Tramming.2mm.zip)

### Checklist of files to include

- [ ] Log file
- [x] Project file

JinGangBambu:
Hello,
Thanks for your feedback.
Could you please provide the log and .dmp files so that we can further investigate this issue?
The steps to obtain the logs and dmp file are as follows：
https://wiki.bambulab.com/en/x1/troubleshooting/bambu-studio-log.
https://wiki.bambulab.com/en/bambu-studio/troubleshoot/crash-freeze-issue

Thanks for your cooperation and patience.

flessaaco:
[bambu-studio.exe.19708.zip](https://github.com/user-attachments/files/25191785/bambu-studio.exe.19708.zip)

Here you go!
Thanks

JinGangBambu:
Hello,

We have confirmed this issue and will address it in the next release.

Thank you very much for your support and understanding.

flessaaco:
Thank you for working on it

mpaperno:
Hello @JinGangBambu & @wtji-bbl ,

Did you identify the cause of the crash?

I ran into this while making some unrelated changes, and spent quite a while debugging it.  Should have checked the reported issues first apparently.  If you already have fixes, it would be great to see them in the public repo here.

Anyway, in case it helps, I traced it to here: 
https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/slic3r/GUI/Tab.cpp#L3331

Because for `top_solid_infill_flow_ratio` there is no mapping in `m_opt_id_map`.  Result UB when creating the new string.

[...]

PR #9956, opened 2026-03-12
Refactor GUI parameter entry validation to allow floating point minimum/maximum values instead of just integers.

mpaperno:
Hello,

As I mentioned in #9925, there is an issue/limitation with UI parameter entry validation for floating point types. The current min/max for a `ConfigOptionDef` are integer types, which of course can't represent something like `0.1`.  There are several places in the current code that actually try to set a default like that, but it doesn't work. For example:
https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/libslic3r/PrintConfig.cpp#L1911

https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/libslic3r/PrintConfig.cpp#L4442

The minimum ends up being zero. And if a user does enter an invalid value (eg. negative or too large), it will auto-correct to zero.

Thee are some other parameters which could benefit from having a non-zero but < 1 minimum, I think, like some of the other flow rates for example.

This turned out to be even simpler than I thought because most of the `Field` derivatives already treated the min/max as `double` types anyway, relying on implicit conversion from int.  For example:

https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/slic3r/GUI/Field.cpp#L291  (and the whole block below that; `val` is a `double` here)

https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/slic3r/GUI/Field.cpp#L2069

I did search extensively and it seems that `ConfigOptionDef::min` and `max` are _only_ used with UI fields and not any other internal validation.
EDIT: also used to validate the config outside of GUI, eg. from CLI, here: https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/libslic3r/PrintConfig.cpp#L8671

Thanks for your consideration,
-Max

---

Related:
#9716
#9976

### DIFF
--- a/src/libslic3r/Config.hpp
+++ b/src/libslic3r/Config.hpp
@@ -3,6 +3,7 @@
 
 #include <assert.h>
 #include <map>
+#include <cfloat>
 #include <climits>
 #include <cstdio>
 #include <cstdlib>
@@ -2384,10 +2385,13 @@ public:
     // Optional width of an input field.
     int                                 width           = -1;
     // <min, max> limit of a numeric input.
-    // If not set, the <min, max> is set to <INT_MIN, INT_MAX>
+    // If not set explicitly, the <min, max> default to <-DBL_MAX, DBL_MAX>, which can be refenced externally
+    // using the static `min_default`/`max_default` members to check if a value was set.
     // By setting min=0, only nonnegative input is allowed.
-    int                                 min = INT_MIN;
-    int                                 max = INT_MAX;
+    static constexpr double             min_default = -DBL_MAX;
+    static constexpr double             max_default = DBL_MAX;
+    double                              min = min_default;
+    double                              max = max_default;
     // To check if it's not a typo and a % is missing
     double                              max_literal = 1;
     ConfigOptionMode                    mode = comSimple;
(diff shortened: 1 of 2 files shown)


mpaperno:
Correction:  The min/max values _are_ used for config validation outside of the GUI, but there are no conflicts from the proposed changes here. I forgot about them earlier, sorry.

https://github.com/bambulab/BambuStudio/blob/6a3e8bb3888b549df716a56266c1f6f3d0259012/src/libslic3r/PrintConfig.cpp#L8671

mpaperno:
Found equivalent (or very similar) changes at Orca:
https://github.com/OrcaSlicer/OrcaSlicer/commit/f277bc80c22e0c9a067481a4301922e2c96aed47
merged as part of https://github.com/OrcaSlicer/OrcaSlicer/commit/4343ddf8234b66714d5da1302164ecf549b148f2

lanewei120:
thanks~

PR #9962, opened 2026-03-13
FIX: Preserve nozzle grouping metadata through Helio Enhance for H2C printers (#53)

HelioPri:
After Helio Enhance, the GCodeProcessor result was missing nozzle_group_result and used_filaments, causing H2C firmware to fail with "failed to build nozzle auto-mapping table". Preserve these fields from the original slicing result through the Helio processing path, and clear stale nozzle data on processor reset.


lanewei120:
already included

Issue #9965, opened 2026-03-13
Studio Crashes after Copy-pasting item across files


lucas-28:
### Bambu Studio Version

2.5.0.66

### Where is the application from?

Bambu Lab Official website

### OS version

Win 11

### Printer

A1 Mini

### Printer Firmware Version

N/A

### How to reproduce

Go to file A. Copy object (I used Road Tire.stl (and I tried removing the space from the object name) from a plate.

Go to file B. Attempt to Paste item.

### Actual results

Get error message shown below: "Unhandled unknown exception". App crashes.

<img width="597" height="246" alt="Image" src="https://github.com/user-attachments/assets/4c31af40-e297-4a48-83c8-048d11bda961" />

At crash time, max CPU usage was 20% and Memory was 55%.

### Expected results

Item pastes successfully in File B. I tried a different set of files and it worked fine.

### Project file & Debug log uploads

[studio_Fri_Mar_13_11_31_34_52088_enc.log.zip](https://github.com/user-attachments/files/25975152/studio_Fri_Mar_13_11_31_34_52088_enc.log.zip)

File A: [M3A3 Bradley - A1 Mini.zip](https://github.com/user-attachments/files/25975168/M3A3.Bradley.-.A1.Mini.zip)

File B: [M1A2_Abrams_A1_Mini.zip](https://github.com/user-attachments/files/25975171/M1A2_Abrams_A1_Mini.zip)

### Checklist of files to include

- [x] Log file
- [x] Project file

mpaperno:
Confirmed with current `master` code base.

@lucas-28  Thanks for discovering and reporting!

What is causing this is that the tire selected in File A uses filament number 7.  The crash happens when that is pasted into File B, which only has 4 filaments in the configuration. Studio tries to assign "filament 7" to the pasted part, but that filament doesn't exist.

Lucas, if you need a workaround, before copying that tire from the first file, change its filament to a lower number, like 4 (which should also match the TPU filament in File B).

-Max

lucas-28:
Hello sir, thank you for the quick response and workaround. - Lucas Pfeifer

PR #9763, opened 2026-02-14
fix crash on h2c due to check_tpu_nozzle_has_multiple_filaments 

raulp:
While trying to open the following 3mf (downloaded from https://makerworld.com/en/models/2338301-m26-pershing-tank-united-states-army-1-35-scale?from=search#profileId-2555795), swapping from A1 from profile to H2C caused my instance to crash.
Doing some debug, found that `get_extruders(true)` produced `filament_idx=9` while `filament_map.size()==3`.
Seems like this function was introduced [here](https://github.com/bambulab/BambuStudio/commit/a3fd02105e2610d6c8805f16c89455a833cf3f63#diff-82f110ccb7ec6f3345a7c304d108075a1c462e811ffb957a8570643bafe8baa4) and a known H2C bug was discovered early, but not fully fixed [here](https://github.com/bambulab/BambuStudio/commit/63b1097485b3439d0ccaa82705a50405b73a820d) and [here](https://github.com/bambulab/BambuStudio/commit/6535ffccbd72230f93f39cacf3a003889c81d793)

The fix worked for me, and i was able to open broken models. I hope it's the correct fix.

PR #9978, opened 2026-03-15
Prevent access violation due to out of bounds index when trying to map to non-existing filament

mpaperno:
Fixes https://github.com/bambulab/BambuStudio/issues/9965 

This prevents a crash, and it is always good to check for index bounds. (hint hint)

However I'm not sure if the filament index from the pasted object was supposed to be validated earlier in the paste object sequence somewhere.... not that I saw, but perhaps there's a separate issue in there somewhere.

Simple fix, if you choose to implement it in your internal version directly, just let me know and go ahead and close this.

Thanks,
-Max


mpaperno:
Duplicate of #9763 , sorry didn't realize earlier.   Please merge that to fix this exception.

PR #9979, opened 2026-03-15
Restore `--datadir` CLI arg. functionality & add local configuration folder search for easily portable Studio instances.

### LINKED ISSUE #1651: Load Profiles from Git Repository
I used to store my slicer profiles in a Git repository to keep track of changes and easily know when I made changes, why and to roll them back. All advantages from a version history apply here.

Slic3r/PrusaSlicer/Superslicer use to have a parameter "-datadir" that allows to open the application with an explicit path from where it should load/save the profiles. This could be stored in a soft link for example to start the application. That way when changes are made, they are tracked via git and can be committed to a repository. However, this does not work with Bambu Studio.

An alternative could be that you set a setting where profiles are stored by the application. 

**Additional context**
Issue is for example explained here: https://github.com/prusa3d/PrusaSlicer/issues/5023

mpaperno:
Hello,

**Issue**

Studio can't be easily run as a truly portable application (w/out creating some kind of virtual environment). Nor is there any way to easily test a different version than currently installed, w/out overwriting all the presets, possibly changing the app config, etc.

Furthermore, there is a popular misconception that the `.zip` distribution provided in GH releases is "portable" and that it's "safe to try it out" alongside another installed version.  This is incorrect, or at least misleading, at best.

**Proposed Solution(s)**

1) Restore the `--datadir` CLI argument (and related functionality).
   * `+` Allows greater flexibility as to where all the configuration data is stored, including network shares, etc.
   * `+` Brings Studio (back) inline with _Slic3r_ and derivatives (_Prusa_, _Orca_, _Super_), all of which support `--datadir` (added back after forking, in the case of the latter two).
   * `-` Requires user action to ensure the desired folder is used (invoking from CLI, using a specific shortcut, etc).
2) Add an automatic search at program start for the presence of a directory, with a specific "magic" name, at the same folder level as the program is being run from.
   * `+` Addresses the last point above, making possible a fully portable "installation" just by adding a folder to store the configs (which could also be a symlink). It would be easier and less error-prone to start, eg. just by executing via file browser from an external disk/network share/etc, or even from a file type association.
   * `-` Adds a small delay/overhead to program startup due to check for directory (potentially twice in the case of an AppImage being used).
   * `-` By itself (w/out `1` above) this would still provide some flexibility with using different versions in parallel, for example, but not as much, possibly _requiring_ symlinks in order to use shared or remote storage locations.
   * `+` This feature parallels the one available now in _SuperSlicer_ and _OrcaSlicer_ (_Prusa_ and _Slic3r_ don't implement this).

So the two features combined provide the most flexibility, though either could stand alone as well.

**Proposed Implementation**

I split these commits on purpose since they have separate functionalities (either could be implemented on their own). OTOH to implement both makes sense to build the 2nd on top of the first, and the functionality is related, so I think it makes sense to present/discuss together.

I can squash, split to different PRs, edit... whatever is preferred.

1. Adding `--datadir` back, essentially, is very straight-forward and doesn't affect any existing functionality. From my tests, when started with `--datadir`, it never tries to write or read config data from anywhere else.  There's a minor issue with where the logs are written (see below) but otherwise this seems like a simple solution.
2. For the local config folder feature:
   * I chose "configuration" for the magic folder name. This matches _SuperSlicer_, but not _OrcaSlicer_, which uses "data_dir".
   * This name could be configured with a macro at build time.
   * The whole search for a local configuration folder step can be disabled at build time by defining the macro as an empty string.
      * I'm not sure the last two are entirely necessary, but mostly I dislike having "magic strings" hidden deep in code.
      * I'm also not sure if that's the "best" place to check for/define the macro, but nothing else jumped out, and that was the closest place to where it needs to be used.
   * On Mac the config folder is expected to be at the same folder level as the .app bundle.
   * On Linux it is expected to be at the same level as the "resources" folder (above "bin"), _or_ at the same folder level as the AppImage, if one is being used (it can also be stored _inside_ the AppImage archive, alongside resources, which is checked first).
   * On Windows it is expected to be in the rood folder where the binary and "resources", etc, are located.

**Issues & Related**

1) When used with a custom configuration folder, right now the main log file ends up in the current working directory.  This is because the act of switching CWD to the log folder is done in [GUI_App::init_app_config()](https://github.com/bambulab/BambuStudio/blob/c7bdea548a0519715d26c2a6c2a4dfc8365ea7ad/src/slic3r/GUI/GUI_App.cpp#L2505) _only_ if the current `data_dir().empty()` and it uses the default storage location.  (The networking log ends up correctly in `log` subfolder of config directory.)
   * The relevant block of code in `GUI_App::init_app_config()` is directly related to issue and code in PR #9477.  @wtji-bbl mentioned in the comments that a fix was already implemented.  I haven't seen that commit (sorry if I missed it), but I didn't want to touch any of that if you've already made changes. It's a relatively minor issue, and I'm happy to work on it if it still exists after whatever change is pending for it now. (Shouldn't the log path be absolute anyway, not related to CWD?)
2) When using either custom config location option, there's no validation being done that the destination is readable or writable (at least at startup).  This may result in crash or other UB.  _However_ I think this check should be done in `init_app_config()`, for whatever `data_dir()` currently is, _after_ it may have created the default storage location. Because really a folder could become un-read/writable at any time (for example app is running with different permissions).
   * Again I didn't want to step on whatever changes are being done in relation to the setting not getting saved, so I didn't implement it here, but happy to add later if needed.

**"Prior art"**
https://github.com/OrcaSlicer/OrcaSlicer/commit/771ef9fd80de4fe2f43e2e1ba2ff5268a84d0e41
https://github.com/OrcaSlicer/OrcaSlicer/commit/6ca1741e83ce8ea7e9338fa1f7a401acd07c2b8d & https://github.com/OrcaSlicer/OrcaSlicer/commit/fb032cd85f2dc7ffb1b32d478192a63ef53688cf
https://github.com/supermerill/SuperSlicer/commit/e4233d4e7dc075919323c72fd8b284ac3ac0ea57 & https://github.com/supermerill/SuperSlicer/pull/2160

---

Please let me know if any question, concerns, issues, etc.  And thank you for your time!
-Max

---
Closes https://github.com/bambulab/BambuStudio/issues/1651

---
Related: #9858  so Windows users can interact with the CLI better.

### DIFF
--- a/src/BambuStudio.cpp
+++ b/src/BambuStudio.cpp
@@ -104,6 +104,15 @@ using namespace Slic3r;
 
 #define MAX_CLONEABLE_SIZE 512
 
+// Folder name for portable version configuration storage location.
+// If a directory with this name is found at startup next to the program's executable,
+// it will be used for all settings/profiles storage instead of the usual OS-dependent location.
+// This check can also be overridden using the --datadir CLI argument.
+// The folder check can be disabled entirely at build time by setting this macro to an empty _string_ value ("").
+#ifndef PORTABLE_DATA_DIR_NAME
+    #define PORTABLE_DATA_DIR_NAME  "configuration"
+#endif
+
 std::map<int, std::string> cli_errors = {
     {CLI_SUCCESS, "Success."},
     {CLI_ENVIRONMENT_ERROR, "Failed setting up server environment."},
@@ -8184,29 +8193,37 @@ bool CLI::setup(int argc, char **argv)
     }
 #endif
 
-    // See Invoking prusa-slicer from $PATH environment variable crashes #5542
-    // boost::filesystem::path path_to_binary = boost::filesystem::system_complete(argv[0]);
-    boost::filesystem::path path_to_binary = boost::dll::program_location();
+    // Fully resolved directory path of program's runtime base "installation" folder, which may be "portable" version run from any arbitrary location.
+    boost::filesystem::path install_path;
+    // Starting with the current location of the binary being executed.
+    try {
+        install_path = boost::filesystem::canonical(boost::dll::program_location()).parent_path();
+    } catch (std::exception &e) {
+        // boost log not initialized yet
+        boost::nowide::cerr << "Could not determine canonical path to application directory!" << std::endl << e.what() << std::endl << std::endl;
+        return false;
+    }
 
-    // Path from the Slic3r binary to its resources.
+    // Path from the program binary to its resources.
 #ifdef __APPLE__
-    // The application is packed in the .dmg archive as 'Slic3r.app/Contents/MacOS/Slic3r'
-    // The resources are packed to 'Slic3r.app/Contents/Resources'
-    boost::filesystem::path path_resources = boost::filesystem::canonical(path_to_binary).parent_path().parent_path() / "Resources";
-#elif defined _WIN32
+    // The application is packed in the .dmg archive as 'BambuStudio.app/Contents/MacOS/BambuStudio'
+    // The resources are packed to 'BambuStudio.app/Contents/Resources'
+    const boost::filesystem::path path_resources = install_path.parent_path() / "Resources";
+    // For file system access outside the bundle, to check for datadir folder at that level.
+    install_path = install_path.parent_path().parent_path().parent_path();
+#elif defined(_WIN32)
     // The application is packed in the .zip archive in the root,
-    // The resources are packed to 'resources'
-    // Path from Slic3r binary to resources:
-    boost::filesystem::path path_resources = path_to_binary.parent_path() / "resources";
-#elif defined SLIC3R_FHS
+    // The resources are packed to 'resources' in the root.
+    const boost::filesystem::path path_resources = install_path / "resources";
+#elif defined(SLIC3R_FHS)
     // The application is packaged according to the Linux Filesystem Hierarchy Standard
     // Resources are set to the 'Architecture-independent (shared) data', typically /usr/share or /usr/local/share
-    boost::filesystem::path path_resources = SLIC3R_FHS_RESOURCES;
+    const boost::filesystem::path path_resources = SLIC3R_FHS_RESOURCES;
 #else
-    // The application is packed in the .tar.bz archive (or in AppImage) as 'bin/slic3r',
-    // The resources are packed to 'resources'
-    // Path from Slic3r binary to resources:
-    boost::filesystem::path path_resources = boost::filesystem::canonical(path_to_binary).parent_path().parent_path() / "resources";
+    // The application is packed in the .tar.bz archive (or in AppImage) as 'bin/bambu-studio',
+    // The resources are packed to 'resources' in the root.
+    install_path = install_path.parent_path();
+    const boost::filesystem::path path_resources = install_path / "resources";
 #endif
 
     set_resources_dir(path_resources.string());
@@ -8252,6 +8269,24 @@ bool CLI::setup(int argc, char **argv)
         // Don't validate existence or create it right now, should be done in GUI_App::init_app_config(), same as for default data dir.
         // We do want to store the full path in native format because that's how all other sources of the global data_dir are stored.
         set_data_dir(boost::filesystem::absolute(m_config.opt_string("datadir")).make_preferred().string());
+    }
+    // Check for special configuration folder at the same level as the installation path.
+    else if (std::strlen(PORTABLE_DATA_DIR_NAME) > 0) {
+        boost::filesystem::path local_data_dir_path = install_path / PORTABLE_DATA_DIR_NAME;
+        if (boost::filesystem::exists(local_data_dir_path)) {
+            set_data_dir(local_data_dir_path.make_preferred().string());
+        }
+#if defined(__linux__) || defined(__LINUX__)
+        // If running from an AppImage, the original package location is stored in APPIMAGE env. var.
+        // The user may have a loclal config folder at the same level as the image file, not inside the image itself.
+        else if (const char *appimage_env = std::getenv("APPIMAGE")) {
+            boost::system::error_code ec;
+            local_data_dir_path = boost::filesystem::canonical(boost::filesystem::path(appimage_env).parent_path() / PORTABLE_DATA_DIR_NAME, ec);
+            // Ignore errors, no local config folder or couldn't resolve APPIMAGE at all for some reason.
+            if (!ec)
+                set_data_dir(local_data_dir_path.string());
+        }
+#endif  // __linux__
     }
 
     return true;
--- a/src/BambuStudio.cpp
+++ b/src/BambuStudio.cpp
@@ -8239,14 +8239,19 @@ bool CLI::setup(int argc, char **argv)
         for (const t_optiondef_map::value_type &optdef : *options)
             m_config.option(optdef.first, true);
 
-    //set_data_dir(m_config.opt_string("datadir"));
-
     //FIXME Validating at this stage most likely does not make sense, as the config is not fully initialized yet.
     if (!validity.empty()) {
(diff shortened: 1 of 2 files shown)


raulp:
i wanted this so bad while i was testing different builds. I ended up doing the same thing, but never committing that part ;) Would love to see this merged!

mpaperno:
With the proliferation of beta versions and many other updates, it would be extremely useful to be able to run multiple versions side-by-side, literally at the same time in some cases, and w/out overwriting user configs each time.

If the proposed solution doesn't fit, I'm open to other suggestions as well.

@MackBambu  @tonghao-bbl  @lanewei120

raulp:
I still think this is the easiest fix for our problem ;) Hope it will be merged soon.

mpaperno:
From a developer's POV this feature is basically essential, IMHO.  How would one compare runtime behavior otherwise?

Currently we could have 3 separate config storage locations based on values of `BBL_RELEASE_TO_PUBLIC` and `BBL_INTERNAL_TESTING` build-time definitions. But this is still very limiting (and build-time only), plus it produces different builds. (Never mind that those defs aren't documented in the build instructions at all, and by default a new dev's existing configs will be overwritten with whatever version they just built.)

-Max

tonghao-bbl:
Thanks @mpaperno! I'll try to test this within the next couple of days when I have some time. I really appreciate all the PRs you've been submitting.

Please bear with us though — we have a heavy workload, and we want to make sure every PR is thoroughly reviewed and tested to maintain the quality of the project.

src/BambuStudio.cpp
```diff
@@ -7890,6 +7905,31 @@ bool CLI::setup(int argc, char **argv)
...
     }
 
+    // Set custom configuration storage location if invoked with --datadir argument.
+    if (m_config.has("datadir") && !m_config.opt_string("datadir").empty()) {
+        // Don't validate existence or create it right now, should be done in GUI_App::init_app_config(), same as for default data dir.
+        // We do want to store the full path in native format because that's how all other sources of the global data_dir are stored.
+        set_data_dir(boost::filesystem::absolute(m_config.opt_string("datadir")).make_preferred().string());
+    }
+    // Check for special configuration folder at the same level as the installation path.
+    else if (std::strlen(PORTABLE_DATA_DIR_NAME) > 0) {
```
tonghao-bbl, line 8274: Hi @mpaperno  is `PORTABLE_DATA_DIR_NAME` intended as the default value of `datadir`?  Is this a bit risk for the user? in case of a user has the `configureation` dir with this comamnd, it may be used as the data dir silently

mpaperno: Hi @tonghao-bbl, thanks for looking. 

Yes, some default folder name must be picked at build time.  As mentioned in the PR notes, the macro is meant to make this choice more obvious/easier to change.  And yes, a folder named "configuration" next to the executable would be used every time.  I'm not sure why anyone would have such a thing already, but it's not impossible.  As mentioned, this name matches _SuperSlicer_ but _Orca_ uses "data_dir" instead (perhaps a user is less likely to have a folder with that name than "configuration", IDK).

I'm not particular about the folder name. We could make it some random string of characters for all I care.  The basic functionality of specifying config location is badly needed.

src/BambuStudio.cpp
```diff
@@ -7890,6 +7905,31 @@ bool CLI::setup(int argc, char **argv)
...
+        if (boost::filesystem::exists(local_data_dir_path)) {
+            set_data_dir(local_data_dir_path.make_preferred().string());
+        }
+#if defined(__linux__) || defined(__LINUX__)
+        // If running from an AppImage, the original package location is stored in APPIMAGE env. var.
+        // The user may have a loclal config folder at the same level as the image file, not inside the image itself.
+        else if (const char *appimage_env = std::getenv("APPIMAGE")) {
+            boost::system::error_code ec;
+            local_data_dir_path = boost::filesystem::canonical(boost::filesystem::path(appimage_env) / PORTABLE_DATA_DIR_NAME, ec);
+            // Ignore errors, no local config folder or couldn't resolve APPIMAGE at all for some reason.
```
tonghao-bbl, line 8285: `$APPIMAGE` is the absolute path of the AppImage file, so should we get the parent before create the data dir?
check https://docs.appimage.org/packaging-guide/environment-variables.html for details.

Also, is error handling needed when creating the data dir here, like pass an error code?

mpaperno: What do you mean by the parent?  The expectation is that a local configuration folder would be on the same directory level as the AppImage being executed.  That's what `filesystem::path(std::getenv("APPIMAGE"))` returns, doesn't it?

```
/
⊢  BambuStudio.AppImage
⊢  configuration
```

We're not creating any directories in this code block. If the local configuration folder doesn't exist the code continues as normal, w/out a local config folder.

tonghao-bbl: I have't test it, but the document says `boost::filesystem::canonical(boost::filesystem::path(appimage_env) / PORTABLE_DATA_DIR_NAME` would be `/BambuStudio.AppImage/configureation` not `/configuration` as expected, right?

tonghao-bbl: `/BambuStudio.AppImage/configureation` in an invlaid path, will the boost::filesystem throw an exception?

tonghao-bbl: seems `/` will not throw exceptions except `bad_alloc`, so we are good here

mpaperno: > would be /BambuStudio.AppImage/configuration

Ah, I think you're right!  I must have misread the APPIMAGE documentation.  Strangely it seemed to work in testing, but maybe I missed something.  Will have to revisit and fix.

Thanks.

tonghao-bbl: Hi @mpaperno Do you have any time to fix this small issue, maybe we can ship this in the latest beta version

mpaperno: Done!  Also rebased onto current `master`.  A merge on this would be superb!

tonghao-bbl:
LGTM @lanewei120

lanewei120:
thanks @mpaperno

mpaperno:
Thanks for the reviews and merge!

@lanewei120 & @tonghao-bbl , may I help document this new feature?
If so, what would be the best way to go about it?  I'm not sure which Wiki it should go in (here or main BL docs site), or maybe both.  Should I just write something up with simple/minimal Markdown formatting and attach it here?

-Max

tonghao-bbl:
@mpaperno Based on the current situation I think is help output of the cli is good to go, Thank you!

mpaperno:
@tonghao-bbl  Odd...  current situation?  The "configuration" folder isn't documented anywhere (besides here), and you yourself pointed out the potential conflict in a question (if someone happens to have a "configuration" folder for some other reason).  And I don't know why someone would check the CLI help w/out knowing to look in the first place.

Well, offer still stands and there's always the _Orca_ docs to copy from I suppose.

Cheers,
-Max

tonghao-bbl:
@mpaperno Sorry I forgot the conflict issue I pointed out, and yes we should document it.
Can you edit https://github.com/bambulab/BambuStudio/wiki/Command-Line-Usage? It maybe a little outdate, it would be a great help if you can review and improve it, Thank you!

mpaperno:
Hi @tonghao-bbl ,

I made changes to my forked version of the Wiki: https://github.com/mpaperno/BambuStudio/wiki

Mostly I made a new page specifically regarding the configuration location and how to set it.  
https://github.com/mpaperno/BambuStudio/wiki/Program-Configuration-Data-Location

I did also update the CLI details page.

https://github.com/mpaperno/BambuStudio/wiki/_history

There is no way to submit a PR for a wiki update.  Commits need to be pulled manually from contributor repo into the source repo by someone with sufficient permissions.

I hope this helps!

-Max

PR #9995, opened 2026-03-16
FIX: Properly initialize `BaseRenderer::m_config` with `nullptr`.

### LINKED ISSUE #9991: G-code file exported from any plate cannot be reopened — Bambu Studio silently closes mid-load
### Bambu Studio Version

2.5.0.66

Bambu Lab Official website

### OS version

Windows 10 and 11 (cheched on 2 devices)

_No response_

### Printer

P1S

### Printer Firmware Version

last

### How to reproduce

1. Open any project in Bambu Studio 2.5.0.66.
2. Slice a model on any plate.
3. Export/save the plate as a .gcode file to disk.
4. Close Bambu Studio.
5. Double-click the saved .gcode file to open it. 

### Actual results

Bambu Studio launches and begins loading the file, but closes silently partway through the loading process. No error message or crash dialog is displayed. The issue is reproducible with any plate saved as .gcode — it is not specific to a single file or project.

### Expected results

Bambu Studio launches and successfully opens the .gcode file.

[Untitled.gcode.3mf.zip](https://github.com/user-attachments/files/26010424/Untitled.gcode.3mf.zip)

- [ ] Log file
- [x] Project file

mpaperno:
Fixes https://github.com/bambulab/BambuStudio/issues/9991

Since: https://github.com/bambulab/BambuStudio/commit/9705d20636718bf47257c184c824251fa56880d6

Since the pointer wasn't initialized, whether the crash happened or not was inconsistent. Required "always show shells" option to be enabled.

The 2nd commit here skips some unnecessary processing if there is no config or no shells.  Seems to be that most gcode-only previews wouldn't have shells, so I assume from the comment this has something to do with Helio.

In case it means anything, I noticed that when loading a gcode file (as in the linked Issue), the code path in the 2nd commit gets hit twice, once without a print config, and then the 2nd time with one.

Cheers,
-Max

### DIFF
--- a/src/slic3r/GUI/GCodeRenderer/BaseRenderer.cpp
+++ b/src/slic3r/GUI/GCodeRenderer/BaseRenderer.cpp
@@ -597,9 +597,10 @@ namespace Slic3r
                             { min.x(), max.y() } };
                     }*/
                 }
-                else if (&wxGetApp() && wxGetApp().app_config->get_bool("show_shells_in_preview")) { // BBS: load shell at helio_gcode
+                else if (m_config && &wxGetApp() && wxGetApp().app_config->get_bool("show_shells_in_preview")) { // BBS: load shell at helio_gcode
                     load_shells(print, initialized,true);
-                    update_shells_color_by_extruder(m_config);
+                    if (!m_shells.volumes.empty())
+                        update_shells_color_by_extruder(m_config);
                 }
                 m_print_statistics = gcode_result.print_statistics;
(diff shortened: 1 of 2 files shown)


Haidiye00:
Hello! Thank you very much for the fix. Our internal team also identified this issue and has already implemented a solution. Thanks again for your efforts!

mpaperno:
Great, please push your fixes to the public repo here and make a note on the original issue so others can know it has been fixed.

mpaperno:
This still isn't fixed in latest beta release or in the code base here.  Uninitialized pointer being checked for null value.

PR #10033, opened 2026-03-19
Vendor -> Hersteller / Serial -> Serie

albbino:
Changed german translation for Vendor and Serial. 
Also some other fixes in Translation


FiredEgg-bambu:
Hi @albbino 
Thank you for the correction. We have already fixed these translations on our internal translation platform, so this pull request will not be merged.

Thank you for your contribution.

Issue #1416, opened 2023-03-07
Focus zoom on selected object


noossyy:
**Is your feature request related to a problem? Please describe.**

Without a focus zoom on an object, your can’t properly rotate around and object after having zoomed on it. 

**Describe the solution you'd like**
Replicate the same behavior as the Z hotkey in superslicer or F hotkey in Autodesk Maya when object is selected. The camera will zoom close to it and center focus on it so you can properly rotate around it. 

**Describe alternatives you've considered**
Having a hotkey to do that like in most 3D animation software 

**Additional context**

BowieInc:
Yes! How can I add my support for this shortcut addition? It is much needed.

kertgartner:
YES PLEASE! Would love to see this added!

Thelvaen:
I'd love to see this feature as well.

CaidenKehrer:
I would also really appreciate this feature. It is nearly impossible to get angles right to view some objects .

travrob1:
Yes please!!! That woulds same me so much time.  Thanks

ctkjedi:
Another vote for this! Zoom to mouse is fine, but the moment you rotate your view...

mpaperno:
Hi all,

Fun fact: Holding `CTRL` (`CMD` on Mac) while dragging with the left mouse button **will rotate around the current <strike>cursor position</strike> view's center point**!  Vs. the center of the current plate.

I just discovered this by looking at the code.  If it's documented somewhere, I can't find it!

It's not quite ideal since the rotation is still dependent on left mouse button, which means you can't click on the actual object you want to rotate around. Which is a general annoyance anyway.  But at least it's something.  Rotating round the center of the plate drives me nuts.

I'd now like the option of having this kind of rotation by default, with the CTRL modifier doing the opposite (rotating around plate center).

-Max

PR #10035, opened 2026-03-19
ENH: Add "Fit camera to scene or selection" function from 3D navigator as a global shortcut.

### LINKED ISSUE #7337: Camera Controls - Focuses the view on the currently selected object

Its not a problem. But it's essential for a smooth workflow when you want to ispect an object.

Select an object at Prepare and focus the view on the currently selected object when `Num.` is pressed.

Image

### LINKED ISSUE #10017: shortcut for fit camera
I use the button "Fit camera to scene or selected object" really often and work usually with the keyboard.

Add a shortcut for the button please.
* home
* alt + home
* strg + home
* ....

mpaperno:
As requested in #10017 , and by myself.  :)

This adds the shortcut event handler directly to `MainFrame` so it can be triggered even if focus is in the sidebar/object tree (unlike the other view shortcuts).  This is handy because it may be the most convenient, or only, way to select a part of an object to focus on. For example if a part/modifier is completely inside another object.

<strike>I chose <kbd>CTRL + HOME</kbd> as the shortcut, as suggested in the issue. Seemed logical...  Just <kbd>HOME</kbd> could also be used, as it doesn't really interfere with anything else right now (in the object tree it will jump to the top of the list, but so will <kbd>PAGE UP</kbd>).</strike>  
Now using `Z` as the shortcut, which matches the other Slic3r forks.  
Sill open to other suggestions/preferences of course.

Thanks,
-Max

--
Closes #10017 
Closes #7337
Related: 
#1416

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -707,6 +707,12 @@ DPIFrame(NULL, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, BORDERLESS_FRAME_
             if (m_plater) { m_plater->add_file(); }
             return;
         }
+
+        if (!evt.HasAnyModifiers() && evt.GetKeyCode() == 'Z') {
+            view_zoom_to_fit();
+            return;
+        }
+
         evt.Skip();
     });
 
@@ -2623,6 +2629,8 @@ static void add_common_view_menu_items(wxMenu* view_menu, MainFrame* mainFrame, 
     append_menu_item(view_menu, wxID_ANY, _CTX(L_CONTEXT("Isometric", "Camera"), "Camera") + " 3", _L("Isometric View") + " 3",
         [mainFrame](wxCommandEvent &) { mainFrame->select_view("iso_3"); }, "", nullptr, [can_change_view]() { return can_change_view(); }, mainFrame);
 #endif
+    append_menu_item(view_menu, wxID_ANY, _L("Fit Camera") + "\t" "Z", _L("Fit camera to scene or selected object."), 
+        [mainFrame](wxCommandEvent &) { mainFrame->view_zoom_to_fit(); }, "", nullptr, [can_change_view]() { return can_change_view(); }, mainFrame);
 }
 
 void MainFrame::init_menubar_as_editor()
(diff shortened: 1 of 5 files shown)


mpaperno:
Come to think of it, simply `Z` is still available as a shortcut.  This would keep it the same as the other Slic3r forks. And easier to press.  But doesn't follow the `CTRL +` convention of the other view shortcuts.

Thoughts?  

Thanks,
-Max

@DWay17

DWay17:
> Come to think of it, simply `Z` is still available as a shortcut.  This would keep it the same as the other Slic3r forks. And easier to press.  But doesn't follow the `CTRL +` convention of the other view shortcuts.
> 
> Thoughts?  
> 
> Thanks,
> -Max
> 
> @DWay17
> 

Only z is of course also good. And making it uniform is always good. 

LG Thomas

mpaperno:
Switched shortcut key to `Z`.  

Much nicer, IMHO.  The "common" shortcut for this in other apps are `F` or `CTRL + 0`, both of which are unfortunately already taken.

guanyun-gudujian:
good job

src/slic3r/GUI/MainFrame.cpp
```diff
@@ -3922,6 +3930,12 @@ void MainFrame::select_view(const std::string& direction)
          m_plater->select_view(direction);
 }
 
+void MainFrame::view_zoom_to_fit() const
+{
+    if (m_plater)
+        m_plater->view_zoom_to_fit();
```
mpaperno: @guanyun-gudujian  I'm not sure what happened to your comment... but I liked it! :-)  And thanks for looking!

[If your comment no longer applies then never mind this one.]

> better to use: m_plater->get_view3D_canvas3D()

Perhaps `Plater::canvas3D()`?  (That will always return the current `GLCanvas3D`, which could be the "prepare" 3D view, slicer preview, or assembly.)

And then have `MainFrame::view_zoom_to_fit()` call `GLCanvas3D::zoom_to_fit()` directly, & remove `Plater::view_zoom_to_fit()` entirely?

Makes sense to me, just checking that's what you meant.

guanyun-gudujian: yes that's correct

mpaperno: Sounds good, will update!

mpaperno:
Updated to simplify code based on review comments.

guanyun-gudujian (approved):

PR #9187, opened 2025-12-24
#5597 - Add support for 12 hour clock format

csm10495:
This adds a checkbox to support using 12 hour time format instead of the default 24 hour time format:

Pref:
image

On the printing progress page: 
image

On the recent projects page:

image

I can definitely refine it a bit as needed. Though would like to know if I'm on the right direction or not. Thanks!

Here's a video of the setting in action:

https://github.com/user-attachments/assets/4fd4a14f-4e62-4dc1-9e36-789bcfd876a4

Issue #9698, opened 2026-02-07
can't print via cloud, curl report self -signed ?


conarsad:
### Bambu Studio Version

2.5.0.66

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11

### Printer

Bambu Lab A1

### Printer Firmware Version

01.07.02.00

### How to reproduce

print via cloud(non-LAN mode)
[http_code]: 60
[request_id]: a14e7cce-d15a-49ad-94b2-42279352661a
[reason]: curl:SS: peer certificate pr SSH remote key was not OK: SSL certificate problem: unable to get local issuer certificate [Error 60]
send G-code to sd card OK!
Print table NOT OK!

### Actual results

[http_code]: 60
[request_id]: a14e7cce-d15a-49ad-94b2-42279352661a
[reason]: curl:SS: peer certificate pr SSH remote key was not OK: SSL certificate problem: unable to get local issuer certificate [Error 60]

<img width="634" height="262" alt="Image" src="https://github.com/user-attachments/assets/9769bf07-d269-4502-8fe5-935b2ddb28a5" />

### Expected results

successfully print

### Project file & Debug log uploads

[debug_network_Sat_Feb_07_10_01_50.log.enc.txt](https://github.com/user-attachments/files/25143211/debug_network_Sat_Feb_07_10_01_50.log.enc.txt)

### Checklist of files to include

- [x] Log file
- [ ] Project file

HoussemEJ:
Same exact issue for me, same version of Bambu studio but my printer is the p2s. Back in December/ January everything was fine this is definitely recent in my case.

Location: Germany

idk if this helps too but these tests are mostly hit or miss for whatever reason never consistent. 

Would really appreciate some support I have to click print between 5 and 15 times to get a successful print run. I prefer not to use Lan only mode for this because I like the cloud convenience. 

Thanks in advance

<img width="529" height="596" alt="Image" src="https://github.com/user-attachments/assets/537ec1c9-9321-4b69-a90b-8b2c8f3786ad" />

<img width="507" height="387" alt="Image" src="https://github.com/user-attachments/assets/e85f0e13-e2e8-41b5-8eec-5774bc790f3a" />

HoussemEJ:
@harry-bbl any info on this? it's quite frustrating as it can take many attempts to send a print this is really killing the experience to use the printer. For info I disabled my firewall entirely and didn't help. The handy app works well all the time so this is exclusive to the bambu studio.

[...]

PR #10055, opened 2026-03-22
ENH: Re-add 12-hour time format support with thread-safety fix

### LINKED ISSUE #5597: Option to display finish time in 12hr format
I'd love an option to display the Finish Time (Under Device Tab -> Printing card) in 12hr format

csm10495:
#5597 and #9187 again.
Re-applies the 12-hour time format feature from commit 0a89a33b which was reverted in 331496e4 due to crashes (likely) caused by thread-unsafe std::localtime() calls.

Fix: Added safe_localtime() wrapper in Utils.hpp that uses localtime_s (Win) or localtime_r (POSIX) to return std::tm by value, eliminating the shared static buffer race condition that caused crashes on multi-threaded access.

All call sites now use safe_localtime() instead of std::localtime():
- get_bbl_finish_time_dhm() in Utils.hpp
- MediaPlayCtrl::SetStatus()
- MainFrame::get_recent_projects()
- CreatePresetsDialog::get_curr_time()

I'm sort of guessing a bit on the crashes, though if we have more logs or info i may be able to confirm/check.

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -487,6 +487,10 @@ void AppConfig::set_defaults()
         set_bool("prompt_for_brittle_filaments", true);
     }
 
+    if (get("use_12h_time_format").empty()) {
+        set_bool("use_12h_time_format", false);
+    }
+
     // Remove legacy window positions/sizes
     erase("app", "main_frame_maximized");
     erase("app", "main_frame_pos");
--- a/src/libslic3r/Utils.hpp
+++ b/src/libslic3r/Utils.hpp
@@ -801,19 +801,55 @@ inline std::string get_bbl_monitor_time_dhm(float time_in_secs)
     return buffer;
 }
 
-inline std::string get_bbl_finish_time_dhm(float time_in_secs)
+// Thread-safe wrapper around localtime. Returns a std::tm by value.
+inline std::tm safe_localtime(const time_t* timep)
+{
+    std::tm result{};
+#ifdef _WIN32
+    localtime_s(&result, timep);
+#else
+    localtime_r(timep, &result);
+#endif
+    return result;
+}
+
+// Centralized function to format time from a std::tm structure
+// This is the single source of truth for time formatting throughout the application
+inline std::string format_time_hm(const std::tm* tm, bool use_12h_format = false)
+{
+    std::ostringstream formattedTime;
+
+    if (use_12h_format) {
+        int hour = tm->tm_hour;
+        std::string suffix = (hour >= 12) ? "PM" : "AM";
+        int display_hour = hour % 12;
+        if (display_hour == 0) display_hour = 12; // Midnight = 12AM, Noon = 12PM
+        formattedTime << std::setw(2) << std::setfill('0') << display_hour << ":"
+                      << std::setw(2) << std::setfill('0') << tm->tm_min << suffix;
+    } else {
+        // 24-hour format
+        formattedTime << std::setw(2) << std::setfill('0') << tm->tm_hour << ":"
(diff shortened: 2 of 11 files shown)


mpaperno:
Hi, thanks for your work.  

I'm wondering, is there anything wrong with using `wxDateTime` for retrieval and formatting?  `::Now()` safely returns current time as an object, and the class has simple to use formatting functions.  Also ones that respect the user's locale.

https://docs.wxwidgets.org/3.3/classwx_date_time.html#a6e6c37a0414bb4831e2cc03b37f498a2

https://docs.wxwidgets.org/3.3/classwx_date_time.html#a6a60f071d74712eba58153113f5af23a

Just a thought.

-Max

csm10495:
I don't know if we want it to just respect locale as that would be a new change and a mismatch from the app. At least by defaulting to 24hr it's the same by default.

I'm open to whatever we think in terms of time fetching mechanism. I just like the idea of trying to be universal-ish if that makes sense.

mpaperno:
Well, it can still default to 24h, which makes sense since that has been the default so far.  Just saying all the formatting and such would be a lot simpler (just a change to the format template basically).  And/or modify based on user preference down the line, potentially (eg. "show dates in local format" option).  And I think I know what you mean about "universal" (portable) but really the whole GUI is in so deep with `wxW` that date formatting would be the least of the issues if one was considering porting it.  IMHO   :)

-Max

csm10495:
I see so the idea is 'default to 24h' then checkbox to 'use local locale' essentially. The problem is wxDateTime doesn't work that way (at least for 12h time). With wxWidgets 3.3.0:

```
#include <wx/datetime.h>
#include <iostream>

int main()
{
    wxDateTime now = wxDateTime::Now();
    std::cout << now.FormatTime().utf8_string() << std::endl;
    return 0;
}

```

I'm getting: 
```
20:52:25
```
Though it should be something like 08:52:25 PM. My windows clock says 8:52 PM as well. I think the way I have it solves the 12hr vs 24hr problem better by being more direct.

mpaperno:
Well, not trying to push anything, but I was suggesting using `wxDateTime::Format()` more directly, with all the available formatting options, depending on if the user chooses 12 or 24 time format.

https://cplusplus.com/reference/ctime/strftime/

```cpp
wxDateTime::Now().Format(use_12h_format ? "%I:%M%p", "%R");
```

Also that whole bit with calculating remaining time could be shortened to one line if wxDateTime was used... but that's another story.  :)

https://docs.wxwidgets.org/3.3/classwx_date_time.html#af854fdd5db4ed3c6bd3e39960241aa9f

Anyway, not attached, just suggesting an alternative.

-Max

csm10495:
I see. Yeah I'm sticking with this for now, though if the nit turns into a more blocking stop, then I can give it a try. 

It gets kind of confusing in the code since sometimes  its a time in seconds, sometimes its a std::tm, sometimes its a wxDateTime. I think I've even seen a string passed around and parsed. If its all the same, I'd leave as is since at least it's (so far) a working combo for me. 

Interestingly I'm having some new trouble testing this.. going to convert to draft while i play with it again. (probably messed up my env by mistake)

csm10495:
Ok working about as well as I can have it. I'm hitting: https://github.com/bambulab/BambuStudio/issues/9698 when trying to print directly which makes it tough to test. So I tested by starting a job in the release version then checking status in my built.
(Though honestly.. that's annoying and I should be able to opt into online printing via my own build).

left is release, right is my build.

image

This is ready for review again.

lanewei120:
Thanks very much for the fix

We have already included a similar fix in the latest version
Could you check the codes and confirm whether need to do more?

csm10495:
LGTM. The latest released has it and it works. Thanks! Closing out.

PR #10069, opened 2026-03-24
Fix multicolor top-surface artifacts when interface shells and one-wall top surfaces are enabled

### LINKED ISSUE #7279: Issue between Interface Shells and Only one wall on top surfaces
### Bambu Studio Version

2.1.0.59

Bambu Lab Official website

### OS version

Mas OS Sequoia

_No response_

### Printer

A1

### How to reproduce

1. have a model with two parts of different material or anything to test interface shells.
2. have Interface Shells and Only one wall on top surfaces turned on.
3. slice

### Actual results

the top surface generated between the two parts, has artifacts

### Expected results

Interface Shells works ok, and there is a full solid clean surface, and Only one wall on top surfaces is working on and we have just one wall on that surface

I've found a workaround using Height range modifier, but is really painful to edit all my models and generate hundreds of modifiers all day to accomplish this.

In the file I will upload you can see that if I manually set just one wall to the interface layer there is no artifacts, so I have to choose "Topmost surface" in the Only on a wall setting, and then add a Height range modifier for just one layer and set manually in that modifier to have just one wall.

Using Height range modifier give us more new problems, like having to deal with new bottom and top layer generated and full bridg layers, so I have to create even more Height range modifiers with special "zero bottom layers" for example etc.

I hope this investigation helps.

[test one wall layer with interface shells.3mf.zip](https://github.com/user-attachments/files/20746707/test.one.wall.layer.with.interface.shells.3mf.zip)

- [ ] Log file
- [x] Project file

geomad:
Fixes #7279.

Problem
When Interface shells and Only one wall on top surfaces = Top surfaces are enabled together, Bambu Studio can leave artifacts on the layer below upper different-color/material geometry.

This happened in two related cases:
1. Multipart multicolor models
2. Single-part multicolor models

Root cause

Multipart multicolor case
Top-surface single-wall detection was checking against full upper-layer slices instead of same-region upper slices when interface_shells is enabled.
As a result, geometry from other color/material regions on the upper layer affected top-surface classification for the current region and produced artifacts.

Single-part multicolor case
In detect_surfaces_type(), the top surface on the layer below the colored text/shape could be created with enclosed holes matching the upper colored geometry.
Those holes later became internal regions and preserved the upper footprint on the visible surface below.
The fix fills only the enclosed top-surface holes that are actually covered by upper geometry.

### Repro from #7279 (multi-part object)
This screenshot uses 2 walls because increasing wall count partially hides the artifact, but does not resolve the underlying bug.

Without the fix:
image

With the fix:
image

### Additional repro (single-part object)
Without the fix:
image

With the fix:
image

### DIFF
--- a/src/libslic3r/LayerRegion.cpp
+++ b/src/libslic3r/LayerRegion.cpp
@@ -166,6 +166,10 @@ void LayerRegion::make_perimeters(const SurfaceCollection &slices, const Perimet
     if (this->layer()->upper_layer != NULL)
         g.upper_slices = &this->layer()->upper_layer->lslices;
 
+    int region_id = this->region().print_object_region_id();
+    if (this->layer()->upper_layer != NULL)
+        g.upper_slices_same_region = &this->layer()->upper_layer->get_region(region_id)->slices;
+
     g.layer_id              = (int)this->layer()->id();
     g.ext_perimeter_flow    = this->flow(frExternalPerimeter);
     g.overhang_flow         = this->bridging_flow(frPerimeter, object_config.thick_bridges);
--- a/src/libslic3r/PerimeterGenerator.cpp
+++ b/src/libslic3r/PerimeterGenerator.cpp
@@ -1127,7 +1127,11 @@ void PerimeterGenerator::process_classic()
                     last_box.offset(SCALED_EPSILON);
 
                     // BBS: get the Polygons upper the polygon this layer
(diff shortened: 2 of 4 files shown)


raulp:
This would be huge. I have soooo much pain when i have to do this with modifiers and other hacks. Please @lanewei120 prioritize this. @geomad can you fix the conflicts?

PR #10076, opened 2026-03-24
Fix multimaterial top surface artifacts

### LINKED ISSUE #2906: I wish a clean layer before text is printed (especially useful for keytags)
### Bambu Studio Version

1.7.7.89

Bambu Lab Official website

### OS version

Windows 10

CPU: Intel i7 7700K
RAM: 16GB DDR4 CL14
GPU: Nvidia RTX 2060

### Printer

Bambulabs P1S

### How to reproduce

Slice any keytag and scroll down the layers until reaching the base surface

Keytag in Prusa Slicer  (Which also has the automatic color-change function) :
Screenshot (283)

Completely even surface with straight lines:
Screenshot (285)

### Actual results

Result in BBL-Studio (Pre-sliced text on flat surface) :
Screenshot (287)

### Expected results

A flat surface like here:  

Screenshot (285)

As when there is no text on that layer, why should there be any trace of it? Does not make any sense.

[KayTag_Test.zip](https://github.com/bambulab/BambuStudio/files/13385247/KayTag_Test.zip)
[debug_network_Fri_Nov_17_00_28_14.zip](https://github.com/bambulab/BambuStudio/files/13385249/debug_network_Fri_Nov_17_00_28_14.zip)

- [X] Log file
- [X] Project file

geomad:
Fixes #2906.

This PR is stacked on top of #10069.
Please review it on top of that PR. After #10069 merges, this branch can be rebased onto `master`.

## Problem
Bambu Studio could still leave visible text/label footprints on the flat layer below in multi-material cases.

After the previous fix in #10069, the `Only one wall on top surfaces = Not apply` path was corrected, but the remaining multi-material artifact still needed follow-up changes.

## Root cause
There were still two remaining paths that could preserve the footprint:

1. In `prepare_infill()`, enclosed `stInternal` / `stInternalSolid` islands under upper geometry could survive on an otherwise top surface.
2. In the `Only one wall on top surfaces = Apply` path, the perimeter generator was recomputing top regions from raw upper-layer slices instead of using the already-classified current-layer `stTop` geometry.

## Fix
This PR adds the remaining follow-up changes:

- promote enclosed internal islands under upper geometry back to `stTop` in `prepare_infill()`
- merge promoted top islands back into the surrounding top geometry
- make the top-one-wall split use the current layer's classified `stTop` surfaces instead of recomputing top regions from raw `upper_slices`

### Repro from https://github.com/bambulab/BambuStudio/issues/7279

Without the fix:
image

With the fix:
image

### DIFF
--- a/src/libslic3r/PerimeterGenerator.cpp
+++ b/src/libslic3r/PerimeterGenerator.cpp
@@ -33,6 +33,23 @@ static double random_value() {
     return dist(gen);
 }
 
+static ExPolygons clipped_top_surfaces(const SurfaceCollection &slices, const BoundingBox &bbox, const double min_width_top_surface)
+{
+    ExPolygons top_surfaces;
+    top_surfaces.reserve(slices.surfaces.size());
+    for (const Surface &surface : slices.surfaces)
+        if (surface.surface_type == stTop)
+            top_surfaces.emplace_back(surface.expolygon);
+
+    if (top_surfaces.empty())
+        return {};
+
+    ExPolygons clipped = union_ex(ClipperUtils::clip_clipper_polygons_with_subject_bbox(top_surfaces, bbox));
+    if (min_width_top_surface > 0.)
+        clipped = opening_ex(clipped, float(min_width_top_surface));
+    return clipped;
+}
+
 // Hierarchy of perimeters.
 class PerimeterGeneratorLoop {
 public:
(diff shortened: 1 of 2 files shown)


geomad:
Closing this for now because the current patch fixes one repro but still introduces regressions in other top-surface cases.
I’m continuing work on the same branch and will reopen this PR once the one-wall top-surface logic is corrected.

PR #10082, opened 2026-03-25
Fix multi-material top-surface artifacts (#2906)

### LINKED ISSUE #2906: I wish a clean layer before text is printed (especially useful for keytags)

geomad:
Fixes #2906.

This PR is stacked on top of #10069.
Please review on top of that PR. After #10069 merges, this branch can be rebased onto `master`.

## Problem
Multi-material objects (e.g., keytag with raised text as a separate part in a different material) show the text outline as visible artifacts on the base layer below the text. Two separate issues contribute:

1. **Fill pattern boundary**: The text footprint area is classified as `stInternal` (rectilinear fill) while the surrounding exposed base is `stTop` (monotonic fill), creating a visible outline.
2. **Wall count difference**: `PerimeterGenerator` computes one-wall-top areas independently of surface classification. The text footprint creates holes in the top area, causing different wall counts at the text boundary.

## Fix

**PrintObject.cpp** — After `detect_surfaces_type`, reclassify `stInternal` surfaces as `stTop` on layers that have both types, when the internal surface is covered by the upper layer. This unifies the fill pattern across the entire visible top surface.

**PerimeterGenerator.cpp** — Fill holes in the one-wall-top area that are covered by the upper layer, for both classic and arachne paths. This unifies the wall count across the entire visible top surface.

**PerimeterGenerator.hpp / LayerRegion.cpp** — Pass a `fill_top_surface_holes` flag set to `!is_mm_painted()`. Both fixes are skipped for mm-painted (single-part multicolor) objects which already have correct surface classification and would otherwise regress.

### Repro from #2906 — multi-material keytag (text as separate part, different material)

Without the fix:
image

With the fix:
image

### DIFF
--- a/src/libslic3r/Layer.cpp
+++ b/src/libslic3r/Layer.cpp
@@ -13,6 +13,26 @@ static const int dist_scale_threshold = 1.2;
 
 namespace Slic3r {
 
+ExPolygons interface_shell_upper_slices_for_top_holes(const Layer &upper_layer, size_t region_id)
+{
+    const bool single_region_interface_shells = upper_layer.object()->num_printing_regions() == 1;
+
+    ExPolygons upper_interface_slices;
+    const LayerRegionPtrs &upper_regions = upper_layer.regions();
+    if (upper_regions.size() > 1) {
+        for (size_t upper_region_id = 0; upper_region_id < upper_regions.size(); ++upper_region_id) {
+            if (upper_region_id == region_id)
+                continue;
+            expolygons_append(upper_interface_slices, to_expolygons(upper_regions[upper_region_id]->slices.surfaces));
+        }
+    } else if (single_region_interface_shells)
+        upper_interface_slices = upper_layer.lslices;
+
+    if (! upper_interface_slices.empty())
+        upper_interface_slices = union_ex(upper_interface_slices);
+    return upper_interface_slices;
+}
+
 Layer::~Layer()
 {
     this->lower_layer = this->upper_layer = nullptr;
--- a/src/libslic3r/Layer.hpp
+++ b/src/libslic3r/Layer.hpp
@@ -285,6 +285,9 @@ private:
     LayerRegionPtrs     m_regions;
 };
 
+// Upper-layer footprint allowed to act as top surface at interface-shell transitions.
+ExPolygons interface_shell_upper_slices_for_top_holes(const Layer &upper_layer, size_t region_id);
+
 class SupportLayer : public Layer
 {
 public:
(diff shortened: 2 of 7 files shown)


EricDimitri:
Hi Hi!
IS there a plan of merging this to Prod @zhimin-zeng-bambulab @ArthurBambulab 

Thanks for your work!!!

bbl-xiaoshu:
Thank you for your hard work on this PR! @geomad  We pulled this branch and tested it locally. The good news is that it successfully eliminates the visible text outline artifacts on the top surface, resolving the core issue of #2906.

However, we noticed that in some cases, this fix inadvertently introduces new regressions affecting the internal structure generation. Specifically:

1. **Incorrect Surface Classification on lower layers:** Several layers directly beneath the actual top surface of the base (like the keytag) are now incorrectly classified as `stTop`. Normally, these should be `stInternalSolid`.
2. **Loss of Sparse Infill (Entire model becomes solid):** The sparse infill（`stInternal`） inside the base completely disappears. After slicing with this PR, the entire base model is filled solidly from bottom to top, which is unexpected behavior and would unnecessarily increase print time and material consumption.

It appears the logic in `PrintObject.cpp` that reclassifies `stInternal` to `stTop` might be too aggressive, applying to layers far below the actual interface.

Here are the comparison screenshots for reference:

**Before this PR:**
*   Layer below top surface (Correct sparse/solid infill generation): 
before1
before3


**After this PR:**
*   Layer below top surface (Incorrectly classified as Top Surface & 100% Solid): 
after1
after3



Could you please check the code, especially the relevant parts involving the `stInternal->stTop` conversion? Thanks again for your excellent contribution!

geomad:
Thanks for checking this and for the screenshots! @bbl-xiaoshu 

I pushed a follow-up commit that fixes the sparse infill regression. The previous approach was modifying the slice classification too broadly, so I moved the interface-shell adjustment to fill_surfaces instead of changing the underlying slices.

discover_vertical_shells() now picks up those interface top patches from fill_surfaces, which keeps the intended top-shell behavior without turning lower internal layers into stTop.

I also gated the multimaterial fix behind Interface shells, so it only applies when that setting is enabled.

PR #10083, opened 2026-03-25
Fix second CLI segfault during --slice in extruder variant config

niccolodevries:
## Summary

Fixes a **second CLI segfault** discovered after #9941. While #9941 fixed the NULL `m_plater` dereference in `PartPlate::set_shape()` (which crashed during model arrangement), this PR fixes a separate SIGSEGV that occurs later during `Print::apply()` when the slicer processes extruder variant configurations.

In CLI mode, config options like `filament_map`, `extruder_type`, and `nozzle_volume_type` may not be present in the merged config when loading individual JSON profile files. The code unconditionally dereferences these without NULL checks, causing a segfault at address `0x8` (NULL + struct member offset).

Additionally, option lookups in the switch/case blocks (e.g., `this->option<ConfigOptionInts>(key)`) can return `nullptr` for keys that exist in the config *definition* but not in the actual config values.

Zero impact on GUI mode where these options are always populated.

## Steps to reproduce

Requires #9941 to be applied first (otherwise the earlier segfault masks this one).

```bash
BambuStudio \
  --load-settings "machine/Bambu Lab X1 Carbon 0.4 nozzle.json;process/0.20mm Standard @BBL X1C.json" \
  --load-filaments "filament/Bambu PLA Basic @BBL X1C.json" \
  --slice 0 \
  --export-3mf output.3mf \
  model.stl
# Crashes with exit code 139 (SIGSEGV)
```

## After fix

```
$ BambuStudio \
  --load-settings "machine/Bambu Lab X1 Carbon 0.4 nozzle.json;process/0.20mm Standard @BBL X1C.json" \
  --load-filaments "filament/Bambu PLA Basic @BBL X1C.json" \
  --slice 0 \
  --export-3mf 3DBenchy_X1C.3mf \
  3DBenchy.stl

$ echo $?
0
```

## Changes

**`src/libslic3r/PrintConfig.cpp`:**
- Add NULL guard for `opt_extruder_type` / `opt_nozzle_volume_type` in both `update_values_to_printer_extruders()` and `update_values_to_printer_extruders_for_multiple_filaments()` — return early if missing
- Add NULL guard for `filament_map` option in `update_values_to_printer_extruders_for_multiple_filaments()`
- Add `if (!opt) continue;` for all option type cases in the switch blocks of both functions, preventing NULL dereference when a config key is defined but has no value

**`src/libslic3r/PrintApply.cpp`:**
- Add NULL guard for `filament_map` in `Print::apply()` with fallback to `{1}` (matching the config definition default)
- Add NULL guard for `opt_extruder_type` / `opt_nozzle_volume_type` in the `filament_map_2` loop

### DIFF
--- a/src/libslic3r/PrintApply.cpp
+++ b/src/libslic3r/PrintApply.cpp
@@ -1271,7 +1271,8 @@ Print::ApplyStatus Print::apply(const Model &model, DynamicPrintConfig new_full_
             print_variant_index[e_index] = e_index;
         }
     }
-    std::vector<int> filament_maps =  new_full_config.option<ConfigOptionInts>("filament_map")->values;
+    auto opt_filament_map = new_full_config.option<ConfigOptionInts>("filament_map");
+    std::vector<int> filament_maps = opt_filament_map ? opt_filament_map->values : std::vector<int>{1};
 
     // Find modified keys of the various configs. Resolve overrides extruder retract values by filament profiles.
     DynamicPrintConfig   filament_overrides;
@@ -1351,7 +1352,7 @@ Print::ApplyStatus Print::apply(const Model &model, DynamicPrintConfig new_full_
     auto opt_extruder_type = dynamic_cast<const ConfigOptionEnumsGeneric*>(new_full_config.option("extruder_type"));
     auto opt_filament_volume_maps = dynamic_cast<const ConfigOptionInts*>(new_full_config.option("filament_volume_map"));
     auto opt_nozzle_volume_type = dynamic_cast<const ConfigOptionEnumsGeneric*>(new_full_config.option("nozzle_volume_type"));
-    for (int index = 0; index < filament_maps.size(); index++)
+    for (int index = 0; opt_extruder_type && opt_nozzle_volume_type && index < filament_maps.size(); index++)
     {
         ExtruderType extruder_type = (ExtruderType)(opt_extruder_type->get_at(filament_maps[index] - 1));
         NozzleVolumeType nozzle_volume_type = (NozzleVolumeType)(opt_nozzle_volume_type->get_at(filament_maps[index] - 1));
--- a/src/libslic3r/PrintConfig.cpp
+++ b/src/libslic3r/PrintConfig.cpp
@@ -7907,6 +7907,10 @@ std::vector<int> DynamicPrintConfig::update_values_to_printer_extruders(DynamicP
         //int extruder_count = opt_nozzle_diameters->size();
         auto opt_extruder_type = dynamic_cast<const ConfigOptionEnumsGeneric*>(printer_config.option("extruder_type"));
         auto opt_nozzle_volume_type = dynamic_cast<const ConfigOptionEnumsGeneric*>(printer_config.option("nozzle_volume_type"));
+        if (!opt_extruder_type || !opt_nozzle_volume_type) {
+            BOOST_LOG_TRIVIAL(warning) << __FUNCTION__ << boost::format(", Line %1%: extruder_type or nozzle_volume_type not found, skipping")%__LINE__;
+            return variant_index;
+        }
 
         if (extruder_id > 0 && extruder_id <= static_cast<unsigned> (extruder_count)) {
             variant_index.resize(1);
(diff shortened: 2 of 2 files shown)


lanewei120:
Hi @niccolodevries 

sorry for the late reply and thanks for this patch

we have fixed serveral related crashes under CLI mode
could you have a try again without this patch?

niccolodevries:
Hi @lanewei120 — thanks for the heads up, and no worries on the timing.

Tested today against current `master` (commit `1e3ee7b`, "FIX: avoid debug asserts in multifilament project loading", 2026-05-03) on macOS arm64. The bug still reproduces, and porting this PR's guards onto `master` fixes it.

### Doing the same as the PR description (unchanged)
```bash
cd resources/profiles/BBL
./BambuStudio --debug 5 \
  --load-settings "machine/Bambu Lab X1 Carbon 0.4 nozzle.json;process/0.20mm Standard @BBL X1C.json" \
  --load-filaments "filament/Bambu PLA Basic @BBL X1C.json" \
  --slice 0 \
  --export-3mf out.3mf \
  3DBenchy.stl
```

### 1. Current master — still crashes (exit 139, SIGSEGV)

Last log lines before the segfault:
```
[info]  apply, Line 1119: enter
[info]  update_values_to_printer_extruders, Line 8252  ... × 3 (succeeds)
[info]  update_values_to_printer_extruders_for_multiple_filaments, Line 8454: extruder_count 1, extruder_nozzle_volume_count 1
[info]  update_values_to_printer_extruders_for_multiple_filaments, Line 8458: different nozzle volume processing
<segfault, exit 139, no 3mf produced>
```

Crash is somewhere between `PrintConfig.cpp:8459` and the next log statement at line 8488, the unguarded `option<...>(key)->...` derefs in `update_values_to_printer_extruders_for_multiple_filaments`.

### 2. Current master + this PR's guards ported on top — slices cleanly (exit 0)

I cherry-picked this PR's NULL guards (the `filament_map` / `opt_extruder_type` / `opt_nozzle_volume_type` early-returns and the `if (!opt) continue;` lines in the switch blocks of both `update_values_to_printer_extruders*` functions, plus the matching guards in `Print::apply()`) onto current master. Same input now slices end-to-end:

```
[info]  update_values_to_printer_extruders_for_multiple_filaments, Line 8469: different nozzle volume processing
[debug] default_status_callback: percent=5, message=Slicing mesh
[debug] default_status_callback: percent=15, message=Generating walls
[debug] default_status_callback: percent=25, message=Generating infill regions
... (all layers)
[debug] default_status_callback: percent=80, message=Generating G-code: layer N
[info]  will export 3mf to .../out.3mf
exit 0   →   3.5 MB out.3mf
```

Interestingly, none of the early-return warnings (`filament_map not found` / `extruder_type or nozzle_volume_type not found`) fired on this run, so the deref that was actually crashing is one of the `opt = this->option<...>(key); opt->get_at(...)` pairs inside the switch block, where the `if (!opt) continue;` silently skips a key that's defined in the schema but absent from the merged config in CLI mode. The PR description called out `filament_map` as a likely culprit; in practice it's the per-key option lookups in the switch that bite first on this input.

### Source-level diff against master

A quick `diff` confirms none of the PR's guards have landed independently:
- `PrintConfig.cpp::update_values_to_printer_extruders` — no NULL check on `opt_extruder_type` / `opt_nozzle_volume_type`, no `if (!opt) continue;` in the switch cases.
- `PrintConfig.cpp::update_values_to_printer_extruders_for_multiple_filaments` — same, plus the unguarded `filament_map->values` deref.
- `PrintApply.cpp::Print::apply()` — `new_full_config.option<ConfigOptionInts>("filament_map")->values` still unguarded; the `filament_map_2` loop still doesn't short-circuit on missing `opt_extruder_type` / `opt_nozzle_volume_type`.

There's also a third site `update_filament_config_values_for_multiple_extruders` (added after this PR was opened), which has the same unguarded `printer_config.option<ConfigOptionInts>("filament_map")->values` pattern at the top. Doesn't trigger on this single-extruder X1C repro, but worth a NULL guard for the same reason. Happy to extend this PR or open a follow-up, let me know which you'd prefer.

Happy to rebase onto current master if you'd like to merge.

lanewei120:
@niccolodevries 
thanks for doing this

PR #10086, opened 2026-03-26
Fix auto-arrangement boundary for skirt loops

ggriffiniii:
The `get_real_skirt_dist` function previously only returned the configured `skirt_distance`, ignoring the actual physical width of the skirt loops themselves. This caused the auto-arrangement algorithm to not allocate sufficient space, potentially placing skirt loops outside the buildable area.

This commit updates `get_real_skirt_dist` to calculate and include the total width of the skirt loops in its returned distance. The width is derived from `initial_layer_line_width`, falling back to `line_width`, `nozzle_diameter`, or a safe default of 0.4mm. It also ensures proper loop handling when a draft shield is enabled.

Additionally, unit tests have been added to verify this correct boundary calculation logic.


ShanChang-BBL:
thanks for your useful commit, the fix would be merged after extra improvements are added

lanewei120:
Hi @ggriffiniii 
thanks for doing this

PR #10088, opened 2026-03-26
ENH: Add 3Dconnexion "SpaceMouse Pro Wireless Bluetooth Edition" product code to properly save settings.

mpaperno:
This "Bluetooth edition" is a relatively new device, adding BT support in addition their existing wireless protocol (and no more sticky rubber coating, yay).  

When connected directly via USB or via BT, it has a unique product code (otherwise it uses the "Universal Receiver" which is already supported).  The mouse worked fine, but customized Studio settings were not being properly loaded/associated for the device when Studio restarted.

Tested on Windows 10 & 11.

Thanks,
-Max


Haidiye00:
agree

mpaperno:
Ping...  @lanewei120 @Haidiye00   Approved?

guanyun-gudujian (approved):

lanewei120:
thanks~ @mpaperno

Issue #8196, opened 2025-09-16
Pressure advance setting is always on "Default"


NVNDO:
### Bambu Studio Version

2.2.1.60

### Where is the application from?

Bambu Lab github releases

### OS version

Windows 10

### Printer

P1S

### How to reproduce

- Load a STL or 3mf
- Select the desired printer
- Sync filaments 
- Check selected PA value in the Device tab

### Actual results

It will always show the "Default" value of 0.020 (for PETG)  

<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/338677b8-1815-43c4-8d42-3112eaf11dc9" />

NOT even listing the existing values:

<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ce3dce17-2c31-4f02-99e2-167e887848ae" />

### Expected results

Once a printer is selected - the manual calibration values for the corresponding filaments in AMS should be selected automatically! If not then... Then latest when one synchronizes the AMS.

It can not be, that every time I want to print something, I have to wait and make all kind of random clicks in order to get my PA-values to show up and then having to manually select them one by one.

### Project file & Debug log uploads

[Empty.zip](https://github.com/user-attachments/files/22362594/Empty.zip)

### Checklist of files to include

- [ ] Log file
- [ ] Project file

d3lt4mik3:
Same problem here. The workaround its to go to Flow dynamics test > Manage Results to open the list, close the list anda then the K values appear on AMS

gary-wlg:
I have been seeing this issue for at least the last few versions of BambuStudio.  There is a workaround I have found.

If you click on the Calibration tab at the top, and then go back to the Device tab, the PA values will show correctly.

I am seeing this on MacOS 15.6.1 with BambuStudio 2.2.x including 2.2.2.56.  I use 2 different MacBook Pro's and I consistently see this issue on one, but not as much on the other.

JinGangBambu:
Hello, thank you for your feedback. It’s possible that the issue you encountered is caused by a corrupted configuration file. Please try the following steps to delete all configuration files. After that, reopen Bambu Studio — it will automatically re-download the plugin and configuration environment. Please check if this resolves your issue.

<img width="1125" height="716" alt="Image" src="https://github.com/user-attachments/assets/f2478ca3-d9bd-4458-950c-3d44a0db61c4" />

[...]

Issue #9325, opened 2026-01-09
PA Profiles saved in the printer are not available in BS


vanowm:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab GitHub releases

### OS version

Windows 10

### Printer

X1C

### Printer Firmware Version

01.08.02.00

### How to reproduce

Device -> Status -> AMS Material settings -> PA Profile has only "Default" profile

### Actual results

It doesn't show available calibration profiles for current filament, even though in printer itself they are available.

### Expected results

List of available PA Profiles

### Project file & Debug log uploads

<img width="383" height="421" alt="Image" src="https://github.com/user-attachments/assets/30020446-9a32-4ce8-b9cf-1db550d478de" />

### Checklist of files to include

- [ ] Log file
- [ ] Project file

Mujin-code:
Hello, thank you for your feedback. You can try restarting the software after checking the file status in the configuration folder. The issue has been resolved in the latest version of the configuration file

vanowm:
Sorry, what file status?

Mujin-code:
Sorry I didn't make myself clear. You can delete all these files and restart the software before trying again. If the retrieval still fails, please provide us with the log files in this folder
<img width="587" height="501" alt="Image" src="https://github.com/user-attachments/assets/a3ae0efb-c68e-4e30-a7ba-8a1c7620ee38" />
<img width="575" height="538" alt="Image" src="https://github.com/user-attachments/assets/c893707f-65af-4dd5-ae87-c4a3deee115d" />

vanowm:
Resetting settings didn't help.
Here is log files after opening BS and attempting set PA profile in 2 filaments in AMS

[log.zip](https://github.com/user-attachments/files/24739072/log.zip)

TXJustin:
I have found that if you go to the calibrations tab and click "manage result", you will see your previously entered PA Profiles.  Make no changes, but close that window.  Now back on the Devices tab, edit your filament and the PA Profile drop-down is now populated again.
This is for the currently selected printer only and goes back to the broken state if you switch away.  Changing to a different printer (printer 2) requires you to do the process again on printer 2.  Switching back to printer 1 and the PA profiles are again gone, requiring another trip to the calibrations tab.

vanowm:
@TXJustin nice find! Works well as a work around. Is it just me or it takes almost 5 seconds for the list to load in `Manage Results`?

shrx:
[...]

Issue #5289, opened 2024-11-15
Ghost K-Factor offset (Flow Dynamics Calibration) in new release.


MooNSparky:
### Bambu Studio Version

1.10.0.89

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 10

### Printer

Bambu Lab X1 Series 

### How to reproduce

1.1 Use auto calibration with the start of the print
1.2 Use previously used K-Factor settings which were working fine before
2. Used K-Factor from Par. 1.1 or 1.2 settings generate bad results
3. Add few points to the "Bad" K-Factor setting manually
4. Used K-Factor from Par. 3 settings generate good results

### Actual results

K-Factor using Lidar Auto Flow Calibration, 6 spools of the same PETG filament: 0.032-0.036 (0.035 chosen - just as before): the same result comes with selecting of Auto Calibration before printing 
![photo_2024-11-15_17-29-32](https://github.com/user-attachments/assets/ad210101-031a-4d5d-8c9b-3386b76428f5)

### Expected results

Manual increase of K-Factor to 0.040 and tying it to spools in AMS
![photo_2024-11-15_17-29-33](https://github.com/user-attachments/assets/d450aba1-3bc7-40f3-b333-83222a155125)

### Project file & Debug log uploads

No files are needed. Same results come from using previous settings or auto-calibration on my X1C and X1E. So, every old or auto recieved K-Factor is needed to be modified manually on my side. 

### Checklist of files to include

- [ ] Log file
- [ ] Project file

zhimin-zeng-bambulab:
@MooNSparky  If you perform automatic calibration again, can the new k value obtained meet the print quality requirements?

shuizhiguo:
**Please upload log through Handy APP, and give us your ticket number**
***
***Upload log files through Bambu Handy***
     Log files from the Bambu printer can be uploaded via Bambu Handy over the network, with an upload speed of around 150 kbps under optimal network conditions.
     When uploading logs, you have the option to select a specific time frame, allowing you to reduce the volume of logs transmitted and shorten the upload duration.
     To upload files, go to **Me** → **Support Tickets** → Choose the ticket for log upload and proceed with the operation.
     ![app_support_ticket](https://wiki.bambulab.com/p1/troubleshooting/upload-log/app_support_ticket.png)

MooNSparky:
@zhimin-zeng-bambulab @shuizhiguo 

Hello!

[...]

Issue #4275, opened 2024-06-13
Unable to select PA profile or enter Factor K


patrickcheah:
### Bambu Studio Version

1.9.2.57

### Where is the application from?

Bambu Lab Official website

### OS version

macOS Sonoma 14.5

### Additional system information

Apple M1 Pro MacBook Pro 2021

### Printer

Bambu P1S

### How to reproduce

In the Device tab, interacting on the filaments on my AMS:
1) Factor K has no value
2) Drop down for PA profile is not working
<img width="595" alt="Screenshot 2024-06-13 at 11 50 10 AM" src="https://github.com/bambulab/BambuStudio/assets/658805/976b10d5-0c96-4703-92c5-f6d63ca48b8b">
<img width="593" alt="Screenshot 2024-06-13 at 11 50 04 AM" src="https://github.com/bambulab/BambuStudio/assets/658805/e0479382-b70f-4455-872e-6f545c8aca1d">

### Actual results

In the Device tab, interacting on the filaments on my AMS:
1) Factor K has no value
2) Drop down for PA profile is not working
<img width="595" alt="Screenshot 2024-06-13 at 11 50 10 AM" src="https://github.com/bambulab/BambuStudio/assets/658805/976b10d5-0c96-4703-92c5-f6d63ca48b8b">
<img width="593" alt="Screenshot 2024-06-13 at 11 50 04 AM" src="https://github.com/bambulab/BambuStudio/assets/658805/e0479382-b70f-4455-872e-6f545c8aca1d">

### Expected results

1) Factor K value should be present and I can update it
2) PA profile should be available to be selected with the dropdown

### Project file & Debug log uploads

[Uploading Magsafe Phone and Watch Stand(2).3mf.zip…]()

### Checklist of files to include

- [X] Log file
- [X] Project file

gitbreiti:
Same problem here. Seems to have come with the new firmware update (01.06.00.00) for P1P/P1S

TaylorN15:
> I have found it. It is not a firmware error. You can create a profile under Dynamic Flow Calibration and then select it.

I found this also, but my issue is that you have to create a profile for every type of filament you have. Even though I have done this before, it wasn't saved like now. I would prefer to be able to choose a profile _or_ enter a value manually.

TinyMito:
I just updated Bambu Studio 1.9.2.57 and latest beta firmware on P1S... it lost all the preset K values under Manage Result in Calibration tab page.

[...]

PR #10106, opened 2026-03-28
FIX issue #8196: AMS slots show "Default" k-profile until Calibration tab is visited

maziggy:
Issue #8196 

PA calibration history is loaded asynchronously via MQTT, but nothing refreshed the AMS material settings dialog when the data arrived. Now the dialog requests PA history on open if not yet loaded, and auto-refreshes the k-profile dropdown once the response arrives.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -1172,6 +1172,19 @@ void AMSMaterialsSetting::Popup(wxString filament, wxString sn, wxString temp_mi
     // Set the flag whether to open the filament setting dialog from the device page
     m_comboBox_filament->SetClientData(new int(1));
 
+    // Request PA calibration history if not loaded yet — needed for k-profile dropdown
+    if (obj->GetCalib()->IsVersionInited() && !obj->GetCalib()->IsPAHistoryReady()) {
+        PACalibExtruderInfo cali_info;
+        int ext_id = obj->get_extruder_id_by_ams_id(std::to_string(ams_id));
+        cali_info.nozzle_diameter        = obj->GetExtderSystem()->GetNozzleDiameter(ext_id);
+        cali_info.use_extruder_id        = false;
+        cali_info.use_nozzle_volume_type = false;
+        CalibUtils::emit_get_PA_calib_infos(cali_info);
+        m_pa_data_pending = true;
+    } else {
+        m_pa_data_pending = false;
+    }
+
     update();
     Layout();
     Fit();
@@ -1183,6 +1196,21 @@ void AMSMaterialsSetting::post_select_event(int index) {
     event.SetInt(index);
     event.SetEventObject(m_comboBox_filament);
     wxPostEvent(m_comboBox_filament, event);
+}
+
+void AMSMaterialsSetting::TryRefreshPAProfiles()
+{
+    if (!m_pa_data_pending || !obj) return;
+    if (!obj->GetCalib()->IsPAHistoryReady()) return;
+
+    m_pa_data_pending = false;
+
+    // Re-trigger filament selection to repopulate PA profile dropdown with real data
+    int sel = m_comboBox_filament->GetSelection();
+    if (sel >= 0) {
+        m_comboBox_filament->SetClientData(new int(1));
+        post_select_event(sel);
+    }
 }
 
 void AMSMaterialsSetting::on_select_cali_result(wxCommandEvent &evt)
--- a/src/slic3r/GUI/AMSMaterialsSetting.hpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.hpp
@@ -110,6 +110,7 @@ public:
                wxString k = wxEmptyString, wxString n = wxEmptyString);
 
     void post_select_event(int index);
+    void TryRefreshPAProfiles();
     void set_color(wxColour color);
     void set_empty_color(wxColour color);
     void set_colors(std::vector<wxColour> colors);
@@ -190,6 +191,7 @@ protected:
     int                 m_filament_selection;
 
     int m_pa_cali_select_id = 0;
+    bool m_pa_data_pending{false};
 
 #ifdef __APPLE__
     wxComboBox *m_comboBox_filament;
--- a/src/slic3r/GUI/StatusPanel.cpp
+++ b/src/slic3r/GUI/StatusPanel.cpp
@@ -3270,7 +3270,12 @@ void StatusPanel::update_ams(MachineObject *obj)
 {
     // update obj in sub dlg
     if (m_ams_setting_dlg && m_ams_setting_dlg->IsShown()) { m_ams_setting_dlg->UpdateByObj(obj); }
-    if (m_filament_setting_dlg) { m_filament_setting_dlg->obj = obj; }
+    if (m_filament_setting_dlg) {
+        m_filament_setting_dlg->obj = obj;
+        if (m_filament_setting_dlg->IsShown()) {
+            m_filament_setting_dlg->TryRefreshPAProfiles();
+        }
+    }
 
     if (obj && obj->GetCalib()->IsVersionExpired() && obj->is_security_control_ready()) {
         obj->GetCalib()->SyncCalibVersion();


BENgin33r:
What's the timeline for a review and merge?  I'm still impacted by this bug.

src/slic3r/GUI/AMSMaterialsSetting.cpp
hemai249, line 1: Thank you for your commit！The data refresh function has already been called in statusPanel→update_ams→emit_get_pa_calib_infos, so TryRefreshPAProfiles is not needed. 
We are currently investigating and fixing this issue. If possible, please provide a log to help us with the development.

hemai249: @maziggy

maziggy: Thanks @hemai249 — happy to provide logs and clarify the design.

  Quick clarification on `TryRefreshPAProfiles`: it isn't a data-fetch duplicate of `emit_get_pa_calib_infos`. It's the UI-side repopulate that runs *after* the MQTT response lands. The two halves of the bug are independent:

1. **Data fetch on dialog open.** `Popup()` adds the `emit_get_PA_calib_infos` call so the request goes out when the    AMS Material Settings dialog is opened directly (no Calibration tab visit). If `statusPanel::update_ams` already fires it on a regular MQTT update, that covers users who came from the device page after a refresh tick — but the dialog can be opened before any update_ams runs (fresh launch + no print active), and then the dropdown is permanently "Default" until the next periodic tick.

2. **Dropdown repopulate on response arrival.** This is the part `TryRefreshPAProfiles` solves and I think the comment may have missed: even when the request is fired (from any source), nothing in the existing code re-triggers the filament dropdown's onSelect path when the PA history payload arrives via MQTT. The dropdown
 was built once at dialog open with `cali_idx → "Default"` because PA history was empty at that moment, and it stays that way for the dialog's lifetime. `TryRefreshPAProfiles` just checks `IsPAHistoryReady()` and re-fires `post_select_event` so `update_pa_profile_items` rebuilds the list with the now-loaded data. Without this, even firing `emit_get_PA_calib_infos` correctly doesn't fix the dropdown — the user has to close and      reopen the dialog.

Happy to drop the `Popup()` half if you confirm `update_ams` is the canonical fetch path and bracket it with a flag so we don't double-fire. The repopulate half is the load-bearing piece either way.

Logs incoming — I'll attach a debug-build trace from a fresh-launch repro (X1C, single-nozzle, AMS slots showing Default until the Calibration tab is opened) in a follow-up comment so you can see the exact sequence: dialog Popup, MQTT push_status with cali_idx, PA history arrival, dropdown still showing Default.

[studio_Wed_May_06_10_20_15_2565_enc.log.0.zip](https://github.com/user-attachments/files/27431571/studio_Wed_May_06_10_20_15_2565_enc.log.0.zip)

hemai249: Hi! @maziggy  Thanks for your log files. In fact, updating data from update_ams is the standard data update path, and TryRefreshPAProfiles is unnecessary.

maziggy: Thanks @hemai249. Need to push back here, because the bug isn't theoretical and isn't going away on its own — users have been hitting this for a long time (issue #8196 has been open with steady reports).

Re-reading the rebuttal: update_ams is on the MQTT push cadence. The reproducible failure is opening the AMS Material Settings dialog before the first update_ams tick fires (fresh launch, no active print, idle printer). At that moment emit_get_PA_calib_infos has not been called, the PA history is empty, and the dropdown is locked to "Default" until the next periodic tick — which on an idle printer can be tens of seconds or longer, well past when the user has already closed the dialog. That is exactly the path the logs I sent demonstrate.

So update_ams is the standard steady-state update path, but it does not cover the cold-open case. That is why Popup() triggers emit_get_PA_calib_infos directly, and that is why the UI needs a repopulate hook (TryRefreshPAProfiles or whatever you'd prefer to call it) once the response lands — otherwise the dialog displays stale "Default" until the user closes it  and reopens.

Two clear questions:

1. Is there an internal fix for #8196 actively in progress? (Your first reply said yes; I want to confirm that's still the case.)
2. If yes, does it cover the cold-open / idle-printer scenario described above? If it does, I'm happy to close this PR and wait for it. If it doesn't, this PR — or an equivalent — is needed.

 Logs, repro steps, and a video are available if any of the above is unclear. This is a real bug affecting real users for a long time, and I'd like to see it actually fixed rather than deferred.

BENgin33r:
@maziggy thanks for fighting the good fight.

src/slic3r/GUI/AMSMaterialsSetting.cpp
hemai249: Thank you very much @maziggy !🎉
1. We are already working on fixing this issue. However, since we have been unable to reproduce the problem, the fix is not progressing smoothly. We sincerely apologize for any inconvenience this may have caused you.
2. update_ams is continuously triggered along with the printer data push — as long as the printer is connected, the data here will keep updating, and emit_get_PA_calib_infos will also be called continuously. Therefore, the cold boot scenario is covered.
Additionally, after further investigation, we believe the issue is more likely related to: obj->GetCalib()->IsVersionExpired() && obj->is_security_control_ready()
We have made an attempted fix for this, which will be included in the next release.

maziggy: One thing worth surfacing — this bug is significantly older than #8196. The same symptom appears in #3376 (Jan 2024), #4275 (Jun 2024), #5289 (Nov 2024), #9325 (Jan 2026), and #10267 (Apr 2026, explicitly worse after v02.05.03). That's 2+ years across multiple Studio versions and multiple printer models, with 60+ aggregate comments. Given that, and given your own comment now identifies a specific guard condition (IsVersionExpired() && is_security_control_ready()) as the likely cause, the bug clearly has been observable internally — the gap may be that the cold-open / idle-printer path isn't part of internal QA's normal flow rather than an actual missing repro.

I'd really like the next-release fix to land with both halves covered (data fetch on cold open + dropdown repopulate on response arrival), so users on this PR's 7+ duplicates aren't back here in 6 months reporting the same thing on v02.07.

hemai249: Hi @maziggy ！ After careful consideration, your fix is reasonable and provides protection for the failure case in the update_ams path. I apologize for my oversight, and once again thank you for submitting the patch. We will adopt your suggestion, Thank you for your support！

lanewei120:
thanks for doing this~
@maziggy

maziggy:
Thank you and hurray!

BENgin33r:
I just downloaded the latest beta where this should have been fixed. It's still not working correctly for me.

maziggy:
Just tested with version 2.7.1.57 and it's still not working.

hemai249:
Hi! @maziggy. Thanks for your feedback! We added log information for this issue in the latest version. Could you please provide some logs?

maziggy:
Sure....

1. Started BambuStudio
2. Connected to my H2D (k factoes are not shown)
3. Navigated to alibration -> Flow Dynamics -> Manage Results
4. Navigated back to device page - k factors are shown

[logs.zip](https://github.com/user-attachments/files/28594656/logs.zip)

hemai249:
Hi! @maziggy  We've identified a logic error triggered by "Developer Mode." It'll be patched in the next network plugin release. You can bypass the issue by disabling Developer Mode for now.

BENgin33r:
I'm still having the same problem as before. when I use bambu filaments with custom k-values, those load appropriately and automatically. when i select a different brand filament, the k-values do not appear. The k-values are visible in my calibration page but do not load per filament. I turned developer mode off. The same issue occurs. When I select bambu filament manually, the custom k-values are not selectable from the device setting page and default remains the only option.

I can see the calibrated values when I manually edit. I deleted all the config files and let them load from the printer again.

MyMjh:
I'm having the same issue, no PA profiles, only default is shown!

hemai249:
@MyMjh @BENgin33r  for this issue, we've included specific monitoring logs in the latest release. Could you share the latest log file with us?

BENgin33r:
@hemai249  here are the logs. I'm on a mac. updated to 2.7.1.62. still having this issue. I can see the calibrated settings in "manage results" but they don't populate in the devices page and I can only select default.

[logs 6-19-2026.zip](https://github.com/user-attachments/files/29148402/logs.6-19-2026.zip)

maziggy:
> 0.2.5b1-daily.20260622 is installed and the problem persists.

What are you doing? This is the BambuStudio repo!

BENgin33r:
> @hemai249 here are the logs. I'm on a mac. updated to 2.7.1.62. still having this issue. I can see the calibrated settings in "manage results" but they don't populate in the devices page and I can only select default.
> 
> [logs 6-19-2026.zip](https://github.com/user-attachments/files/29148402/logs.6-19-2026.zip)

@hemai249 looks like this is still not fixed. @maziggy has the same issues still. Any updates?

maziggy:
I'm in doubt that this will get fixed in this millenium.

hemai249:
Hi! @maziggy @BENgin33r  We have fixed this issue in Beta version 2.8.0.50.You can update your version and check if it works.

BENgin33r:
> Hi! @maziggy @BENgin33r We have fixed this issue in Beta version 2.8.0.50.You can update your version and check if it works.

I downloaded it yesterday and I'm still having the same problem. it's not populating on the device page and i can still see the calibration results when viewing on the calibration page.

hemai249:
Hi! @BENgin33r Could you provide the lastest logs?

maziggy:
I can confirm it's working in 2.8.0.50.

BENgin33r:
I just deleted my entire config folder and let it repopulate, including install the network plugin. still no custom k values are available to be selected
[log.zip](https://github.com/user-attachments/files/29462203/log.zip)


here are my logs @hemai249 

@maziggy did you do anything to get it to work?

maziggy:
Nope, just started it and connected to my printer. Custom values showed up immediately.

BENgin33r:
@hemai249 Tried a fresh install. deleted bambu studio, network assistant, full config file, reinstalled bambu studio and still nothing. It had me login again and it downloaded the presets. I'm able to see the presets downloaded in the config folder but I cannot select saved calibrations still.  I confirmed I am on the 2.8.0.50 beta version and 02.08.00.51 plugin version. 

I'm using a macbook pro if that matters. I recall there was a recent mac only bug.

hemai249:
Hi! @BENgin33r  Are you using an E3D nozzle? Calibration for E3D nozzles is not supported yet and will be available in a future firmware update.

BENgin33r:
@hemai249 I am using a high flow nozzle. I'm confused why this isn't working though. This is a manual calibration, shouldn't it pull the information over and be available? Can you explain why and how having a different nozzle impacts the software?

It was working up until the end of march just fine; I was able to use my calibrated k-profiles without issue.

hemai249:
@BENgin33r There is a known firmware bug where E3D nozzles are misidentified as High Flow nozzles, but then incorrectly saved as Standard Flow after calibration. This conflict causes the PA value dropdown menu in the AMS settings to appear empty, as it cannot find values matching the current nozzle type. Our firmware team is currently working on a fix based on your feedback.

BENgin33r:
@hemai249 any ETA? can i manually edit the json to work around this issue?

GrafJM:
I have the BIQU Panda Revo High Flow hotend, and the problem seems to exist there too. If I set the nozzle type to "Standard" on the X1, it works; all the PA profiles are displayed. Everything worked perfectly up until the 2.5.2 Public Beta. I hope this gets fixed soon.

hemai249:
Hi! @BENgin33r @GrafJM In fact, Studio does not support calibration for unofficial nozzles. For other nozzles, you can select a nozzle type on the printer and treat it as an official nozzle to run the corresponding logic/workflow. Could you provide the nozzle flow type selected on the printer, and the nozzle information shown in Studio → Device → Printer Parts, and a screenshot of the calibration history results?

GrafJM:
@hemai249 I am not aware of the X1 being able to detect which nozzle is installed. The issue lies with the nozzle type setting. If you select "High Flow" and "Hardened Nozzle," you cannot adjust the K-factor. It worked fine up to the 2.5.2 Public Beta, and it works in OrcaSlicer as well.

BENgin33r:
@GrafJM @hemai249 I had the same experience. worked fine up to 2.5.2 and now it's broken.

any ETA on a fix? do you need me to open a new issue since this one is merged?

PR #10141, opened 2026-04-03
feat: Port Z Anti-Aliasing (ZAA) / Z Contouring feature from adob/Bam…

mproper23:
…buStudio-ZAA

Integrates the ZAA non-planar slicing feature which dynamically varies nozzle Z-height within layers to follow model surface topology, reducing visible stair-stepping (layer lines) on sloped surfaces.

Key changes:
- Point3/Polyline3: 3D point types for per-point Z tracking in extrusion paths
- ContourZ.cpp: Core algorithm using ray casting against mesh for Z adjustment
- ExtrusionEntity: Polyline→Polyline3 conversion for 3D-aware extrusion paths
- GCode: Variable Z-height output with extrusion ratio compensation
- Print pipeline: posContouring step with parallel TBB layer processing
- Config: zaa_enabled, zaa_min_z, zaa_minimize_perimeter_height, zaa_region_disable, zaa_dont_alternate_fill_direction, ironing_expansion
- GUI: "Z Contouring" option group in Print > Quality tab

Based on work by Aleksandr Dobkin (adob/BambuStudio-ZAA). Ported from BambuStudio v2.1.1 to v2.5.0.66.


JohnSmithCYM:
Thanks for the contribution. After reviewing this PR, we've decided not to merge it at this time for this 
feature requires specific conditions (model slope angle, print speed, etc.) and its complexity for end users is too big. We prefer to keep such experimental features in developer mode rather than the main workflow.
We appreciate the effort put into this port again, and we are glad to see more discussions in this feature.

iamapig120:
@JohnSmithCYM Could it be add as one part of developer mode? ZAA is a useful feature.

Issue #8738, opened 2025-11-09
P2S PETG Cooling mode


Scootworks:
### Bambu Studio Version

2.3.1.51

### Where is the application from?

Bambu Lab Official website

### OS version

macOS 26.0.1

### Printer

P2S

### Printer Firmware Version

01.01.00.00

### How to reproduce

make any geometry, export the g-code one in PLA and the other one in PETG

### Actual results

In general, PLA and PETG should use the cooling mode of the printer, but in fact, only PLA is using it and in PETG, it changes to heating mode. Furthermore, even if we disable the AUX fan in PETG, the AUX fans will always run with 50%

**PETG**
<img width="960" height="362" alt="Image" src="https://github.com/user-attachments/assets/2a0f6d10-5568-4bd7-a78c-943a020db623" />

**PLA**
<img width="938" height="488" alt="Image" src="https://github.com/user-attachments/assets/9a30a958-8633-4bb2-868b-d3112aa73585" />

### Expected results

disabled AUX fan

### Project file & Debug log uploads

[test cube.3mf.zip](https://github.com/user-attachments/files/23440017/test.cube.3mf.zip)
[debug_Sun_Nov_09_09_59_12_31403.log.2.zip](https://github.com/user-attachments/files/23440016/debug_Sun_Nov_09_09_59_12_31403.log.2.zip)

### Checklist of files to include

- [x] Log file
- [x] Project file

MadSciSlack:
Yes, I too was confused by PETG using "Heating Mode" instead of cooling mode and my being unable to change it.  I was able to open the door and lid to let some heat out, but we should have some control over the mode during printing.

capsel22:
https://github.com/bambulab/BambuStudio/issues/8801

thepokky:
I have the same problem.

thomasvnl:
This is also happening with the P2S and a custom PLA filament I added when I set the softening temperature too high. Now I lowered it but the AUX fan is still having a mind of its own.

I have opted to print a fan duct that blows the AUX fan air to the top of the printer to prevent warping by direct exposure. I am printing a model now that is using that filament with the following cooling options set:

<img width="934" height="759" alt="Image" src="https://github.com/user-attachments/assets/87b8a61f-52c2-4d24-b5d0-ce4667bc6f48" />

And yet when I look at the output of the machine it is now running the AUX fan at 70%!:

[...]

Issue #9131, opened 2025-12-16
Wrong fan status and mode


theshortcircuit:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11

### Printer

Bambu Lam P2S

### Printer Firmware Version

01.01.01.00

### How to reproduce

Print with a PETG profile. Customize filament profile to set Aux fan to 0 (default anyways for Bambu PETG HF profile).

### Actual results

<img width="573" height="345" alt="Image" src="https://github.com/user-attachments/assets/46607bb8-25c0-4ebe-900b-d0ad3d2e0e3a" />

When printing PETG, printer shows that it is in "heating" mode, even though on the printer it says that "cooling" mode is for PETG. Additionally when checking the fan status from Bambu Studio, it incorrectly shows that the Aux fan is on and the Part fan is off (the opposite should be true).

### Expected results

I think Bambu should be using "cooling" for PETG based on the "cooling" mode description but it always seems to default to "heating" for PETG. Additionally the correct fan status should be shown, but appears that it might be mixing them up on the display.

### Project file & Debug log uploads

[P2S Project - Drawer STANDARD BLANK - P2S(2).zip](https://github.com/user-attachments/files/24195670/P2S.Project.-.Drawer.STANDARD.BLANK.-.P2S.2.zip)

### Checklist of files to include

- [ ] Log file
- [x] Project file

XinZhangBambu:
Thank you for your feedback.   In the recently released P2S firmware version
1. PETG has been adjusted to be suitable for printing in the chamber temperature heating mode. The fan-related text in the Studio software will be updated in the next Studio version.
2. Also, the fan status will be updated in the next Studio version.

didusee:
Seeing the same on a P2S. The reported/active mode does not match what the material actually needs: PLA ends up in heating mode.

This is traceable to the airduct logic in the P2S start G-code, where cooling mode (M145 P0) is only selected when min_vitrification_temperature <= 50; everything else, explicitly including PLA/TPU, is routed to M145 P1 (heating). Since PLA's glass-transition temperature is ~55–60 °C, it sits above the 50 °C gate and is put into heating mode by design.

In the passive P2S chamber this keeps the enclosure hot and produces warping and dimensional/surface issues. Please fix the mode selection (PLA/PETG should default to cooling) and let users see and override the actual airduct mode.

PR #10158, opened 2026-04-06
P2S: handle PETG cooling-mode gating in template gcode

### LINKED ISSUE #8801: P2S can't use cooling when printing PETG
### Bambu Studio Version

2.3.1.51

Bambu Lab Official website, Bambu Lab github releases

### OS version

Win11

_No response_

### Printer

P2S

### Printer Firmware Version

01.01.00.00

### How to reproduce

1. Import model
2. Pick PETG
3. Print
4. You will see the fan is in "Heating" mode and cant be switched back to Cooling

5. Edit PETG filament settings to enable AUX fan 100%
6. Print
7. This is totally ignored, fan is again in heating mode stuck at 50%

### Actual results

Printer and the studio are ignoring cooling settings from Filament options
Can't switch manually to cooling either

Image

Image

Image

### Expected results

Same as on my P1S and X1C.
I'd expect to be able to use cooling when using PETG and be able to turn it on/off on demand

[PETG no cooling.zip](https://github.com/user-attachments/files/23559873/PETG.no.cooling.zip)

- [ ] Log file
- [x] Project file

Matthiator:
## Summary
P2S template gcode currently gates cooling-mode behavior using `min_vitrification_temperature <= 50`.
For PETG profiles this usually resolves above 50, so PETG can end up in heating-mode paths even though PETG cooling behavior is expected.

This patch adds explicit PETG/PETG-CF checks in the P2S template gcode so PETG families are treated consistently in start/layer/toolchange fan logic.

## Changes
- `machine_start_gcode`: include `PETG`/`PETG-CF` in the cooling branch condition.
- `layer_change_gcode`: treat PETG/PETG-CF as cooling materials for first-layer and Xth-layer fan reset logic.
- `change_filament_gcode`: treat PETG/PETG-CF as cooling materials in post-toolchange fan/autocooling logic.

## Why this fixes the issue
The issue is caused by using a vitrification-threshold proxy that does not map well to PETG profiles. By checking filament type directly (`PETG` / `PETG-CF`), PETG no longer falls into heating-mode logic due solely to a >50 vitrification value.

## Scope
- Limited to P2S template machine gcode files.
- No firmware changes.

Fixes #8801
Refs #8738


misterff1:
This seems like a bad idea. Petg prints well in heating mode, in fact they changed it to use heating mode later after the release of the p2s and it was a good change. Sure it should be possible to switch to cooling mode manually, but this is not the way to fix it.

Matthiator:
@misterff1 Thanks for the feedback. I agree chamber heat can improve PETG layer bonding in some cases, but this PR is addressing a different problem.

This PR does not change any temperature targets:
- `overall_chamber_temperature >= 40` stays unchanged
- `M142 ... S40 ...` stays unchanged

What it fixes is the gating logic:
PETG/PETG-CF should not be forced into the heating path only because of the `min_vitrification_temperature <= 50` check. With current defaults, PETG typically resolves above 50, so PETG can end up in heating-mode behavior even when cooling behavior should remain available.

This matches open reports:
- https://github.com/bambulab/BambuStudio/issues/8738
- https://github.com/bambulab/BambuStudio/issues/8801
- https://github.com/bambulab/BambuStudio/issues/9131

It also aligns with Bambu’s own guidance:
- Chamber temperature page recommends avoiding high chamber temperature for low-temp materials (including PETG) and defaulting to 0°C for those.
  https://wiki.bambulab.com/en/software/bambu-studio/chamber-temperature
- Overhang guidance recommends opening front door / top lid for PLA/PETG/TPU to improve cooling.
  https://wiki.bambulab.com/en/filament-acc/filament/print-quality/overhang

So this PR is not “aggressive cooling”; it restores correct PETG/PETG-CF control-path behavior.

I’m happy to discuss chamber target tuning in this same PR thread as a follow-up topic (for example evaluating 30–35°C or 25–30°C), but I kept this change focused on fixing the PETG/PETG-CF gating bug first.

misterff1:
> @misterff1 Thanks for the feedback. I agree chamber heat can improve PETG layer bonding in some cases, but this PR is addressing a different problem.
> 
> This PR does not change any temperature targets:
> 
>     * `overall_chamber_temperature >= 40` stays unchanged
> 
>     * `M142 ... S40 ...` stays unchanged
> 
> 
> What it fixes is the gating logic: PETG/PETG-CF should not be forced into the heating path only because of the `min_vitrification_temperature <= 50` check. With current defaults, PETG typically resolves above 50, so PETG can end up in heating-mode behavior even when cooling behavior should remain available.
> 
> This matches open reports:
> 
>     * [P2S PETG Cooling mode #8738](https://github.com/bambulab/BambuStudio/issues/8738)
> 
>     * [P2S can't use cooling when printing PETG #8801](https://github.com/bambulab/BambuStudio/issues/8801)
> 
>     * [Wrong fan status and mode #9131](https://github.com/bambulab/BambuStudio/issues/9131)
> 
> 
> It also aligns with Bambu’s own guidance:
> 
>     * Chamber temperature page recommends avoiding high chamber temperature for low-temp materials (including PETG) and defaulting to 0°C for those.
>       https://wiki.bambulab.com/en/software/bambu-studio/chamber-temperature
> 
>     * Overhang guidance recommends opening front door / top lid for PLA/PETG/TPU to improve cooling.
>       https://wiki.bambulab.com/en/filament-acc/filament/print-quality/overhang
> 
> 
> So this PR is not “aggressive cooling”; it restores correct PETG/PETG-CF control-path behavior.
> 
> I’m happy to discuss chamber target tuning in this same PR thread as a follow-up topic (for example evaluating 30–35°C or 25–30°C), but I kept this change focused on fixing the PETG/PETG-CF gating bug first.

Maybe I am missing something, but with this PR it is set to cooling mode for PETG, right? That means it will lock it to that mode instead. So if the problem is that you are locked into heating mode, that is what needs fixing. It should not be that petg is defaulted for cooling mode as that would actually be the worse choice out of the two.

I agree that locking to a mode is a bad thing, but again correct me if I am wrong but this looks like it will just lock us into the other one. That is not the solution.

Matthiator:
Thanks, that is a fair point.

To be transparent: I am probably not qualified enough to say that my PR is the technically correct final solution. My intention was not to claim that this is definitely the right fix.

I opened the PR mainly because the related issues have been open for months and, from a user perspective, the problem is still very real on the P2S. I wanted to provide a concrete starting point / trigger for discussion instead of only adding another “same problem here” comment.

What I can say from my own testing is:

- I have been printing PETG successfully on an A1 mini for more than half a year.
- Same PETG, dried before printing.
- Same print files.
- Normal room temperature, around 20–23 °C.
- On the P2S, PETG overhangs are much worse with the default Generic PETG behavior.
- If I manually change the profile/template G-code to allow more cooling behavior, the result improves significantly, although it is still not perfect.

I attached a comparison photo directly to this comment to show the practical impact.

In the photo:
- Right side: default/problematic P2S PETG behavior, with visible sagging on the overhang.
- Left side: improved result after manually changing the cooling/heating-related profile G-code.

Both parts were printed with the same PETG, dried before printing, using the same print file and same room conditions. The main difference is the cooling/heating-related workaround in the profile G-code.

This is not meant as proof that my PR is the correct technical fix, but it shows why I believe the current default behavior causes a real PETG print-quality problem on the P2S.

So I completely agree that the final fix should probably not just “lock PETG into cooling mode” either. I may have chosen the wrong technical approach.

The real point I wanted to raise is that PETG currently seems to end up in a behavior path that produces bad overhang results on the P2S, and the only workaround I have found is to manually edit cooling/heating-related profile G-code.

If there is a better condition or variable in the template G-code that represents the actually selected chamber/cooling mode, I would be happy if someone could point me in the right direction. I can also provide test prints, photos, 3MF files, G-code comparisons, or try suggested changes on my P2S.
left_custom_cooling_right_default_petg_p2s

alexanderbuhler:
> Overhang guidance recommends opening front door / top lid for PLA/PETG/TPU to improve cooling.

This is the point. It's absurd we need to open the door to allow for some highly volatile "environment circulation" that comes with its own set of problems. 

Heat buildup is a problem I had to deal with before. Can't say whether this is the "best" implementation, but controlling fan speeds would be a good way to tune the cooling. With an enclosed printer like the P2S heating is well taken care of by the heat bed alone if you want.

misterff1:
> > Overhang guidance recommends opening front door / top lid for PLA/PETG/TPU to improve cooling.
> 
> This is the point. It's absurd we need to open the door to allow for some highly volatile "environment circulation" that comes with its own set of problems.
> 
> Heat buildup is a problem I had to deal with before. Can't say whether this is the "best" implementation, but controlling fan speeds would be a good way to tune the cooling. With an enclosed printer like the P2S heating is well taken care of by the heat bed alone if you want.

I feel compelled to reply to this but also to response before to make it clear that I do fully agree that cooling mode should be usable. Absolutely! 

My point for this specific PR is that this is not the way to fix it, because it locks you out of heating mode, which is very much a suitable mode to handle PETG in. I don't know where you guys live, but there are huge areas in the world where cooling mode is not the correct strategy for PETG. On the other hand I can also understand that in warmer environments the heat creep becomes a problem rather than a solution.

What is needed here is a different approach: a change on the printer side for how it handles the commands for auto cooling or heating mode, where it allows the user to switch between the two. Just setting it to cooling mode by default may fix the problem for you, but cause issues for many others. That is not the right way to approach it.

But again, I do support that this is brought up and needs to be addressed. Just in a different, more effective way.

PR #10201, opened 2026-04-14
NEW: Add counterbore hole bridging support

szc188:
Port the counterbore hole bridging feature from OrcaSlicer to allow printing counterbore/countersink holes without support material.

A new option "Bridge counterbore holes" is added under Quality settings with three modes:
  - None: default behavior, no bridging applied
  - Partially bridged: bridge only part of the unsupported area
  - Sacrificial layer: create a full sacrificial bridge layer

The implementation detects unsupported circular holes via BridgeDetector and modifies surface types in the perimeter generator to enable bridging over these areas.

Includes localization for 18 languages sourced from OrcaSlicer.

Ported from OrcaSlicer (AGPL-3.0), originally by SoftFever/OrcaSlicer contributors.

**UI**

The new option appears under Quality tab, with a dropdown for the three modes:

image

**Slicing comparison**

**None** — The slicer attempts to print the inner ring in mid-air before bridging, causing the print to fail:

image

**Partially bridged** — Only the unsupported perimeter area is bridged, leaving the center open:

image

**Sacrificial layer** — A full bridge layer covers the entire hole, acting as a sacrificial base for subsequent layers:

image

### DIFF
--- a/src/libslic3r/PerimeterGenerator.cpp
+++ b/src/libslic3r/PerimeterGenerator.cpp
@@ -1,6 +1,9 @@
 #include "PerimeterGenerator.hpp"
+#include "BridgeDetector.hpp"
 #include "ClipperUtils.hpp"
 #include "ExtrusionEntityCollection.hpp"
+#include "ExPolygonCollection.hpp"
+#include "Geometry.hpp"
 #include "ShortestPath.hpp"
 #include "VariableWidth.hpp"
 #include "CurveAnalyzer.hpp"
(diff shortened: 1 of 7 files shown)


lanewei120:
Hi, thanks for this patch
Could you remove the translation-related files first?
Meanwhile, we will review the other files

szc188:
Thanks for the review! I've removed all translation-related files (.pot, .po, .mo) from the commit and rebased onto the latest master. The PR now contains only the 7 source files for the counterbore hole bridging feature. Please take a look when you have a chance.

JohnSmithCYM:
thanks for your contribution and it looks good to me.

lanewei120:
thanks for doing this~

PR #10224, opened 2026-04-15
ENH: Allow using system Blosc

hadess:
To make it easier to test newer versions, or apply system specific patches.


MackBambu (approved):

MackBambu:
@lanewei120 Add a build option to the blosc dependency to facilitate separate compilation on Flathub. This can be merged.

hadess:
> @lanewei120 Add a build option to the blosc dependency to facilitate separate compilation on Flathub. This can be merged.

The goal in Flathub was to be able to easily apply https://github.com/Blosc/c-blosc/pull/392 which breaks the build on newer versions of GCC for ARM.

lanewei120:
thanks

PR #10228, opened 2026-04-16
Allow 3D view preset shortcuts (CTRL + [0-7]) to still work when canvas doesn't have focus

mpaperno:
Currently the 3D view shortcuts only work when the 3D canvas has focus. If the last focus was on the sidebar, for example, the shortcuts are ineffective.  The current focus area usually isn't obvious, and switching focus to the canvas typically requires a mouse click.

There was also an odd situation when a combo box type `Field` had focus in the sidebar -- pressing `CTRL+[1-3]` keys would actually switch the active main tab (between Home/Prepare/Preview, depending on the digit).  Looks like some very legacy code there, since those shortcuts don't switch tabs in any other case.

Suggested solution is to simply pass the view shortcut key events from `MainFrame` through to the current canvas view.  And removing the legacy code in `Field`.

Thanks for your consideration,
-Max

### DIFF
--- a/src/slic3r/GUI/Field.cpp
+++ b/src/slic3r/GUI/Field.cpp
@@ -123,43 +123,10 @@ void Field::PostInitialize()
 
 	BUILD();
 
-	// For the mode, when settings are in non-modal dialog, neither dialog nor tabpanel doesn't receive wxEVT_KEY_UP event, when some field is selected.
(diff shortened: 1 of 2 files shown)


Haidiye00 (approved):
approve

lanewei120:
thanks for doing this

mpaperno:
Thank you!

Issue #8515, opened 2025-10-23
Mouse Scroll Wheel Direction


BenImmdich:
**Is your feature request related to a problem? Please describe.**
When I move the scroll wheel away from me, the objects in BambuStudio come closer. Means, I zoom in. When I move the mouse wheel towards me, objects go further away. I zoom out. For me, this is counter-intuitive. 

**Describe the solution you'd like**
In the CAD programs I use (e.g. Fusion or SketchUp), there is an **option to invert scroll wheel behaviour**. There I can configure that objects move in the direction of scrolling. A simple option in the preferences would do perfectly :)

**Describe alternatives you've considered**
None.

**Additional context**
None, just hope :)

Issue #6520, opened 2025-04-14
Mouse improvements


DolphTheRhubarbe:
I'd like to reduce the sensitivity of the mouse when i rotate the view. It's actually way faster than every other 3D software.

Also, the right and the middleclick have the same functions, when the left click is used to select a piece AND rotate camera. Move the rotatory camera to the right click could be a really great and simple upgrade of bambu studio !

Thanks, have a great day.

FelixKainz:
I also have the issue that the rotation speed is way too high. It makes it very annoying and unintuitive to use.

Issue #5639, opened 2024-12-28
Bambu Studio's, View Controls


ShoxDanger:
Can we have controls like in fusion or any other cad program. when i try to look around the camera is all over the place and not turning where my mouse is.

Haidiye00:
![image](https://github.com/user-attachments/assets/7b81a838-81cf-4ce4-9cbf-22155583d352)
Hello, are you referring to this camera angle control control control？

ShoxDanger:
no just the camera when i press the left mouse button. it always wants to rotate from the center of the build plate and this is not good when you try to turn to color your designs. why not rotate from where my mouse cursor is.

Issue #3273, opened 2024-01-01
reverse mouse.. 3d mouse?


KennyBeaudoin:
we cannot reverse the mouse when when scroll towards or backwards on the model, also the 3d mouse connexion is not working great as all parameters seem to be reversed wich is a pain when you are doing cad and cannot change one to another

MrDB42:
I would like to add: Ability to reassign ALL mouse button functions. 

This is so one can stay consistent with ones preferred CAD software and its mouse button functions. (I agree trying to go back and forth with different functions can be frustrating)

PR #10234, opened 2026-04-16
ENH: 3D Camera Manipulation -- Configurable pan/rotate mouse buttons, new rotation modes, free camera, reverse zoom, & other QoL..

mpaperno:
This PR combines multiple enhancements to camera interactions in the various 3D canvas views.

High level overview of changes (details in commit notes):
* The mouse buttons used for panning and rotating can now be configured as per user preference (including buttons 4 & 5, aka. forward/back buttons on some mice).
  * This also means the rotate action button can now be separated from item selection (item selection is still always on the left button). This allows rotating while clicking on an object in Prepare mode, for example, w/out selecting it.
* Added 3 new rotation "modes" that use various focal points as the rotation center. The full list of modes is now (first 3 are new):
  * **Selection or Cursor** - Rotate around the center of selected object(s), or around cursor position when no selection (always around cursor in Preview mode).
  * **Cursor (always)** - Always around where the cursor was positioned on the initial click, regardless of selection.
  * **Camera Target** - Rotate around whatever the camera is currently focused on.
  * **View Center** - Rotate around the center of the current view window, clamped at zero Z height. _This was previously available when holding CTRL key._
  * **Selected Plate** - Rotate around the center of the currently active build plate. _This was/is the default rotation mode._
* Up to 3 desired rotation modes can be assigned by user preference, depending on modifier key, if any:
  * With no modifier key pressed.
  * With ALT/OPT pressed.
  * With CTRL (Win/Linux) / CMD (Mac) pressed.
* The rotation speed (resolution) can now be configured in Preferences. Additionally, holding `SHIFT` while rotating will reduce the speed (increase the resolution) by half of this configured value.
* Added (restored) the "free camera rotation" option as a toggle on the "View" menu and `CTRL + U` shortcut.
* Mouse wheel zoom direction can now be reversed in Preferences.
* View pan and rotate movements will continue if mouse leaves the canvas area (instead of terminating entirely).
* Prevent inadvertent mouse clicks from interrupting pan and rotate actions.
* Improved transition between using a 3D and regular (2D) mouse to be completely smooth.

No current default behaviors were changed. Transition should be seamless for users.

---

**Updated Preferences dialog** 
Broke out all camera settings into their own section ("Zoom to mouse" moves down there), showing defaults on Windows (on Mac the modifier keys are labeled appropriately):
image

**Rotate button choices:**
image

**The pan button choices add "Right or Middle"** to account for the "legacy" default behavior:
image

**Rotation mode choices** (there is a tooltip explaining the modes):
image

I will try to make a video demonstrating the new behaviors. Hard to find an overlay utility that shows _which_ mouse button was pressed, but I'm working on it.  For now suffice to say this improves my interactions with Studio tremendously. Having the rotate button separate from selection is such a relief.

---

I kept commits separated for easier review or possible rollback. The first commit in the series is atomic but essentially superseded by later ones. At this point could be tricky to squash, but I'm flexible.

**Absolutely any and all feedback is very welcome.**  I tried to keep the options flexible and practical, but I'm just one user. And although I can do basic testing on Linux and Mac, I primarily use Windows, so I (and my testing) may be biased.

Thank you for your consideration,
-Max

---

Related/fixes:
https://forum.bambulab.com/t/move-pan-and-tilt-to-right-mouse-button-left-button-selecting/33857
#8515
#6520
#5639
#3273
#3044
#1416
#795
#429

...?

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -162,10 +162,8 @@ void AppConfig::set_defaults()
     if (get("use_perspective_camera").empty())
         set_bool("use_perspective_camera", true);
 
-#ifdef SUPPORT_FREE_CAMERA
     if (get("use_free_camera").empty())
         set_bool("use_free_camera", false);
-#endif
 
 #ifdef SUPPORT_REVERSE_MOUSE_ZOOM
     if (get("reverse_mouse_wheel_zoom").empty())
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -159,16 +159,6 @@ void AppConfig::set_defaults()
 #endif // _WIN32
     }
 
-    if (get("use_perspective_camera").empty())
-        set_bool("use_perspective_camera", true);
-
-    if (get("use_free_camera").empty())
-        set_bool("use_free_camera", false);
-
-#ifdef SUPPORT_REVERSE_MOUSE_ZOOM
-    if (get("reverse_mouse_wheel_zoom").empty())
-        set_bool("reverse_mouse_wheel_zoom", false);
-#endif
     if (get("enable_append_color_by_sync_ams").empty())
         set_bool("enable_append_color_by_sync_ams", true);
     if (get("enable_merge_color_by_sync_ams").empty())
@@ -181,8 +171,6 @@ void AppConfig::set_defaults()
     if (get("export_sources_full_pathnames").empty())
         set_bool("export_sources_full_pathnames", false);
 
-    if (get("zoom_to_mouse").empty())
-        set_bool("zoom_to_mouse", false);
     if (get("show_shells_in_preview").empty())
         set_bool("show_shells_in_preview", true);
     if (get("enable_text_styles").empty())
@@ -215,6 +203,48 @@ void AppConfig::set_defaults()
         set("3d_middle_tooltip_offset_y", "0.0");
     if (get("cancel_glmultidraw").empty())
         set_bool("cancel_glmultidraw", false);
+        
+    // 3D camera view options
+
+    if (get("use_perspective_camera").empty())
+        set_bool("use_perspective_camera", true);
+    if (get("use_free_camera").empty())
+        set_bool("use_free_camera", false);
+    if (get("zoom_to_mouse").empty())
+        set_bool("zoom_to_mouse", false);
+    if (get("reverse_mouse_wheel_zoom").empty())
+        set_bool("reverse_mouse_wheel_zoom", false);
+    // Rotation speed adjustment ratio
+    if (get("view_rotate_speed_factor").empty())
+        set("view_rotate_speed_factor", "0.8");
+    // Which mouse button is used to rotate 3D view:  left|mid|right|aux1|aux2
+    if (get("view_rotate_mb").empty())
+        set("view_rotate_mb", "left");
+    // Which mouse button is used to pan 3D view:  any|left|mid|right|aux1|aux2
+    // Value is exclusive with "view_rotate_mb" (should not be set to same button).
+    // "any" legacy setting means either right and/or mid btns, whichever aren't used for rotation.
+    if (get("view_pan_mb").empty())
+        set("view_pan_mb", "any");
+    // Which camera rotation type is used when no modifiers are pressed: plate|center|selection|cursor|target
+    // \sa Camera::RotationFocusType
+    if (get("view_rotate_mode_nomod").empty())
+        set("view_rotate_mode_nomod", "plate");
+    // Which camera rotation type is used when CTRL (Win/Linux) or CMD (Mac) is pressed.
+    if (get("view_rotate_mode_ctrl").empty())
+        set("view_rotate_mode_ctrl", "center");
+    // Which camera rotation type is used when ALT/OPT is pressed.
+    if (get("view_rotate_mode_alt").empty())
+        set("view_rotate_mode_alt", "selection");
+
+    // Remove legacy camera control options
+    erase("app", "keyboard_supported");
+    erase("app", "mouse_supported");
+    erase("app", "rotate_view");
+    erase("app", "move_view");
+    erase("app", "zoom_view");
+    erase("app", "precise_control");
+    erase("app", "mouse_wheel");
+
 //#ifdef SUPPORT_SHOW_HINTS
     if (get("show_hints").empty())
         set_bool("show_hints", false);
@@ -368,40 +398,13 @@ void AppConfig::set_defaults()
         set_bool("sync_user_preset", false);
     }
 
-    if (get("keyboard_supported").empty()) {
-        set("keyboard_supported", std::string("none/alt/control/shift"));
-    }
-
-    if (get("mouse_supported").empty()) {
-        set("mouse_supported", "mouse left/mouse middle/mouse right");
-    }
-
     if (get("privacy_version").empty()) {
         set("privacy_version", "00.00.00.00");
-    }
-
-    if (get("rotate_view").empty()) {
-        set("rotate_view", "none/mouse left");
-    }
-
-    if (get("move_view").empty()) {
-        set("move_view", "none/mouse left");
-    }
-
-    if (get("zoom_view").empty()) {
-        set("zoom_view", "none/mouse left");
-    }
-
-    if (get("precise_control").empty()) {
-        set("precise_control", "none/mouse left");
     }
 
     if (get("download_path").empty()) {
         set("download_path", "");
     }
-
-    if (get("mouse_wheel").empty()) {
-        set("mouse_wheel", "0"); }
 
     // helio options
     if (get("helio_enable").empty()) {
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -159,16 +159,6 @@ void AppConfig::set_defaults()
 #endif // _WIN32
     }
 
-    if (get("use_perspective_camera").empty())
-        set_bool("use_perspective_camera", true);
-
-    if (get("use_free_camera").empty())
-        set_bool("use_free_camera", false);
-
-#ifdef SUPPORT_REVERSE_MOUSE_ZOOM
-    if (get("reverse_mouse_wheel_zoom").empty())
-        set_bool("reverse_mouse_wheel_zoom", false);
-#endif
     if (get("enable_append_color_by_sync_ams").empty())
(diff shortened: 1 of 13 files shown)


Haidiye00:
approve

JohnSmithCYM:
@mpaperno Thanks for your contribution first.
1. The number of configurable options introduced in this PR is quite large. While each individual option is reasonable in isolation, the combined effect risks making the settings UI feel overwhelming — especially for casual users who just want things to "work well" without having to understand the difference between orbit modes, pivot strategies, and button mappings.What we'd suggest exploring is a preset-first approach: offer a small set of named presets (e.g. Bambu Default, CAD Style, Blender Style, Legacy) that cover the most common use cases out of the box. 
2. Bambu Studio runs on Windows, macOS, and Linux, and mouse/input behavior varies meaningfully across these platforms. We'd like to make sure this PR accounts for those differences before merging.A few specific areas of concern like macOS users frequently don't have a physical middle mouse button. 

Thanks for your contribution again.

mpaperno:
Hi @JohnSmithCYM , thank you for the feedback.

Regarding all the new options, I agree they definitely add up.  I do think the larger issue is that all the settings are now on one page.  I'm not sure why they were combined like that initially for the Studio fork, but breaking them up again would go a long way to simplifying the settings UI (at least in terms of it feeling overwhelming, as you say).  I think that would be a big help for users in general, and something I'd be willing to work on if there is interest from your side.

Specifically in terms of button/modifier assignments, I'm not opposed to some presets, though besides "default" I don't know what they'd be. To be honest I think the names are somewhat meaningless. There has to be some explanation of what "CAD Style," etc, means.  This seems more confusing than anything.  I also don't know what "works well" for other people (apparently someone long ago thought the current scheme works well... ;-) ).  Seems like the simplest option for a user would be to just leave the settings at defaults.

At any rate, if presets are desirable, I'd want to keep a "custom" choice which reveals the full settings UI as I've proposed it (or similar).  This is the most flexible and doesn't depend on remembering what some scheme name means.

Perhaps a new "Camera Control Scheme" option selector with "Default" and "Custom" choices, with the latter revealing the other options (mouse button assignments and rotation modes).  This would somewhat alleviate the immediate issue of new users being overwhelmed with options, though add another field overall.

--

Regarding mouse buttons, unfortunately it's notoriously difficult to properly detect how many buttons a user's mouse/mice may actually have, especially cross-platform. That's a reason I left the "aux" buttons available for everyone, even though many mince won't have them.  I'm going with the assumption that a user knows how many buttons their mouse has.  

I'm not sure why Mac or any other users with 2-button mice would have any issue here... the left button will of course still need to be shared with either rotate or pan actions, but at least there's now that option.  But certainly if I'm missing something with that, or there could be an improvement, please let me know.

Thanks again, and please let me know your thoughts.
-Max

JohnSmithCYM:
"Camera Control Scheme"  is a better idea, some default profiles is better for most of users but others can costumize their need.  You can change your work by that.

And, I must address this add a lot of options for preference page, it makes a long page to be more loong :( It seems these options is more 'elegant' if we can re-modified preferene page like a camera tab setting. 

Why i mention the crossplatform is due to I want to know if you tested it with different os to make sure the main operation works well, no offense here. 

Thanks for your great work again dude.

mpaperno:
Ah yes, sorry, I should have mentioned in the OP that this was tested for basic functionality on all 3 major OSs.  But my Linux and, especially, MacOS testing ability and time is much more limited, and my primary dev/usage OS is Windows (both 10 and 11).  Definitely would be great to get more testing involved. (I tried to solicit some help or feedback on the forums but no one replied.)

So, next steps:

* I'll create a new combo box choice option in the new Camera Controls section titled "Camera Control Scheme" (unless anyone has a better suggestion). This will have at least 2 selections:
  * Default - Current control settings are in effect, Mouse Button Assignment & Rotation Mode sections are hidden.
  * Custom - The 2 additional configuration sections are revealed and user sets everything as they wish.
* I will create a way to define preset "schemes" with "default" being the first. 
  * Should "custom" settings be preserved independently, so switching to a preset doesn't overwrite any customized options? Not sure if that is important.

Optional / To Consider:

* If anyone can suggest some other presets to ship this with, let me know. Or others could be added later.  I guess I could make something up... but besides "Max's Scheme" I have no idea what 😃
* If we're not really going to have presets, besides default, perhaps "Camera Control Scheme" could instead be just a checkbox, eg. "Customize Camera Controls."  Enabling it would reveal the customization fields.
  * If (when) the preferences dialog is updated to use separate pages, this checkbox could potentially just be removed w/out loosing any functionality.
  * Creating a "scheme presets" system wouldn't really be necessary in this case (besides for possible future use).

Let's confirm the next steps before I make changes.

Separately I will look at modifying the preferences dialog (unless someone else is already working on that).  I think something with a sidebar to select the section, with the settings editor pages on the right. More flexible for future use than tabs across the top, and better use of 16:9 ratio screen space.

Thanks!
-Max

mpaperno:
@JohnSmithCYM  Hi, did you get a chance to review my last message?  I'm holding off on further changes until this is clarified.

Thanks!
-Max

mpaperno:
@JohnSmithCYM  Not sure if you're still interested in working on this, but I did finally deliver the Preferences dialog changes I "promised," in #11459. There's a screenshot at the end of the post showing the new Camera page with all the new settings from this PR moved into there.

MisterAnderson91:
I would like to express a desire to see mouse controls configuration implemented as all other software I'm using has rotate on right-mouse button, and having to use left-mouse is driving me away from Bambu Studio at this point in time. I hope this can be prioritized now that the Prefereneces UI is being updated as mentioned in the comments in #11459.

If creating some presets, please ensure touchpad compatibility (either taking advantage of precision touchpad gestures for pan or ensuring only left-and-right-button are used.

mpaperno:
Hello, I've updated the PR to rebase onto latest `master` branch (v2.8.1 beta), and placed the "Camera Controls" settings in the new "3D" preferences tab as a new sub-section.

I have _not_ created any presets (or preset system) as discussed above. I don't think it is necessary with the preferences pages being broken up now, nor did I get any feedback about what the presets would be.

Could we move on this PR, one way or another?  Thanks!

@Haidiye00 , @JohnSmithCYM, @tonghao-bbl

mpaperno:
**Instructions for testers:**
(https://forum.bambulab.com/t/move-pan-and-tilt-to-right-mouse-button-left-button-selecting/33857/17)

I've downloaded, cleaned up, and am posting here the versions of _Studio_ with the camera features enabled. It is currently the same as **v2.8.1** public beta, plus the proposed camera changes. These versions are built by the GitHub workflows here (the "checks").

**Configurations**: These test versions will, by default, use the same user configuration location as the **release version** ("BambuStudio", _not_ the beta config location "BambuStudioBeta", even though it is based on a beta release). It may be preferable to use a different config location so as not to overwrite anything from an installed release version.  Luckily the latest beta already includes a new feature to be able to specify a config folder location (thanks, me! :).  See this wiki article for more details: https://github.com/mpaperno/BambuStudio/wiki/Program-Configuration-Data-Location

The files are too large to upload here, so I'm hosting them on my own server.  Another option is to go find these in the test build "artifacts" list: https://github.com/bambulab/BambuStudio/actions/runs/29618924452#artifacts  though I'm not sure how long they'll be available, and some more digging through archives may be required.

**Windows**:
https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_Windows_PR10234.zip  
Simply unzip the contents to any location and from inside the `BambuStudio` folder run `bambu-studio.exe`.

**Linux (any Debian-based distribution of a corresponding version)**:
https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_ubuntu-22.04_PR-10234.zip  
https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_ubuntu-24.04_PR-10234.zip  
https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_ubuntu-26.04_PR-10234.zip  
Simply unzip the contained `.AppImage` file to a location of your choice and run it.

**MacOS v26**
This is somewhat complicated by the test version being unsigned.  Also I'm not entirely sure which binary will actually run on which architectures (old Intel vs. Apple M silicon).  If you need a different version for an older MacOS, let me know.

https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_MacOS-26_universal_PR-10234.zip  
Extract archive contents to any location of your choice and then try running `BambuStudio_PR-10234.app` directly (double-click or from a terminal command line).

If the above version refuses to run at all ("application cannot be started," not a signing issue), then try this version instead:  
https://max.paperno.us/3dp/bstudio/pr10234/BambuStudio_MacOS-26_arm_PR-10234.zip  

If it _does_ start on a basic level, you will likely get an error about the app being unsigned.  Follow the usual procedures to unblock an app on MacOS and it should work.  If you don't have administrator privileges, it is still possible to unblock an app from the command line.  If you need further assistance, let me know.

**Network Plugin**

If the network plugin doesn't install automatically for this version (and you still want it), follow the instructions linked in the _Studio_ plugin download error message to manually download and install the corresponding plugin version. If using a custom user configuration location, the plugin files need to be placed in there.

---

To contact me directly send a DM on the BL forums (see post link above).

Your feedback is greatly appreciated!  Thank you!
-Max

AGluk:
Hi. Mouse rotation speed is too high in Bambu.
Also buttons function settings are also very useful.
It would be great to see such options as mentioned here.
Hope them will be added to Bambu Studio.
Thank you!

FelixKainz:
I would also appreciate a camera rotation speed setting. It is way too high on my computer. Is there a config file I can edit to change this with the current release version?

mpaperno:
Hi @FelixKainz ,

> Is there a config file I can edit to change [the camera rotation speed setting] with the current release version?

There is not. Unfortunately the rotation speed ratio is currently coded as a fixed value.  The only way to change it right now would be to adjust the mouse's movement resolution externally (system/mouse settings, some "gaming" mice have this feature on the mouse itself, eg. holding a button will allow for finer movements).

-Max

FelixKainz:
Unfortunate, thanks for the explanation.

guanyun-gudujian (changes requested):
should not be merged for now

We have an internal scheduler to support SpaceMouse

mpaperno:
> We have an internal scheduler to support SpaceMouse

I have no idea what this means.  It's not clear what changes you're requesting.

-Max

rehlers:
I really wish this PR would be accepted. I hate having the left mouse button for rotate.

FelixKainz:
The lack of this setting was one of the main reasons for me to switch to Orca Slicer. It has that feature:
image

PR #10242, opened 2026-04-16
Build and packaging fixes for CMake 4.x + Linux runtime compatibility

Shadowrom2020:
This PR fixes a set of Linux build/runtime issues encountered on CachyOS and newer CMake toolchains, while keeping compatibility with older supported CMake versions.

**Problems addressed**

1. Dependency configure failures under CMake 4 due to removed compatibility with policy versions below 3.5.
2. Boost detection breakage with Boost 1.84 (notably regex) when using legacy FindBoost behavior.
3. Linux link failures on systems without OSMesa (-lOSMesa not found).
4. App startup failures in Snap-hosted environments (libpthread.so.0 / GLIBC_PRIVATE symbol mismatch).
5. AppImage runtime instability caused by bundled glibc core libs.
6. Missing distro setup target for CachyOS/Arch-based package workflows.
 
**What changed**

1. Added CMake 4 compatibility argument (-DCMAKE_POLICY_VERSION_MINIMUM=3.5) across affected dependency recipes: TBB.cmake, PNG.cmake, JPEG.cmake, FREETYPE.cmake, Blosc.cmake, Qhull.cmake, Cereal.cmake, libnoise.cmake, OpenCSG.cmake, TIFF.cmake, OCCT.cmake, NLopt.cmake, wxWidgets.cmake.
10. Disabled wx precompiled headers to avoid cotire try-compile failures under newer CMake: wxWidgets.cmake.
11. Switched Boost discovery to Boost config mode when available (CMP0167-aware), preserving compatibility guards: CMakeLists.txt:387.
12. Made OSMesa optional on Linux instead of unconditionally linking it: CMakeLists.txt:701.
13. Sanitized Snap-injected GTK/GIO environment variables on Linux startup to prevent pulling incompatible core20 runtime libs: BambuStudio.cpp:1389.
14. Updated AppImage script to remove bundled glibc core libs (libpthread, libc, libm, libdl, loader) so system glibc is used: build_appimage.sh.in:18.
15. Added/updated CachyOS distro dependency target and package list: linux.d/cachyos.

**Compatibility notes**

1. Top-level minimum CMake remains 3.13, so older supported versions are still valid.
2. CMP0167 handling is guarded with if(POLICY CMP0167), so older CMake versions safely skip that block.
3. Windows version gating logic in CMakeLists.txt:4 remains intact.

**AI disclaimer**: GitHub copilot (GPT-5.3-Codex) was used to find and fix some CMake specific errors and problems


Shadowrom2020:
Also added a build fix for Ubuntu versions > 24

PR #10250, opened 2026-04-17
Use Slic3r::filament_mixer_lerp for GUI preview mixing

justinh-rahb:
## Wire color preview to FilamentMixerModel

**Fixes the "color preview is not yet fully accurate" disclaimer from the 2.5.3 release notes.**

`FilamentMixerModel.hpp` was added in 2.5.3 but the color preview widget was still computing display colors via linear RGB interpolation. This patch wires the preview through `filament_mixer::lerp()` so the swatch reflects the polynomial pigment model rather than a naive blend.

### What changed
- Color preview now calls `filament_mixer::lerp()` for display-side color computation
- No changes to slicing logic, G-code output, or ratio UI

### Why this matters
Linear RGB interpolation gives perceptually wrong results for pigment mixing — blue + yellow stays muddy instead of going green, red + white goes too bright, etc. The polynomial model in `FilamentMixerModel.hpp` was trained specifically to approximate perceptual pigment behavior, so wiring the preview to it is the intended use.

### Disclosure
I'm the author of `FilamentMixerModel.hpp`. Glad to see it in the codebase, this PR makes it actually get utilized as was intended.

---
*I don't have a Bambu machine capable of testing these prints but the previews now match behaviour of my library as used elsewhere*

### DIFF
--- a/src/slic3r/GUI/MixedFilamentDialog.cpp
+++ b/src/slic3r/GUI/MixedFilamentDialog.cpp
@@ -22,6 +22,7 @@
 #include "Widgets/ComboBox.hpp"
 #include "Widgets/DropDown.hpp"
 #include "Widgets/Label.hpp"
+#include "libslic3r/FilamentMixer.hpp"
 
 namespace Slic3r {
 namespace GUI {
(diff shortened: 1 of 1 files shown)


justinh-rahb:
Current version released:
Screenshot 2026-04-17 at 9 17 27 AM
Screenshot 2026-04-17 at 9 17 33 AM

My fixed version:
Screenshot 2026-04-17 at 9 09 23 AM
Screenshot 2026-04-17 at 9 08 53 AM

justinh-rahb:
Ahh, I see this is superseded by v2.6 now.

PR #10258, opened 2026-04-17
Use filament profile default colour on selection

A-Addinall:
# PR Draft: Use Filament Profile Default Colour When Selecting Filament Presets

## Proposed Title

Use filament profile default colour when selecting filament presets

## Summary

This change updates Bambu Studio so filament selections use the selected filament preset's `default_filament_colour` instead of keeping the previous project or AMS slot colour.

The affected flows are:

- Prepare tab filament preset selection
- Project config update after changing a filament preset
- AMS Materials Setting dialog colour preview

This makes the displayed project filament colour follow the selected filament profile, which is useful when filament profiles are used as the source of truth for material colour.

## Problem

When a user selects a different filament profile, Bambu Studio can keep using the existing `filament_colour` value from the project or AMS slot instead of the selected profile's `default_filament_colour`.

That means the filament type/name can change while the visible colour remains stale.

Example:

- AMS slot or project filament is currently magenta.
- User selects a Beige PLA profile whose `default_filament_colour` is beige.
- The selected profile changes, but the UI may continue showing the old magenta colour.

## Implementation

Added a small GUI helper:

`src/slic3r/GUI/FilamentPresetUtils.hpp`

The helper centralises:

- Resolving selected preset names and aliases
- Finding filament presets by filament id
- Reading `default_filament_colour`

Updated:

- `src/slic3r/GUI/PresetComboBoxes.cpp`
  - `PresetComboBox::update_ams_color()` now prefers the selected preset's `default_filament_colour`.
  - Falls back to AMS slot `filament_colour` when the preset has no default colour.

- `src/slic3r/GUI/Plater.cpp`
  - After a successful filament preset change, project `filament_colour` and `filament_multi_colour` are updated from the selected preset's default colour.

- `src/slic3r/GUI/AMSMaterialsSetting.cpp`
  - The AMS Materials Setting dialog colour chip updates when a filament preset is selected.
  - The value used on Confirm comes from the selected profile colour.

## Behaviour

If the selected filament preset has a non-empty `default_filament_colour`, that colour is used.

If the selected preset does not define a default colour, existing fallback behaviour is preserved.

## Validation

Built locally on Windows with Visual Studio 2022, Release configuration.

Tested with local filament profiles and `.3mf` projects:

- Prepare tab updates to the selected profile colour.
- Saved `.3mf` files retain the corrected project colours.
- Reopening the saved file in official Bambu Studio shows the corrected visual colours.
- AMS Materials Setting dialog colour chip updates when changing the filament profile.

## Known Limitation

Hardware AMS slot updates cannot be fully validated from a locally compiled unsigned Bambu Studio build.

The network/printer layer reports:

`Your software is not signed, and some printing functions have been restricted. Please use the officially signed software version.`

Because of that, local testing can validate the UI and project config behaviour, but actual printer/AMS hardware persistence requires an officially signed build.

## Suggested Reviewer Notes

The change intentionally affects only the selected preset colour resolution and keeps existing fallback behaviour when no default colour is defined.

The AMS hardware command path is not changed directly; the dialog simply uses the selected profile colour before the existing Confirm flow sends the current dialog colour.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -2,6 +2,7 @@
 #include "ExtrusionCalibration.hpp"
 #include "MsgDialog.hpp"
 #include "GUI_App.hpp"
+#include "FilamentPresetUtils.hpp"
 #include "libslic3r/Preset.hpp"
 #include "I18N.hpp"
 #include <boost/log/trivial.hpp>
@@ -1418,6 +1419,7 @@ void AMSMaterialsSetting::on_select_filament(wxCommandEvent &evt)
 
     m_filament_type = "";
     PresetBundle* preset_bundle = wxGetApp().preset_bundle;
+    const Preset* selected_filament_preset = nullptr;
     if (preset_bundle) {
         std::ostringstream stream;
         if (obj)
@@ -1439,6 +1441,7 @@ void AMSMaterialsSetting::on_select_filament(wxCommandEvent &evt)
                         }
                     }
                     if (!it->is_system && !has_compatible_printer) continue;
+                    selected_filament_preset = &(*it);
                     // ) if nozzle_temperature_range is found
                     ConfigOption* opt_min = it->config.option("nozzle_temperature_range_low");
                     if (opt_min) {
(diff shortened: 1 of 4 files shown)


lanewei120:
Hi @A-Addinall 

thanks for this patch~
We have several different input scenarios, and in many cases we need to preserve the existing color while only changing the preset.
Using the default color is not the primary use case.

Issue #10178, opened 2026-04-10
View all object settings crashes app


iainmcphersn:
### Bambu Studio Version

2.5.0.66

### Where is the application from?

Bambu Lab GitHub releases

### OS version

Ubuntu 24.04-4 LTS

### Additional system information

Intel® Core™ Ultra 7 265 × 20
64.0 GiB
Onboard graphics

### Printer

Bambu Lab P1S

### Printer Firmware Version

01.10.00.00

### How to reproduce

Go to "Prepare"
Click on "Process" -- "View all object settings" (beside the "Advanced" slider)
App immediately crashes

### Actual results

<img width="666" height="664" alt="Image" src="https://github.com/user-attachments/assets/04146a4b-67d8-4a06-b695-8af5b3848dc8" />

https://github.com/user-attachments/assets/2fac3a7d-b33d-42eb-a157-99c09f0dad69

### Expected results

A popup console should open, allowing the object settings to be changed

### Project file & Debug log uploads

No project file included as this happens with or without a project open

I just installed this version after upgrading to Ubuntu 24.04
I do not see any logs that were created after the OS update and app install.

### Checklist of files to include

- [ ] Log file
- [ ] Project file

qusheng-bambu:
Hello, you can try the newly released version 2.5.3.61. I tested it locally and it worked fine. If the crash issue still occurs, please upload the relevant log files. Thank you very much.
https://github.com/bambulab/BambuStudio/releases/download/v02.05.03.62/BambuStudio_ubuntu-22.04-v02.05.03.61-20260414220857.AppImage

<img width="1124" height="690" alt="Image" src="https://github.com/user-attachments/assets/ec4d852e-36ac-457b-8b52-6a71de1b3e48" />

The steps to obtain the logs are as follows：
https://wiki.bambulab.com/en/x1/troubleshooting/bambu-studio-log

hadess:
Fix is available in this repository's sources, and applied to the Flathub Flatpak:
https://flathub.org/en/apps/com.bambulab.BambuStudio

Issue #10015, opened 2026-03-18
Crash when clicking "View all object's settings"


lucasfalcao3d:
### Bambu Studio Version

2.5.0.66

### Where is the application from?

Bambu Lab GitHub releases

### OS version

Ubuntu 24.04.4 LTS x86_64

### Additional system information

CPU: AMD Ryzen 7 5800X (16) @ 4.853G
Memory: 32005MiB
GPU: NVIDIA GeForce RTX 4070 SUPER

Bambu Studio crashes click on the "View all object's settings", it only happens on the Flatpak version and not on the AppImage. 

### Printer

A1

### Printer Firmware Version

01.11.02.00

### How to reproduce

1. Using the Flatpak version
2. Click on the "View all object's settings" on the Process tab. 

### Actual results

It should crash. 

### Expected results

It should crash. 

### Project file & Debug log uploads

Don't need a project to test. 

### Checklist of files to include

- [ ] Log file
- [ ] Project file

rouyuanzi-bambu:
Hello, could you please provide the dmp file? This will help us further investigate the issue.
The steps to obtain the file are as follows：
https://wiki.bambulab.com/en/bambu-studio/troubleshoot/crash-freeze-issue

lucasfalcao3d:
I could only found the .log file here.

[studio_Fri_Mar_20_10_55_21_2_enc.log.0.tar.gz](https://github.com/user-attachments/files/26143031/studio_Fri_Mar_20_10_55_21_2_enc.log.0.tar.gz)

hadess:
Fix is available in this repository's sources, and applied to the Flathub Flatpak:
https://flathub.org/en/apps/com.bambulab.BambuStudio

lucasfalcao3d:
Thank you @hadess !

Issue #9423, opened 2026-01-17
Flatpak version 2.4.0.70 closes when clicking on "View all object's settings" button


g1bwt:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab GitHub releases, Built from source by yourself or others

### OS version

Kubuntu 25.10

### Additional system information

Operating System: Kubuntu 25.10
KDE Plasma Version: 6.5.5
KDE Frameworks Version: 6.20.0
Qt Version: 6.9.2
Kernel Version: 6.17.0-8-generic (64-bit)
Graphics Platform: Wayland
Processors: 12 × AMD Ryzen 5 3600 6-Core Processor
Memory: 16 GiB of RAM (15.0 GiB usable)
Graphics Processor: NVIDIA GeForce GTX 1650

### Printer

P1S

### Printer Firmware Version

01.09.01.00

### How to reproduce

1. Open Flatpak 2.4.0.70 version
2. Either click on _Prepare_ tab or load project file. It doesn't seem to matter if a .3mf file is loaded.
3. Click on the _View all object's settings_ button.

### Actual results

Instant closure of app with no warning.

### Expected results

Object settings pop up window is displayed.

The current Appimage 2.4.0.70 (file Bambu_studio_ubuntu-24.04_PR-8834.appimage) from Bambu Labs works as expected with no closure.

I have started the Flatpak Bambu Studio app with a new config in _$HOME/.var/app/com.bambulab.BambuStudio/_ with the same problem (the old folder was renamed).

### Project file & Debug log uploads

[debug_Sat_Jan_17_11_23_42_2.log.0.zip](https://github.com/user-attachments/files/24688961/debug_Sat_Jan_17_11_23_42_2.log.0.zip)

### Checklist of files to include

- [x] Log file
- [ ] Project file

fahaaaxiki:
Hello, since this issue does not occur in the AppImage provided by Bambulab, we recommend you report it to flatpak to see if they can help resolve this issue：https://github.com/flatpak/flatpak/issues

g1bwt:
Thanks, will do.

hadess:
> Hello, since this issue does not occur in the AppImage provided by Bambulab, we recommend you report it to flatpak to see if they can help resolve this issue：https://github.com/flatpak/flatpak/issues

[...]

PR #10303, opened 2026-04-20
FIX: Guard m_side_window against null/stale pointer in ObjectGridTable (fixes Flatpak crash on 'View all object settings')

tinezivic:
## Summary

Fixes a crash when clicking **View all object settings** on Linux Flatpak (GNOME Platform 48). Reported in #9423, #10015, #10178.

## Crash location

Valgrind output from issue #9423 (credit: @hadess):

    Invalid read of size 4 at wxWindowBase::Freeze()
    called from ObjectGridTable::OnSelectCell()
    Address 0x1cc is not stack'd, malloc'd or recently free'd

`OnSelectCell()`, `OnCellValueChanged()`, and `OnCellLeftClick()` all call `m_panel->m_side_window->Freeze()` without verifying that `m_side_window` is still valid, which is the crash path identified in the Valgrind trace.

## Root cause

The crash is a **use-after-free during teardown** caused by a race between `ObjectTablePanel` destruction and pending wxGrid events.

The original destructor:

```cpp
ObjectTablePanel::~ObjectTablePanel()
{
    if (m_top_sizer)
        m_top_sizer->Clear(true);   // (1) deletes m_side_window (it is a child widget in the sizer)
    m_side_window = nullptr;        // (2) nulled too late — pending events can fire between (1) and (2)
    ...
}
```

`m_side_window` is a child widget owned by `m_top_sizer`. `Clear(true)` deletes it. wxWidgets can process pending events (e.g. `EVT_GRID_SELECT_CELL`) during widget destruction — at that point `m_side_window` is already deleted but the pointer has not been nulled yet, so the null-guard in `OnSelectCell()` does not help. The result is an invalid read at `0x1cc` (`m_freezeCount` in `wxWindowBase`), exactly as seen in the Valgrind trace.

The fix is to null the pointer **before** `Clear(true)`:

```cpp
ObjectTablePanel::~ObjectTablePanel()
{
    m_side_window = nullptr;        // (1) null first — pending events now hit the guard safely
    if (m_top_sizer)
        m_top_sizer->Clear(true);   // (2) deletes the widget; no live pointer remains
    ...
}
```

## Changes

- `~ObjectTablePanel()`: null `m_side_window` **before** `m_top_sizer->Clear(true)` to close the race window
- `OnSelectCell()`: early return if `m_side_window` is null
- `OnCellLeftClick()`: wrap Freeze/ValueChanged/Thaw in null check
- `OnCellValueChanged()`: wrap Freeze/ValueChanged/Thaw in null check

The null-guards in the event handlers are still necessary as a defence in depth — the destructor fix closes the primary race, but the guards protect against any other path where `m_side_window` might be null or stale.

## Testing

**Not runtime-tested by the contributor** (no Flatpak build environment available). The fix is based on static code analysis of the destructor and the crash path from the Valgrind trace. The crash has been reproduced locally on BambuStudio 2.5.0.66 Flatpak + GNOME Platform 48 (see comments below). If a maintainer or contributor can confirm the fix resolves the crash, that would be very helpful.

---

**AI Disclosure**: This fix was developed with the assistance of GitHub Copilot (Claude Sonnet 4.6). The diagnosis, code analysis, and validation were performed by the contributor. The AI assisted with code generation and diff review.

### DIFF
--- a/src/slic3r/GUI/GUI_ObjectTable.cpp
+++ b/src/slic3r/GUI/GUI_ObjectTable.cpp
@@ -2554,9 +2554,11 @@ bool ObjectGridTable::OnCellLeftClick(int row, int col, ConfigOptionType &type)
                     //update the right side setting list
                     bool is_object = (grid_row->row_type == row_object);
                     ModelObject* object = m_panel->m_model->objects[grid_row->object_id];
-                    m_panel->m_side_window->Freeze();
-                    m_panel->m_object_settings->ValueChanged(row, is_object, object, grid_row->config, grid_col_2->category, grid_col_2->key);
-                    m_panel->m_side_window->Thaw();
+                    if (m_panel->m_side_window) {
+                        m_panel->m_side_window->Freeze();
+                        m_panel->m_object_settings->ValueChanged(row, is_object, object, grid_row->config, grid_col_2->category, grid_col_2->key);
+                        m_panel->m_side_window->Thaw();
+                    }
                     //m_panel->m_plater->update();
                 }
                 else {
@@ -2575,6 +2577,8 @@ bool ObjectGridTable::OnCellLeftClick(int row, int col, ConfigOptionType &type)
 void ObjectGridTable::OnSelectCell(int row, int col)
 {
     m_selected_cells.clear();
+    if (!m_panel->m_side_window)
+        return;
     m_panel->m_side_window->Freeze();
     if (row == 0 || col == col_filaments) {
         m_panel->m_object_settings->UpdateAndShow(row, false, false, false, nullptr, nullptr, std::string());
@@ -2651,11 +2655,11 @@ void ObjectGridTable::OnCellValueChanged(int row, int col)
         bool is_object = (grid_row->row_type == row_object);
         ModelObject* object = m_panel->m_model->objects[grid_row->object_id];
 
-        m_panel->m_side_window->Freeze();
-
-        m_panel->m_object_settings->ValueChanged(row, is_object, object, grid_row->config, grid_col->category, grid_col->key);
-
-        m_panel->m_side_window->Thaw();
+        if (m_panel->m_side_window) {
+            m_panel->m_side_window->Freeze();
+            m_panel->m_object_settings->ValueChanged(row, is_object, object, grid_row->config, grid_col->category, grid_col->key);
+            m_panel->m_side_window->Thaw();
+        }
         //update volume cell
         /*if (is_object) {
             int next_row = row + 1;
@@ -3171,6 +3175,7 @@ ObjectTablePanel::~ObjectTablePanel()
         delete m_object_grid_table;
         m_object_grid_table = nullptr;
     }*/
+    m_side_window = nullptr;  // null before Clear(true) so pending events hit the guard safely
     if (m_top_sizer)
         m_top_sizer->Clear(true);
     delete m_object_settings;


hadess:
> On Flatpak with org.gnome.Platform/x86_64/48, the runtime provides a newer wxWidgets ABI than the version bundled in the AppImage

ABI shouldn't factor into it at all, because we're recompiling from source.

So either there's a warning during the build which we're ignoring and we shouldn't, or it's a bug that just happens not to crash on the AppImage version because, for example, FORTIFY_SOURCE isn't set during its build.

If the problem is an ABI/API problem, then we should be able to find the exact upstream commit that started causing crashes.

tinezivic:
Thanks @hadess — that makes sense. I've updated the PR description to remove the ABI mismatch explanation and mark the root cause as unconfirmed.

The intent of this PR is narrower: add a defensive null-guard around `m_side_window` in the exact crash path shown by Valgrind (`ObjectGridTable::OnSelectCell` → `wxWindowBase::Freeze()`). Even if the underlying cause turns out to be a suppressed build warning or a `FORTIFY_SOURCE` difference, the guard should still prevent the crash by blocking access to an invalid pointer.

I don't have a Flatpak build environment, so I can't reproduce or further diagnose from my side. If you have a direction you'd like me to investigate (build warnings, compiler flags, a specific wxWidgets upstream commit), I'm happy to look into it.

tinezivic:
## Crash reproduced locally

**Environment:**
- BambuStudio 2.5.0.66 (Flatpak, `com.bambulab.BambuStudio`)
- GNOME Platform 48 (Flatpak runtime)
- Ubuntu 24.04, x86_64

**Steps to reproduce:**
1. Open BambuStudio Flatpak
2. Load any 3MF project with objects
3. Right-click an object → **"View all object settings"**
4. BambuStudio crashes immediately

**Kernel crash log (`journalctl -k`):**
```
bambustu_main[2033063]: segfault at 1cc ip 000059d05153e574 sp 00007fff25cad018 error 4 in bambu-studio[52b1574,59d04c905000+4f17000]
bambustu_main[2027576]: segfault at 1cc ip 000059b8feafa574 sp 00007fffe6282fb8 error 4 in bambu-studio[52b1574,59b8f9ec1000+4f17000]
```

The crash address `0x1cc` matches exactly the Valgrind output from issue #9423:
```
Invalid read of size 4 at wxWindowBase::Freeze()
Address 0x1cc is not stack'd, malloc'd or recently free'd
```

The offset `0x1cc` corresponds to the `m_freezeCount` field inside `wxWindowBase` — the read fails because `m_side_window` points to an already-destroyed window object (use-after-free during teardown), not an ABI mismatch. This is consistent with @hadess's assessment that the issue is likely a pre-existing use-after-free that other build configurations happen to survive.

lanewei120:
@tinezivic 

Do you mean this issue has already been fixed by this patch?
Thanks.

hadess:
> Do you mean this issue has already been fixed by this patch?

The issue is not fixed, and the patches in this MR, as well as the explanations for it, bear the mark of being LLM generated. I wouldn't merge anything.

tinezivic:
Fixed the ordering bug — [m_side_window](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) was nulled after Clear(true) instead of before (8d92274). Can't test locally without a Flatpak environment. If the root cause is elsewhere, I can close the PR.

src/slic3r/GUI/GUI_ObjectTable.cpp
```diff
@@ -2575,6 +2577,8 @@ bool ObjectGridTable::OnCellLeftClick(int row, int col, ConfigOptionType &type)
 void ObjectGridTable::OnSelectCell(int row, int col)
 {
     m_selected_cells.clear();
+    if (!m_panel->m_side_window)
```
hadess, line 2580: This conditional is the only thing that's needed, you can remove the rest of the patch.

hadess (commented):
The commit message also needs a lot of work. It speaks of ABI changes, of stale vtable pointer when the problem is a null pointer dereference. Everything else is made up from my hypothesis.

The commit subject also needs to be changed to say what the fix does ("Fix crash etc."), not how it was implemented.

If you can't or won't make the changes, I can create a new MR and credit you for the help with the root causing.

hadess:
> If you can't or won't make the changes, I can create a new MR and credit you for the help with the root causing.

I created https://github.com/bambulab/BambuStudio/pull/10607 as I wanted to be able to add it straight away to the Flatpak build.

tinezivic:
Closing this — @hadess picked up the diagnosis from here and turned it into a clean minimal fix in #10607, which he's already tested on Flatpak 2.6.1 with wxWidgets 3.2.9. Good to see it land properly!

hadess:
> Closing this — @hadess picked up the diagnosis from here and turned it into a clean minimal fix in #10607, which he's already tested on Flatpak 2.6.1 with wxWidgets 3.2.9. Good to see it land properly!

Thank you for the patches, saved me quite a bit of time in experimentation.

PR #10335, opened 2026-04-22
fix: incorrectly-included `/usr/local/include` in Boost patches

leo9800:
install prefix (`usr/local`) is included by accident in a patch to boost library, which breaks cmake on archlinux when invoking `makechrootpkg`

this PR fixed the paths in the patch file mentioned above


lanewei120:
Hi @leo9800 

just checked the path on mac, 
usr/local/include/boost/process/detail/posix/executor.hpp
it should be correct

maybe we need to exclude this patch for non-mac platform?

leo9800:
@lanewei120 unfortuately nope, it seems to be a installing path which includes the installation prefix (something like `/usr` or `/usr/local`) and it indeed break building on some linux distros, at least a building error linked to this issue was confirmed with arch linux and by applying the patch everything worked like a charm with arch `pkgbuild`

lanewei120:
@leo9800 ok
thanks for doing this

PR #10348, opened 2026-04-23
FIX: StatusPanel, show current-print thumbnail on wxGTK

ClusterM:
On Linux (wxGTK) wxStaticBitmap is a native GtkImage wrapper and does
not dispatch wxEVT_PAINT to Bind()'d handlers. As a result, the
PrintingTaskPanel::paint() override that wxMSW/wxOSX rely on to
composite m_thumbnail_bmp_display over the panel is never invoked, and
set_thumbnail_img() has no visible effect - the placeholder sent to
the widget at construction time stays on screen for the whole session.

Push the bitmap through the widget's native SetBitmap() path in
addition to the existing Refresh() so the new image shows up on all
platforms regardless of whether wxEVT_PAINT fires.


ClusterM:
Rebased/fixed.

XinZhangBambu:
Thanks for the fix! This seems like a solid improvement.

Issue #9738, opened 2026-02-11
Hide non-active plates in Prepare mode


Terhiman:
**Is your feature request related to a problem? Please describe.**
When editing and working on the horizontal plane, parts and plates other than the selected plate are visible in the foreground/backgroud etc very confusing and requires some rotationall jiggling to see the selected plate parts from the side.

**Describe the solution you'd like**
Some toggle to allow hiding/showing of the non selected plates.

vanowm:
Yes please! Try view something from a side when there are 5 plates on the way... Plus it could help with performance.

dmarnel:
I am experiencing the same issue. It's difficult to work with objects on the active plate when objects from other plates are also visible.

Issue #8512, opened 2025-10-23
Toggle switch for selected part only visible while editing


techiesatbir-oss:
It is tough to work on a single part when lots of parts are loaded on multiple plates. View of the part being edited is blocked by other parts

I would request a toggle switch in the top menu of the perspective screen to hide plates/parts other than the selection. 

OR right clicking the selection (one or more parts) may have an option of "Hide other models/parts" 

<img width="1628" height="932" alt="Image" src="https://github.com/user-attachments/assets/7516b840-899d-4be3-946a-1072aa75d454" />

Issue #2558, opened 2023-09-21
Plate names, and more plates


Kingsidorak:
Users should be able to assign names to plates

More plates could be possible through resource limitation. Disable plates once they haven't been active in a certain amount of time, and/or if someone is just clicking on beds then something could be in place to where "the 11th bed last clicked should be disabled", or something like that.
Users could also toggle lock plates to always being active

Issue #1708, opened 2023-04-29
add plate list sidebar to prepare


julie777:
**Is your feature request related to a problem? Please describe.**

When in prepare mode, working with plates is difficult when there are multiple plates all in the workspace. 

Although not a bug, I would consider this a **high priority** issue. I am working with more than ten plates in one project and it is very difficult to manage the preview.

**Describe the solution you'd like**

It should be possible to show/hide plates in the workspace to limit the number of plates currently being shown in order to make managing plates easier.

The same plate list that is used in preview can be used to manage plates, listing all plates. 
  - Clicking on a plate in the side list would show it on the canvas, but not affect other plates on the canvas. This differs from preview when only one plate can be on the canvas.
  - A button can be added to plates on the canvas to "hide" them, or more properly minimize them to the list.

**Describe alternatives you've considered**

**Additional context**

julie777:
@bambu123 
Thank you for assigning this. It is really driving me crazy working in prepare in a project with 26 plates.

I have some addition thoughts after spending the last couple of days working on this project.

### General
1. Plates should work the same way in prepare and preview. In addition to being easier for the users, it allows the code to be the same.

### Plate Naming
1. The context menu should have rename added. 
2. The context menu should be available from the object list, the sidebar entry, and the plate in the workspace
12. Plate names must be shown in the sidebar, instead of just numbers. I would suggest that plate numbers are not shown unless they are part of a default name for the plate which should be changed from "Untitled" to "Plate 1", etc.
13. plate names must be used for gcode files. 
    14. Currently gcode files only have the plate number and that is useless when looking at the SD card file list. 
    15. Gcode file names should be `<projectname>.<platename>.gcode` This works for plates with user given names and when the user leaves the default name, which will be common for projects with only one plate.
16. direct edit of name in the object list and on the name of the plate in the workspace would be a nice addition and work like names in OS file managers and other applications.

### Plate Ordering
1. Plates in the sidebar should support dragging to reposition them in the list. 
2. The list can use hidden plate IDs (numbers) to manage ordering in the list and on the workspace. These identifiers should be hidden from the user.

### Plate Management
1. new plates are shown in the workspace when created
3. Plates should have a new icon added below the X to allow hiding (closing) the plate. A closed plate is in the sidebar list, but not shown in the workspace. Closed plates are not affected by the plates toolbar actions, such as auto-arrage. 
4. clicking on a plate in the sidebar toggles the show/hide attribute of the plate
[...]

PR #10369, opened 2026-04-24
ENH: Add plate visibility toggle

jomixlaf:
## Summary

This PR adds the ability to show and hide individual plates in the 3D workspace, without deleting them. It is a non-destructive workflow tool for users who want to focus on one plate at a time while keeping their full project intact.

## Features

- **Eye icon on each plate** — appears on hover; click to toggle visibility. Hidden when the plate is the only one, or would be the last visible plate (minimum one plate always visible).
- 
image

- **Toolbar button** — eye icon in the top toolbar (visible only when 2+ plates exist) opens a popup with:
  - Show All Plates
  - Hide Other Plates (hides all except the currently selected plate)
  - Per-plate checkmarks to toggle individually

SCR-20260424-la1

- **View menu** — View > Show/Hide Plates submenu with the same Show All / Hide Other Plates options plus per-plate toggles, rebuilt dynamically on open.
- **Right-click object list** — Show All Plates and Hide Other Plates also available when right-clicking a plate name in the object list.
- **Visual feedback** — hidden plates render as faded/ghosted in the 3D view; their objects are invisible and not pickable (except the eye icon slot to re-enable).
- **[hidden] label** — shown next to the plate name in the object list panel.

SCR-20260424-lbg

- **Slice/print blocked** — hidden plates cannot be sliced or sent to print; toolbar buttons are disabled. Auto-reslice on switching to Preview is suppressed for hidden plates.

## Files Changed

| File | Change |
|------|--------|
| `src/slic3r/GUI/PartPlate.hpp` / `.cpp` | `m_visible` flag, eye icon rendering, picking, slice/print guards, `get_visible_plate_count()` |
| `src/slic3r/GUI/Plater.cpp` | Toggle action on eye icon click, slice guard, auto-reslice guard |
| `src/slic3r/GUI/GLCanvas3D.cpp` | `update_plate_volumes_visibility()`, toolbar button + popup menu |
| `src/slic3r/GUI/MainFrame.cpp` | View > Show/Hide Plates submenu |
| `src/slic3r/GUI/GUI_Factories.cpp` | Right-click object list menu items |
| `src/slic3r/GUI/ObjectDataViewModel.cpp` | `[hidden]` label in plate list |
| `resources/images/` | Eye icon SVGs (light/dark, normal/hover, visible/hidden states) + toolbar icon |

## Notes

- Visibility state is intentionally **not persisted** to the 3MF file in this PR to keep the change minimal and avoid format compatibility concerns. It can be added in a follow-up.
- The minimum-one-visible-plate constraint is enforced across all three toggle surfaces (plate icon, toolbar menu, right-click menu).

## Screenshots
SCR-20260424-l1o

I believe there is still work to do, but at least if it's here, we can discuss it :)


Haidiye00 (approved):

Haidiye00:
image
Hello, would it be better to place this icon here?

jomixlaf:
hmmm I personally don't mind both places, but I might affect people's flow if we move the icon to something they are not expecting, a new icon, they'll be mad about? I don't know. My thought was that when I add it, people may notice a new Icon in the corner, next to another recently added icon. If you prefer, you can move it, or feel free to ask me too as well 😃

JohnSmithCYM:
Thanks for your contribution, we want to make it clear: which condition or situation is this feature suitable for? Can you describe it here.

jomixlaf:
It helps with visual navigation in big files. It is not uncommon for me to reopen the same file only to work on a specific plate for a client, having only the plate I want help with zoomed and rotated. To help with that even more, I plan to work on the view cube so that when we press on his face, the active plate zooms in. But that's for later. 
image


I believe this feature will be very well received by the community.
https://github.com/bambulab/BambuStudio/issues/9738
https://github.com/bambulab/BambuStudio/issues/1708
https://github.com/bambulab/BambuStudio/issues/8512
https://github.com/bambulab/BambuStudio/issues/2558

image

JohnSmithCYM:
Thanks for the contribution. One product requirement I’d like to clarify is that plate visibility should probably be treated as a temporary view state, not as an actual plate state. From a user perspective, “Hide plate” feels like a non-destructive workspace/focus tool: it helps reduce visual clutter while editing a large multi-plate project. It should not imply that the plate itself is disabled, excluded from the project, or excluded from future slicing/printing unless we explicitly design that as a separate feature.

I'm not sure if the pr content contains:
- Hidden/visible state should not be saved into the 3MF/project file by default and reopening a project should show all plates again.
- Hiding a plate should not invalidate its slice result or change the underlying plate data.

jomixlaf:
Good morning JohnSmithCYM, 

I agree with you that it should not be saved in the 3MF. That's why I didn't implement it. The mechanism is very simple; if the plate is hidden, it can be selected in the file browser, but cannot be sliced or printed. I made sure there is enough way to let the user know that one or more plates are hidden. I think it's a very elegant way to begin, more ideas and improvements may come from users once it's in the wild.

jomixlaf:
Hi ! Just a friendly ping here, as it was approved but missed merging to main. 

Have a great day 😄

OskisJ:
@jomixlaf 
**/src/slic3r/GUI/PartPlate.cpp**   
m_partplate_list is dereferenced to check get_plate_count() without verifying it is non-null. Since the PartPlate default constructor initializes m_partplate_list to nullptr, any rendering pass or state check on an unattached plate will trigger a null pointer dereference.

maybe write like this?' line 790 partplate.cpp 
```c++
bool can_toggle_visibility = m_partplate_list != nullptr && m_partplate_list->get_plate_count() > 1 &&
                             (!this->is_visible() || m_partplate_list->get_visible_plate_count() > 1);
``` 

it is good to implement some defence. adding a defensive null check ensures the UI rendering loop remains robust and does not crash if a plate's parent list is unassigned or still initializing.

guanyun-gudujian (approved):

lanewei120:
> Hi ! Just a friendly ping here, as it was approved but missed merging to main.
> 
> Have a great day 😄

hello @jomixlaf 

thanks for doing this

please check the comment above, 
and rebase this patch to the newest master base

jomixlaf:
@OskisJ Thanks for the catch! You're right that m_partplate_list is null on a default-constructed plate. 
But render_icons() already dereferences it unconditionally ~37 lines earlier (e.g. line 753 for the delete icon), so a guard only at line 790 wouldn't actually prevent a crash, execution would never reach it. render_icons() is only called from the plate-list render loop over plates the list owns, so m_partplate_list is guaranteed non-null on this path, consistent with the existing upstream code. 

If we want real hardening, I'd add an early if (!m_partplate_list) return; at the top of the function rather than a single mid-body check — happy to do that if you'd prefer the defensive version.

OskisJ:
@jomixlaf 
Thanks for the detailed explanation that makes sense.

You're right: since render_icons() dereferences m_partplate_list earlier in the function, a guard only around the visibility toggle wouldn’t prevent a crash on a default‑constructed plate. My intention was to harden the code, but I agree that a single mid‑function check doesn’t actually guarantee safety.

Go ahead with whichever approach you prefer!

PR #10370, opened 2026-04-24
ENH: Add alternate extra wall option

jomixlaf:
## Summary

This PR ports the **Alternate Extra Wall** feature from OrcaSlicer (which itself ported it from Ultimaker Cura). A big thank you to the OrcaSlicer and Cura teams for the original work.

The feature adds an extra perimeter on alternating layers, improving layer bonding and part strength without the full material and time cost of a permanent extra wall on every layer.

## How it works

- When enabled, odd-numbered layers receive one additional wall beyond the configured wall count
- Even-numbered layers use the normal wall count
- The alternating pattern creates a staggered wall structure that mechanically interlocks adjacent layers

**ON**
SCR-20260424-rlu-2

**OFF**
image

Thanks to Angus Deveson from Maker's Muse for his video on the subject links here --> [https://www.youtube.com/watch?v=c7CI6yBTKMc](url)

## Behavior

- **Spiral vase mode**: automatically skipped (spiral vase forces a single wall anyway)
- **0% infill density**: automatically skipped (no structural benefit without infill)
- **Ensure vertical shell thickness set to Enable**: a warning dialog is shown, offering to automatically switch it to Partial for best results — matching OrcaSlicer's behavior
- 
SCR-20260424-rm8

## Location

Process settings → **Strength** → **Walls** → *Alternate extra wall* checkbox

## Files Changed

| File | Change |
|------|--------|
| `src/libslic3r/PrintConfig.hpp` / `.cpp` | New `alternate_extra_wall` bool config option |
| `src/libslic3r/PerimeterGenerator.cpp` | Apply extra wall on odd layers, both classic and Arachne paths |
| `src/libslic3r/Preset.cpp` | Register option in `s_Preset_print_options` |
| `src/libslic3r/Print.cpp` | Invalidate perimeters step on option change |
| `src/slic3r/GUI/Tab.cpp` | Add checkbox to Strength > Walls UI |
| `src/slic3r/GUI/ConfigManipulation.cpp` | Conflict warning dialog with ensure vertical shell thickness |

## Credits

- Original implementation: [Ultimaker Cura](https://github.com/Ultimaker/Cura)
- Ported to OrcaSlicer by the [OrcaSlicer](https://github.com/SoftFever/OrcaSlicer) team

### DIFF
--- a/src/libslic3r/PerimeterGenerator.cpp
+++ b/src/libslic3r/PerimeterGenerator.cpp
@@ -916,6 +916,9 @@ void PerimeterGenerator::process_classic()
         const Surface &surface = this->slices->surfaces[surface_order[order_idx]];
         // detect how many perimeters must be generated for this island
         int        loop_number = this->config->wall_loops + surface.extra_perimeters - 1;  // 0-indexed loops
+        int sparse_infill_density = this->config->sparse_infill_density.value;
+        if (this->config->alternate_extra_wall && this->layer_id % 2 == 1 && !m_spiral_vase && sparse_infill_density > 0)
+            loop_number++;
         //BBS: set the topmost and bottom most layer to be one wall
         if (loop_number > 0 && ((this->object_config->top_one_wall_type != TopOneWallType::None && this->upper_slices == nullptr) || (this->object_config->only_one_wall_first_layer && layer_id == 0)))
             loop_number = 0;
@@ -1500,6 +1503,9 @@ void PerimeterGenerator::process_arachne()
     for (const Surface& surface : this->slices->surfaces) {
         // detect how many perimeters must be generated for this island
         int loop_number = this->config->wall_loops + surface.extra_perimeters - 1; // 0-indexed loops
+        int sparse_infill_density = this->config->sparse_infill_density.value;
+        if (this->config->alternate_extra_wall && this->layer_id % 2 == 1 && !m_spiral_vase && sparse_infill_density > 0)
+            loop_number++;
 
         bool apply_circle_compensation = true;
         // Orca: properly adjust offset for the outer wall if precise_outer_wall is enabled.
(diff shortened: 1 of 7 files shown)


MAODAN-Bambu:
Hi@jomixlaf, thanks for your PR!!
My understanding is that alternating extra walls are meant to create a kind of 'interlocking teeth' effect to make the model stronger. But in practice, I’m seeing two main problems:
1.When using non-concentric infill, the start/end points of the lines end up hanging in mid-air.
2.Since it’s not 100% density infill, the lines from one layer don’t line up with the layer below.
I’d like to hear your thoughts on this.

Naifuu:
> Hi@jomixlaf, thanks for your PR!! My understanding is that alternating extra walls are meant to create a kind of 'interlocking teeth' effect to make the model stronger. But in practice, I’m seeing two main problems: 1.When using non-concentric infill, the start/end points of the lines end up hanging in mid-air. 2.Since it’s not 100% density infill, the lines from one layer don’t line up with the layer below. I’d like to hear your thoughts on this.

Sorry for jumping into the conversation — I'm just a 3d printing enthusiast hoping to see this feature land.

@MAODAN-Bambu, both of your concerns are valid, with sparse non-concentric infill, parts of the alternate wall do span over voids and don't align with the layer below. But as far as I understand, those are inherent properties of the feature rather than bugs introduced by this PR — the same behavior exists in Cura and Orca, where the feature has been available for years across all infill patterns without being treated as blockers.

The PR also already guards the edge cases where the mechanism wouldn't help: it auto-skips at 0% infill, and warns when "ensure vertical shell thickness = Enable" would conflict.

Maybe additional tooltip or docs note clarifying that this is designed to work with concentric 100% infill would help users understand the intent.

Maker's Muse video shows the strength improvement in practice and gained over 200k viewes in 2 weeks. A lot of bambu users would love to see the feature.

Thanks to both of you for the work on this.

PR #10374, opened 2026-04-25
Fix STEP import: MANIFOLD_SOLID_BREP body names lost for bodies inside root components

jomixlaf:
## Problem

When importing a STEP file that contains bodies (solids) living **directly inside the root Fusion 360 component** — rather than inside a named sub-component — those bodies were always imported with the name **\"SOLID\"**, regardless of their actual name in the STEP file.

Root cause: OCCT's XCAF layer only preserves names from `PRODUCT` entities. `MANIFOLD_SOLID_BREP` entity names are **not** propagated to XCAF labels for shapes that are not their own product. The existing `getNamedSolids()` function read names exclusively from XCAF labels, so any solid whose geometry came from a `MANIFOLD_SOLID_BREP` directly under an assembly product would silently lose its name.

## Fix

After `STEPCAFControl_Reader::Transfer()` completes, iterate all STEP entities via the transfer process (`TransientProcess`) to find every `StepShape_ManifoldSolidBrep`. For each one:
- Read the user-assigned name from the entity
- Resolve the corresponding `TopoDS_Shape` via `TransferBRep::ShapeResult`
- Store a `TShape*` → name mapping

`getNamedSolids()` now consults this map when naming solid shapes. When the XCAF label name is missing or generic, the correct `MANIFOLD_SOLID_BREP` name is used instead.

## Result

Bodies at any level of the assembly hierarchy are now imported with their correct STEP entity names.

| Before | After |
|--------|-------|
| Body at root component imported as `SOLID` | Imported with its actual name (e.g. `logo`, `bracket`) |
| Sub-component bodies unaffected | Sub-component bodies unaffected |

### DIFF
--- a/src/libslic3r/Format/STEP.cpp
+++ b/src/libslic3r/Format/STEP.cpp
@@ -30,12 +30,20 @@
 #include "TopoDS_Builder.hxx"
 #include "TopoDS.hxx"
 #include "TDataStd_Name.hxx"
+#include "TDF_ChildIterator.hxx"
+#include "TopoDS_Iterator.hxx"
 #include "BRepBuilderAPI_Transform.hxx"
-#include "TopExp_Explorer.hxx"
 #include "TopExp_Explorer.hxx"
 #include "BRep_Tool.hxx"
 #include "BRepTools.hxx"
 #include <IMeshTools_Parameters.hxx>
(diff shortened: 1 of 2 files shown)


jomixlaf:
Closing to rework — the fix needs revision before it's ready.

Issue #9478, opened 2026-01-24
Reload from disk not working for step files


phenix1990:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11

### Additional system information
not relevant

### Printer

X1C

### Printer Firmware Version

latest

### How to reproduce

When I import any step file i can use "Reload from disk" as long as Bambu studio keeps open. After I close bambu studio and open again, "Reload from disk" does no more work. step file is within same folder as 3mf file. Error says "Unable to relaoad: <filename>.stp". 
Step file is from lates Alibre Design Proffessional V28. I tried step AP 203 and 242. 

- import step
- close bambu studio
- open bambu studio
- right click on object > reload from disk

### Actual results

Error message "cant reload"

### Expected results

reload updated model form disk

### Project file & Debug log uploads

[Bottom.zip](https://github.com/user-attachments/files/24834566/Bottom.zip)

<img width="259" height="160" alt="Image" src="https://github.com/user-attachments/assets/3d3b0bd4-c72e-422d-ab2e-b8947fd08b11" />

### Checklist of files to include

- [ ] Log file
- [x] Project file

JohnSmithCYM:
Thanks for your feedback and this is due to project file doesn’t record your step file information, if we did that, this may cause your personal info leak.

phenix1990:
I see, so just works for stl even if step file is within the same folder than the 3mf project? 

Is this about kind of meta data in step files?

MrDB42:
> Thanks for your feedback and this is due to project file doesn’t record your step file information, if we did that, this may cause your personal info leak.

Why don't you just use relative file locations?
i.e.
If STEP file is in the same tree location of the 3MF use ".\Sample.STEP"
If STEP file is one level below the location of the 3MF use ".\Step Files\Sample.STEP"
If STEP file is up one level the 3MF "\..\Sample.STEP"
etc.

Or, just fix the open file dialog box to reload the STEP file.

vanowm:
@JohnSmithCYM, and with .stl files you are ok with personal info leak? What kind of twisted answer is this?

Issue #8982, opened 2025-11-30
STEP files do not "Reload from disk"


markeld:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11

### Printer

X1C

### Printer Firmware Version

01.10.00.00

### How to reproduce

Create a project with objects imported from STEP files
Save it to disk
Close and re-open the project
Right click on any object, select "Reload from disk"
You get an error message "Unable to reload: <filename>.step"

### Actual results

<img width="319" height="175" alt="Image" src="https://github.com/user-attachments/assets/8630dfda-e1d1-4120-9a58-b1b3ca4c8a69" />

### Expected results

File should be reloaded with no error message.
Reloading works until you save and open the proejct again.

### Project file & Debug log uploads

happens with any project file

### Checklist of files to include

- [ ] Log file
- [ ] Project file

fahaaaxiki:
Sorry, I couldn't reproduce your issue using my step file. Could you please provide your file?

![Image](https://github.com/user-attachments/assets/d4444830-9e78-467f-8fab-e0ec03df7468)

Issue #1600, opened 2023-04-11
Reload from disk does not work with STP-files loaded as a "single object with multiple parts"


nervnet65:
The feature under Process Object "Reload from disk" leads to the error message "Unable to reload: filepath/filename(s)" when the object was loaded from multiple STP files ("Load these files as a single object with multiple parts? -> Yes). This error message appears both when trying to reload a single STP file or the entire object.

Reload from disk works when an obejct consists of only one single STP file.

If STL files are used, the error does not occur.

BambuStudio 1.5.0.61 unter Windows 10

zhimin-zeng-bambulab:
@nervnet65  Could you share your step files, it will help us to solve this problem.

nervnet65:
@zhimin-zeng-bambulab I have created a simple part and attach it as STL and STP. I was also able to successfully test the described behavior with it.
[object_STL_and_STP.zip](https://github.com/bambulab/BambuStudio/files/11329761/object_STL_and_STP.zip)

BambulabRobot:
This issue has been marked as inactive due to no response for 90 days.

BambulabRobot:
We will close this issue in one week. If you think this issue still needs to be resolved, please reply within this period

PR #10380, opened 2026-04-25
Fix reload-from-disk: matching, positions, deleted bodies, STEP names

jomixlaf:
## Summary

This PR fixes a number of long-standing issues with the **Reload from disk** action when the source file is a STEP exported from CAD (Fusion 360 in particular). It also tightens up the STEP loader's body name extraction so that bodies don't lose their identity on reload.

I believe it would fix most, if not all, of these issues listed here, some I experienced myself. 

[https://github.com/bambulab/BambuStudio/issues/9478](url)
[https://github.com/bambulab/BambuStudio/issues/8982](url)
[https://github.com/bambulab/BambuStudio/issues/1600](url)

All changes live behind the existing `ENABLE_RELOAD_FROM_DISK_REWORK` compile flag for the reload-path edits, plus targeted fixes in the STEP loader.

## STEP loader

- **Resolve `MANIFOLD_SOLID_BREP` body names via the entity transfer map.** Bodies that previously came in as `Body1`/`Body4`/etc. now carry their authored names from the STEP product hierarchy.
- **Use the component name for bodies in root-level assembly components**, fixing cases where a body in a sibling component lost its name and inherited the parent assembly's name instead.
- **Normalize non-ASCII codepoints in body names.** OCCT's `TCollection_AsciiString(extString)` preserved raw high bytes, which then failed `StepPreProcessor::isUtf8` (a lone `0xA0`/`0xE9` is not a valid UTF-8 start). The body would silently fall back to the parent component name. Walking the extended string codepoint-by-codepoint and emitting a benign space for anything outside printable ASCII keeps the user's name intact (e.g. `cylinder<NBSP>[f4]` is no longer rejected).

## Reload-from-disk

- **Reload now picks up new components and name changes** instead of dropping bodies that don't match the original source-index.
- **Prevent source-index mismatch** when the STEP file gains or reorders bodies. Source indices are updated on each reload so the next reload's matching is stable.
- **Refresh the object list sidebar** for each modified object so names, types, and filament columns reflect the reloaded volumes.
- **Three-pass matching** for old → new volume reconciliation:
  1. Source-index + name (unchanged bodies)
  2. Name search (body moved to a different index)
  3. Source-index only (renamed body, same slot)
- **Ungate Pass 2** so the name search runs whenever the volume has a name. The previous gate (`old_volume->name == path.filename()`) was never true for multi-body STEP, so reorders fell through to Pass 3 and matched by index alone, swapping transforms between bodies.
- **Handle deleted bodies on reload** silently (no error dialog) and move `sort_volumes` out of the matching loop so it doesn't corrupt subsequent `vol_idx` values.
- **Correct position of new STEP bodies on reload** via a `coord_shift` captured from the first matched volume — the difference between the initial bb-center and the new bb-center. Applied to unmatched new bodies so they land in the existing scene's coordinate frame.
- **Matched volumes follow Fusion repositioning.** Replaced the unconditional `set_transformation(old_volume->get_transformation())` with applying `coord_shift` to the new volume's offset. For the first matched volume this yields exactly the old offset (preserving the object's plate anchor); for subsequent volumes it tracks any repositioning the user did in CAD between exports while keeping the object anchored.

## Test plan

- [x] Fresh import of a multi-body STEP, then reload the same file unchanged → no movement, all bodies matched.
- [x] Add a new body in Fusion, reload → new body lands at correct position relative to the existing layout.
- [x] Delete a body in Fusion, reload → body silently removed.
- [x] Reorder bodies in Fusion (so name and source-index disagree), reload → all bodies stay correctly matched and positioned.
- [x] Rename bodies between exports → matched by source-index (Pass 3) with offsets preserved.
- [x] Reposition / resize bodies in Fusion (e.g. enlarge a base plate, move a sub-body to a new corner), reload → matched bodies follow new positions, plate anchor stable.
- [x] Body name with non-ASCII characters (`é`, NBSP) → name imports without falling back to the parent assembly name.
- [x] Iterative add → rename → retag → reposition cycles across multiple consecutive reloads stay stable.

### DIFF
--- a/src/libslic3r/Format/STEP.cpp
+++ b/src/libslic3r/Format/STEP.cpp
@@ -30,20 +30,11 @@
 #include "TopoDS_Builder.hxx"
 #include "TopoDS.hxx"
 #include "TDataStd_Name.hxx"
-#include "TDF_ChildIterator.hxx"
-#include "TopoDS_Iterator.hxx"
 #include "BRepBuilderAPI_Transform.hxx"
 #include "TopExp_Explorer.hxx"
 #include "BRep_Tool.hxx"
 #include "BRepTools.hxx"
 #include <IMeshTools_Parameters.hxx>
-#include "StepShape_ManifoldSolidBrep.hxx"
-#include "TransferBRep.hxx"
-#include "XSControl_WorkSession.hxx"
-#include "XSControl_TransferReader.hxx"
-#include "Transfer_TransientProcess.hxx"
-#include "Interface_InterfaceModel.hxx"
-#include <unordered_map>
 
 
 namespace Slic3r {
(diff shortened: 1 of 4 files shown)


MackBambu:
@jomixlaf There is a compilation error.
```
2026-04-25T23:29:25.7016757Z D:\a\BambuStudio\BambuStudio\src\slic3r\GUI\Plater.cpp(10775,39): error C3861: 'default_import_naming_rules': identifier not found [D:\a\BambuStudio\BambuStudio\build\src\slic3r\libslic3r_gui.vcxproj]
2026-04-25T23:29:25.7018601Z   (compiling source file '../../../src/slic3r/GUI/Plater.cpp')
2026-04-25T23:29:25.7094180Z   
2026-04-25T23:29:25.7096504Z D:\a\BambuStudio\BambuStudio\src\slic3r\GUI\Plater.cpp(10787,21): error C3861: 'apply_naming_rules_to_volume': identifier not found [D:\a\BambuStudio\BambuStudio\build\src\slic3r\libslic3r_gui.vcxproj]
2026-04-25T23:29:25.7174499Z   (compiling source file '../../../src/slic3r/GUI/Plater.cpp')
2026-04-25T23:29:25.7175386Z  
```

jomixlaf:
> @jomixlaf There is a compilation error.

My bad, I forgot to come back to let you know that the problem was solved. Also, funny enough, I never get the first email saying the build failed, only the second one. Weird 🤷🏻‍♂️ Have a great weekend!

Haidiye00:
Hello, the internal version is currently upgrading the code related to the step, so please pause merging this patch for a few days. Sorry for the inconvenience, please be patient.

jomixlaf:
Thanks! I had more improvement for this PR in the oven, so I'll wait to see what's new and improved 👍

jomixlaf:
Hi, Just checking if the code has or will be updated for the step portion. 

Thanks

PR #9705, opened 2026-02-08
Add Traditional Chinese (zh_TW) translation

hezhen0816:
Hello, thank you for the great work on BambuStudio!

I noticed that there is currently no Traditional Chinese (zh_TW) translation available. As a user from Taiwan, I would love to contribute a zh_TW translation to help more Traditional Chinese users.

- Converted from `zh_CN` using OpenCC `s2twp` (Simplified to Traditional with Taiwan standard phrases)
- Contains 4,997 translated entries covering the full UI
- Taiwan-specific phrasing is used (e.g. 列印 instead of 打印, 支援 instead of 支持, 噴嘴 instead of 喷嘴)

I hope this translation can be helpful to the community. Please let me know if any changes are needed. Thank you!

PR #10386, opened 2026-04-26
Add Traditional Chinese language entry

hezhen0816:
This adds Traditional Chinese to the supported language list so the existing zh_TW translation can appear in the language selector.

The zh_TW translation files were added in #9705, and this follows up on the maintainer note that the language entry would be added later.

Testing:
- Not run locally; macOS build is currently blocked by local dependency build issues unrelated to this one-line language entry change.


tonghao-bbl:
LGTM @lanewei120

hezhen0816:
Thanks. I have a zh_TW translation update ready as well, and I will submit it as a separate follow-up PR after this one is merged.

lanewei120:
@hezhen0816 thanks for doing this

hezhen0816:
@lanewei120 You're welcome. It is my honor to help make this software more accessible to more users.

Issue #10289, opened 2026-04-19
Systemic UI slowness: entire application sluggish even with no project open — rooted in ~37 wxTimer instances and busy wxEVT_IDLE loops (static code evidence included; consolidates #3702, #9413, #9967, #1950)


L0rdS474n:
# Systemic UI slowness: entire application sluggish even with no project open — rooted in ~37 wxTimer instances and busy wxEVT_IDLE loops (static code evidence included; consolidates #3702, #9413, #9967, #1950)

---

## Bambu Studio Version

2.6.0.51

## Where is the application from?

Bambu Lab Official website

## OS version

Windows 11 Pro, version 25H2 (OS-version 26200.8246)

## Additional system information

CPU: Intel Core i7-12800HX (12th gen Alder Lake-HX, 16 cores / 24 threads)
Memory: 64 GB
Display Card (dGPU): NVIDIA RTX A2000 Laptop GPU, 8 GB VRAM — driver 32.0.15.9166
                     (Ampere-generation workstation card; two architectural generations behind current Ada Lovelace / Blackwell)
Display Card (iGPU): Intel UHD Graphics (driver 32.0.101.7084)

System is a mobile workstation — well in excess of Bambu Studio's stated requirements.

The same UI slowness has been independently reproduced on a second, physically separate PC equipped with a newer-generation NVIDIA dGPU. The problem is therefore not hardware-specific to one machine or one NVIDIA generation.

## Printer

Bambu Lab H2D and Bambu Lab X1C

## Printer Firmware Version

N/A — issue reproduces before and without any printer connection.

## How to reproduce

1. Launch Bambu Studio from the Start menu (no project file opened).
2. Wait for the splash / initial load to complete and the main window to appear.
3. Click any of the main tabs (Prepare, Preview, Project, Calibration, Device, Cloud, ...).
4. Observe that UI responsiveness is severely degraded globally — mouse clicks register with multi-second delay, tab switches take several seconds, scrolling and hover feedback are sluggish.
5. Leave the application idle on the Prepare tab with no project loaded and no printer connected. Observe that CPU and page-fault activity for `bambu-studio.exe` remain elevated indefinitely.
6. (Optional diagnostic) In Task Manager → Details tab → right-click column headers → add "Page Faults" and "Page Fault Delta" columns. You will typically see `bambu-studio.exe` generating thousands of page faults per second while idle (same signature as #3702).

## Actual results

[...]

PR #10387, opened 2026-04-26
fix: remove duplicate CMake sources, plug timer leak, debounce idle config save

tinezivic:
Fixes three small independent issues from the static analysis in #10289.

**F-010 — duplicate sources in `CMakeLists.txt`**
`Scrollbar.cpp` and `ScrolledWindow.cpp` were listed twice in `SLIC3R_GUI_SOURCES`. Removed the redundant entries.

**F-006 — `timer_sync_printer` leak in `Sidebar::priv::~priv()`**
`timer_sync_printer` is heap-allocated (`= new wxTimer()`) but the destructor never deleted it. Added `Stop()` + `delete` + null-out.

**F-015 — unbounded `app_config->save()` in idle handler**
The idle handler called `app_config->save()` on every dirty idle event — on a busy loop this means a synchronous disk write at idle frequency. Added a 5-second rate limit via `std::chrono::steady_clock`. `OnExit()` flushes any remaining dirty state so no settings are lost on clean shutdown.

---

**Tests**

Extracted the rate-limit logic into `src/libslic3r/Debounce.hpp` (pure C++, no wx dependency) and added `tests/libslic3r/test_debounce.cpp` with 5 Catch2 scenarios covering first call, suppression, interval expiry, zero interval, and no mutation on suppressed calls.

```
All tests passed (7 assertions in 5 test cases)
```

Build verified on Linux/GCC: `cmake --build build --target BambuStudio -j$(nproc)` completes without new warnings or errors.

Ref: #10289

### DIFF
--- a/src/slic3r/CMakeLists.txt
+++ b/src/slic3r/CMakeLists.txt
@@ -70,10 +70,6 @@ set(SLIC3R_GUI_SOURCES
     GUI/Widgets/FilamentLoad.hpp
     GUI/Widgets/FanControl.cpp
     GUI/Widgets/FanControl.hpp
-    GUI/Widgets/Scrollbar.cpp
-    GUI/Widgets/Scrollbar.hpp
-    GUI/Widgets/ScrolledWindow.cpp
-    GUI/Widgets/ScrolledWindow.hpp
     GUI/Widgets/StepCtrl.cpp
     GUI/Widgets/StepCtrl.hpp
     GUI/Widgets/ProgressBar.cpp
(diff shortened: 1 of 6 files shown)


src/slic3r/CMakeLists.txt
```diff
@@ -70,10 +70,6 @@ set(SLIC3R_GUI_SOURCES
     GUI/Widgets/FilamentLoad.hpp
     GUI/Widgets/FanControl.cpp
     GUI/Widgets/FanControl.hpp
-    GUI/Widgets/Scrollbar.cpp
-    GUI/Widgets/Scrollbar.hpp
-    GUI/Widgets/ScrolledWindow.cpp
```
weishibambu, line 75: Why these files were removed from cmake file?

tinezivic: Those files are listed twice — the first entries are above, around line 39.

lanewei120:
hello @tinezivic 

could you resolve the conflicts firstly

PR #10451, opened 2026-04-29
ENH: Smart import naming rules — auto-assign filament & part type from body name tags

jomixlaf:
## Summary

Adds a small naming-rules system that parses tags inside body/object names at import time (and on reload) to auto-assign **filament** and **part type**. Tagging a body in your CAD as `wheel[f3][mod]` is enough to land it in BambuStudio with filament 3 and type Modifier — no clicking through the type submenu, no remembering which body needs which filament after every reload.

Works for STEP, 3MF, OBJ, and STL — every multi-body format BambuStudio supports.

This builds on top of [PR #10380](https://github.com/bambulab/BambuStudio/pull/10380) (reload-from-disk fixes). The reload path here calls into the naming-rules helpers added by this PR, so #10380 should land first.

## Tag syntax

`body_name[tag1][tag2]…` — tags are case-insensitive, order-independent, combinable.

| Tag | Meaning |
|---|---|
| `[f1]` … `[f16]` | Filament 1–16 |
| `[part]` | Type: Part (default) |
| `[neg]` | Type: Negative Part |
| `[mod]` | Type: Modifier |
| `[blk]` | Type: Support Blocker |
| `[enf]` | Type: Support Enforcer |

Examples:
- `gear[f2]` → Part on filament 2
- `cavity[neg]` → Negative Part
- `support_zone[enf][f5]` → Support Enforcer on filament 5

**Why brackets?** They survive every CAD export I tested, are ASCII-safe, path-safe, readable, and don't collide with normal naming.

## Behavior

- **On import** (STEP / 3MF / OBJ / STL): every volume's name is parsed; matched tags set filament and/or type.
- **On reload from disk**: rules re-applied to matched and new bodies, so renaming or retagging in CAD is reflected immediately.
- **Tag removal reverts to default**: removing a tag in CAD on the next export resets that property (filament → 1, type → Part). If the name has no brackets at all, the volume is left alone — manual filament/type tweaks survive reload.
- **Always overridable** by the user from the sidebar after import.

## Files added

- `src/libslic3r/ImportNamingRules.hpp` / `.cpp` — rule definitions, tag parser, and `apply_naming_rules_to_volume` / `apply_naming_rules_to_objects` helpers.

## Files touched

- `src/slic3r/GUI/Plater.cpp` — calls `apply_naming_rules_to_objects` after import; calls `apply_naming_rules_to_volume` on matched + new volumes during reload.
- STL import path — naming rules now applied.

## Test plan

- [x] STEP import with `[f1]`…`[f16]` tags → filament column populated correctly
- [x] STEP import with `[neg]` / `[mod]` / `[blk]` / `[enf]` → volume type set
- [x] Combined tags like `[f3][mod]` → both applied
- [x] Reload from disk picks up tag changes (add, remove, change tag)
- [x] Removing a tag in CAD between exports reverts the property (e.g. `[mod]` → no tag → back to Part)
- [x] Names without any `[...]` left alone — manual filament adjustments survive reload
- [x] Tags work on STL imports
- [x] 3MF and OBJ imports apply rules (verified in code path; pair with sample multi-body files)
- [x] Bodies with non-ASCII chars in name (when paired with #10380) keep their tags intact


MackBambu:
I will review #10380  first.

ArthurBambulab (commented):
Review: Smart import naming rules (#10451)

Nice feature idea — auto-assigning filament/type from CAD body tags like [f3][mod] is genuinely useful. However, there are a couple of data-loss risks and a large unrelated reload rewrite bundled in. Requesting changes before merge.

P0 — Blocking

1. load_model_objects applies naming rules unconditionally — clobbers saved settings when loading 3MF project files (src/slic3r/GUI/Plater.cpp:8973)

load_model_objects is the shared entry point for all load paths, including 3MF project loads (it's called with is_project_file at the project-load site). This PR runs apply_naming_rules_to_objects({object}, naming_rules) unconditionally inside the per-object loop.

As a result, if any body name in a saved project happens to contain [...] (even unrelated brackets like [v2] or [copy]), apply_naming_rules_to_volume fires and apply_import_name_tags force-resets the extruder to 1 and the volume type to MODEL_PART, silently overwriting the per-volume filament/type the user saved in the project. A user reopening an already-configured multi-color / modifier project can lose those settings.

Naming rules should only apply to raw mesh imports (STEP/STL/OBJ/single-mesh model import), never to project-file loads. Please move the call out of load_model_objects, or gate it with a flag that the project-load path sets to false.

2. "Revert to default" semantics wipe legitimate settings when brackets are unrelated (src/libslic3r/ImportNamingRules.cpp:183)

apply_naming_rules_to_volume triggers whenever the name contains any bracket pair, even when no recognized tag matched (e.g. gear[rev3], bracket[v2]). It then forces filament→1 and type→PART. The implementation cannot distinguish "user removed a tag" from "the name simply contains unrelated brackets," so it misfires on legitimate names. Combined with #1, this causes erroneous resets in project/reload scenarios.

Suggestion: only apply when parse_import_name_tags matches at least one known rule; otherwise leave the volume untouched.

P1 — Important

3. Debug logging left in production code (src/slic3r/GUI/Plater.cpp:10580-10800)

reload_from_disk now contains at least 6 blocks of BOOST_LOG_TRIVIAL(info) << "[RELOAD_DEBUG] ..." printing per-volume offsets. These are clearly development leftovers — they pollute production logs and add per-object overhead on large models. Please remove them all before merge (or drop to trace and remove the [RELOAD_DEBUG] prefix).

4. PR bundles an unrelated, high-risk reload rewrite (src/slic3r/GUI/Plater.cpp:10625+)

Beyond naming rules, this PR rewrites the reload matching/coordinate/deletion logic: 3-pass matching (source-idx, name, source-idx-only), a new coord_shift translation, deleted_vol_idxs, and moving sort_volumes out of the loop. This is an independent, high-risk behavioral change (coordinate-frame translation, plus silently dropping bodies deleted from the STEP file instead of adding them to fail_list) that is a separate concern from "smart naming."

The PR also states it depends on #10380, and the branch is currently mergeable_state: dirty (conflicts with master).

Please split the PR: keep naming rules (ImportNamingRules + STEP name normalization + STL extension stripping + apply-on-import) as one PR, and submit the reload rewrite separately with proper regression coverage. As-is, both review and rollback are difficult.

5. Pass-2 object-name match sets new_object_idx without match_found (src/slic3r/GUI/Plater.cpp:10677-10690)

When obj->name == old_volume->name matches, the code breaks but leaves match_found = false and new_volume_idx = -1. Subsequent logic then depends on an ambiguous combination of new_object_idx >= 0 and new_volume_idx. This is hard to follow and bug-prone. Please make the intended state explicit (set the match flag and volume idx), or add a comment explaining why object-level matching with volume_idx == -1 is valid here.

P2 — Medium

6. Regex recompiled on every call (src/libslic3r/ImportNamingRules.cpp:163)

parse_import_name_tags declares std::regex tag_re(...) as a local, so it recompiles for every volume. Large assemblies (hundreds/thousands of bodies) pay a noticeable cost. Make it static const std::regex (as bracket_re already is in apply_naming_rules_to_volume).

7. Conflicting type tags silently resolved as "last wins," undocumented (src/libslic3r/ImportNamingRules.cpp:130-134)

Multiple conflicting type tags (e.g. [neg][mod]) silently take the last one. This is acceptable as a design choice, but please document it in the header. Also default_import_naming_rules() rebuilds the whole table on every import — consider making it static const.

8. STL extension stripped only in the object_name_in == nullptr branch (src/libslic3r/Format/STL.cpp:34-44)

When a caller passes object_name_in (with extension), the .stl suffix remains during tag parsing. It doesn't break [tag] matching itself, but it's inconsistent with the STEP path. Please confirm all STL call paths or strip the extension uniformly before parsing.

9. PR description vs. implementation mismatch (OBJ/3MF)

The description claims import applies to "STEP / 3MF / OBJ / STL," but OBJ/3MF only get rules via the generic load_model_objects apply — which is exactly the source of P0-1 (project 3MF gets clobbered). After fixing P0, please clarify that model 3MF import and project 3MF load must be treated differently.

jomixlaf:
Hi @ArthurBambulab I am waiting for another PR to be resolved by Bambu Team before working on this one again. [https://github.com/bambulab/BambuStudio/pull/10380](url)

Once this is done on their side I will look at those. Thanks :)

PR #10547, opened 2026-05-03
FIX: avoid debug asserts in multifilament project loading

altendky:
## Summary
- Fix a debug-build assertion when loading by-object multifilament projects that use filament slot 1.
- Treat `PrintObject::object_extruders()` values as 0-based filament indexes in by-object multifilament validation.

## Details
`PrintObject::object_extruders()` returns 0-based filament indexes, so `0` is valid and represents the first filament. The by-object branch in `Print::check_multi_filament_valid()` incorrectly asserted that each returned extruder index was greater than zero, causing debug builds to abort while loading valid multifilament `.3mf` projects.

The fix removes that invalid assertion and uses an unsigned local set matching the returned index type.

## Testing
- `git diff --check master...HEAD`

## Notes
- No regression test was added; the existing FFF print test target is currently stale and fails to build on unrelated API drift.


lanewei120:
thanks

PR #10585, opened 2026-05-06
Clean Czech translation update

jakubhencl:
This PR contains only the Czech translation update in a single commit.

Previous PR included unrelated history/commits, so this clean PR replaces it.


lanewei120:
thanks

PR #10667, opened 2026-05-11
Fix: add distance threshold to reduce_infill_retraction optimization (#3423)

### LINKED ISSUE #3423: CRITICAL: Fails on Z retractions and in the starting g-code process.
### Bambu Studio Version

1.8.4.51

Bambu Lab Official website

### OS version

macOS Sonoma 14.2.1

Apple M1 Macbook

### Printer

Bambu Lab A1 series

### How to reproduce

Printing a G-code generated with Bambu Studio with a non planar superior surface.
I attached an example.

### Actual results

The nozzle scratch over piece.
Check the movement between 43684 and 43685, there is no a Z retraction.

### Expected results

The nozzle should be retract in Z axis between movements in the same layer and in the same type of line. Always.

[debug_Fri_Jan_19_08_40_03_43376.log.0.zip](https://github.com/bambulab/BambuStudio/files/13989230/debug_Fri_Jan_19_08_40_03_43376.log.0.zip)
[scratching example.3mf.zip](https://github.com/bambulab/BambuStudio/files/13989232/scratching.example.3mf.zip)

- [X] Log file
- [X] Project file

tinezivic:
## Problem

The infill retraction bypass in `needs_retraction()` (GCode.cpp) skips retraction for **any** travel move that stays inside an infill region without crossing a perimeter wall — regardless of travel distance.

On small prints this is fine. On larger parts a single travel move can span 30–50 mm across infill with no retraction, causing:
- Stringing across infill surfaces
- Nozzle drag / scraping on infill
- Support knock-offs (the core issue reported in #3423)

This affects BambuStudio, OrcaSlicer, and PrusaSlicer equally — none of them have a distance threshold on this optimization.

## Root cause

```cpp
// GCode.cpp — needs_retraction()
    m_config.sparse_infill_density.value > 0 &&
    m_retract_when_crossing_perimeters.travel_inside_internal_regions_no_wall_crossing(*m_layer, travel))
    return false;  // no distance check — skips retraction for ANY length
```

## Fix

Reuse the existing `retraction_minimum_travel` threshold (already used at the top of `needs_retraction()`) for the infill bypass as well:

```cpp
    m_config.sparse_infill_density.value > 0 &&
    travel.length() < scale_(FILAMENT_CONFIG(retraction_minimum_travel)) &&
    m_retract_when_crossing_perimeters.travel_inside_internal_regions_no_wall_crossing(*m_layer, travel))
    return false;
```

- Short hops (< `retraction_minimum_travel`): unretracted — no speed regression
- Long moves (≥ `retraction_minimum_travel`): retracted — eliminates stringing and drag

No new settings needed. `retraction_minimum_travel` already has the correct semantics and is already exposed in the UI.

## Testing

The current workaround (disabling *Reduce infill retractions*) confirms the issue is in this code path — enabling it with this fix should produce equivalent quality to disabling it, while preserving the speed benefit for short infill hops.

Closes #3423

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -7083,9 +7083,11 @@ bool GCode::needs_retraction(const Polyline &travel, ExtrusionRole role, LiftTyp
         }
         // rirDisabled: should_reduce remains false
         if (should_reduce && !is_perimeter(role) && m_layer != nullptr && m_config.sparse_infill_density.value > 0 &&
+            travel.length() < scale_(FILAMENT_CONFIG(retraction_minimum_travel)) &&
             m_retract_when_crossing_perimeters.travel_inside_internal_regions_no_wall_crossing(*m_layer, travel))
             // Skip retraction if travel is contained in an internal slice *and*
-            // internal infill is enabled (so that stringing is entirely not visible).
+            // internal infill is enabled (so that stringing is entirely not visible) *and*
+            // travel distance is shorter than retraction_minimum_travel (to prevent oozing on long moves).
             //FIXME any_internal_region_slice_contains() is potentionally very slow, it shall test for the bounding boxes first.
             return false;
     }


Diatom-Bambu:
Thank you for your interest and contribution. However, your proposed change effectively disables the "reduce infill retraction" feature permanently at the code level—regardless of whether the setting is toggled on or off—which is not the behavior we intend. Acknowledging that print quality can indeed suffer with this feature enabled for certain materials, we have introduced a `filament_metal_stickiness` parameter within the filament properties. Consequently, when using the "reduce infill retraction" mode (Auto setting), materials prone to nozzle adhesion—such as PETG—will continue to perform retraction movements.
While this may address your needs to some extent, we do not want the "reduce infill retraction" feature to be governed solely by a minimal distance threshold (specifically `retraction_minimum_travel`, which is typically just a few millimeters), as that would be tantamount to removing the feature entirely.
If your goal is simply to disable the "reduce infill retraction" feature completely, there is no need to modify the code; you can simply turn it off within the process preset.
image

PR #10693, opened 2026-05-13
Modernize build flow with Pixi (conda-forge)

jijinbei:
## What

Replace the legacy `BuildLinux.sh` / `BuildMac.sh` / `build_win.bat` + `deps/` ExternalProject flow with a [Pixi](https://pixi.sh) (conda-forge) based build. Single command bootstrap on all three OSes:
```bash
pixi run bootstrap   # builds bambulab forks of libnoise + wxWidgets
pixi run build       # cmake configure + parallel build
```
CI (`.github/workflows/build_pixi.yml`) builds the three platforms in a matrix and uploads the Linux AppImage as an artifact.

### Diff size

| Scope                                          | Files |   +added |  -removed |
| ---------------------------------------------- | ----: | -------: | --------: |
| Total (headline)                               |   226 | +11,586  | -139,058  |
| Excluding `pixi.lock` (auto-generated)         |   225 |  +2,660  | -139,058  |
| Above + excluding `deps/` + `diff_summary.txt` |    98 |  +2,604  |   -4,354  |

The headline −139k / +12k LoC is easy to misread:

- ~129k of the deletions are vendored upstream source under `deps/` (Boost, CGAL, OCCT, GLEW, EXPAT and 22 others). They are not BambuStudio's own code — they were bundled copies of other projects, now supplied by conda-forge.
- ~8.9k of the additions are the auto-generated `pixi.lock`.
- ~5.6k of the deletions are `diff_summary.txt`, an upstream-tracked auto-generated diff log (now removed).
- **The hand-written change is ~+2.6k / -4.4k LoC across 98 files (net −1,750)**, with new build infra under `scripts/pixi/` + `CMakePresets.json` plus deletion of legacy build scripts (`BuildLinux.sh`, `BuildMac.sh`, `BuildFedora.sh`, `build_win.bat`, `Build.PL`, the Dockerfile flow, `deps-*.cmake`) and three legacy CI workflows (`build_bambu.yml`, `build_deps.yml`, `build_ubuntu.yml`).

## Why

FreeCAD recently switched its builds to Pixi and saw a noticeable jump in contributor velocity; their `sub_buildPixi.yml` is the direct reference for this branch's CI. Same pattern adapted here:

- All deps from conda-forge, with `pixi.lock` as the source of truth
- Same commands on Linux / macOS / Windows
- Distributable artifacts in the same flow (~~AppImage today, `.app` / DMG next, Windows portable zip after~~ → AppImage / DMG / MSI all built today)

## Verified on

- **Linux**: Ubuntu 26.04 LTS (local) + `ubuntu-24.04` (CI, glibc ≥ 2.34)
- **Windows**: Windows 11 (local) + `windows-2022` (CI)
- **macOS**: M1 / Apple Silicon (local) + `macos-14` (CI)

## Feedback wanted

Posting as draft to invite hands-on feedback from core contributors.
Particularly interested in:

- Whether `pixi run bootstrap && pixi run build` runs cleanly on your  daily-driver OS (build time and disk footprint in particular)
- Any slicing / STEP I/O / GUI regressions from the dep version bumps (Boost 1.75 → 1.88, CGAL 5 → 6, OCCT 7.6 → 8, OpenCV 4.6 → 4.13)
- Whether the "no native-host build path" trade-off (see below) is acceptable
- How the pixi env fits your usual development workflow (IDE integration, debugger, resource hot-reload, etc.)

Issue comments or PR review are both welcome.

## Status

The overall direction is settled, but the finer details are still being iterated on (~~wiring the Windows portable zip and macOS `.app` into CI~~ done, ~~removing the legacy workflow files~~ done, designing a tag-triggered release, polishing error messages and bootstrap scripts, etc.). Posting as draft now to collect feedback on the direction itself early.

See the "Updates since posting" comment below for per-area progress.

### Native-host compatibility dropped (intentional)

The original plan kept the option of building outside the pixi env, but the resulting `CMakeLists.txt` would have needed branches in every `find_package` / target-link spot to handle both the conda-forge layout
and the distro layout. This bloats the file and makes both paths fragile. The branch goes all-in on the pixi env; if a `BuildLinux.sh`- style flow is needed again it should be reintroduced as a thin wrapper, not as CMake-level conditionals.

### Source modernization required

conda-forge ships newer deps than the vendored `deps/` did (Boost 1.88 vs 1.75, CGAL 6 vs 5, OCCT 8 vs 7.6, OpenCV 4.13 vs 4.6). Inevitable small C++ fixes are included:

- `boost::process` v1/v2 split (Boost 1.86+) — pull v1 sub-headers
- `io_service` → `io_context`, `branch_path()` → `parent_path()`, `copy_option::overwrite_if_exists` → `copy_options::overwrite_existing`
- CGAL 6 `Surface_mesh::property_map()` now returns `std::optional`
- OCCT 8 `TKXDESTEP` / `TKSTEP` consolidation → use `TKDESTEP`
- conda-forge `libopencv` ships modular targets (link `opencv_core` etc. instead of `opencv_world`)
- wxString → utf8 conversion before `json::parse` (the json lib drops bytes otherwise on wide-string platforms)

### WebView2 from NuGet, not vendored

The `deps/WebView2` Microsoft SDK isn't on conda-forge, so the vendored tree was removed and `scripts/pixi/windows/setup-webview2.ps1` downloads the same NuGet package (pinned to the previously vendored `1.0.1418.22`) into `$CONDA_PREFIX/Library/{include,lib,bin}` on first bootstrap. Contributors no longer have to commit binaries; the trade-off is that offline first-time bootstrap requires internet for this step.

### Build performance choices

- **sccache** as the C/C++ compiler launcher via `CMAKE_<LANG>_COMPILER_LAUNCHER` in `CMakePresets.json`. CI (`build_pixi.yml`) uses sccache's built-in GitHub Actions cache backend (`SCCACHE_GHA_ENABLED=true`, via `mozilla-actions/sccache-action`), so no separate `actions/cache` wrapper is needed. The PoC originally used ccache; sccache was picked because it transparently supports Windows / MSVC (cl.exe) and shares one cache layer across the three OSes.
- **mold** as the linker on Linux (~13× faster relink in local measurements; the BambuStudio binary links ~50 static archives, so link time was the dominant cost).
- **Ninja** generator on Linux / macOS (Visual Studio generator stays on Windows for toolchain alignment).
- **Configure-skip** in `scripts/pixi/build.sh` when the CMake cache is still valid (~40× faster incremental rebuilds).

These are swappable; the choice reflects "what worked best on the three matrix OSes for this PoC", not a final recommendation.


jijinbei:
## Updates since posting

### 1. Distributables: AppImage / DMG / MSI all green
All three produced by `pixi run dist` and uploaded by `build_pixi.yml`. The Windows MSI dropped 84 lines of manual `_payload.wxs` generation in the WiX 7 upgrade (`f04e6ff14`).

### 2. Windows defaults to RelWithDebInfo (`4782fae91`)
conda-forge ships Release-only deps on Windows. Pure Debug would need ~50 lines of hacks (TBB pragma suppression, wxWidgets dep `*d.lib → *.lib` rewriting, building wxWidgets twice). RelWithDebInfo gets the same `/INCREMENTAL` link speed and pdb-driven debugging with zero hacks. `pixi run build` defaults to it; ship builds use `build-release`.

### 3. Cleanup
- `scripts/pixi/` reshaped into `shared / linux / macos / windows` (`aa11e7b7d`)
- Legacy CI workflows removed (`665adac2c`) — `build_pixi.yml` is the sole workflow now
- Dead-code purge (`5e7491e4c`, `02d689316`, `1ca9a27cb`): unreferenced `deps/Assimp/Assimp.cmake`, `cmake/modules/{bin2h, MacOSXBundleInfo.plist.in}`, plus the `SLIC3R_PERL_XS / BUILD_TESTS / BUILD_SANDBOXES` options + Perl-XS / Catch2 helpers — pixi flow never flipped them on. Followed upstream's own comment: `FIXME Port the tests into C++`.

### 4. Misc build/runtime fixes
- wxWidgets sys-lib hard-pin (`3f332366a`): `wx_add_thirdparty_library` silently fell back to vendored libpng-1.4 when `find_package(PNG)` missed, breaking `wxImage::LoadFile` at runtime on macos-14 runners. Hard-pin `PNG_LIBRARY` / `JPEG_LIBRARY` / `TIFF_LIBRARY` / `ZLIB_LIBRARY` / `EXPAT_LIBRARY` to the conda env so the `sys` path always wins, and patch-rm the vendored sources so the silent fallback can no longer compile.
- `FindGTK3.cmake` restored (`b2ac99e2b`) — an earlier cleanup deleted it but it's still referenced indirectly via wxWidgets find rules.
- macOS dock icon as PNG (`09926957d`, `32640e980`) — was loading a PNG embedded in a `.ICO` container; now reads the PNG directly and skips `dock SetIcon` when `wxIcon` load fails instead of asserting.

### 5. CI tweaks
- Windows OCCT junction recreated on cache hit (`acef124c7`) — `actions/cache` restores junctions as empty dirs, which broke `bambustudio_copy_dlls()`'s OCCT lookup. Re-run setup-occt-bridge on cache hit (idempotent, ~instant).
- mp3lame / x265 import-lib aliases cached + `_jobs.ps1` path fix (`1c2d6495f`).
- `build_pixi.yml` path triggers extended to `packaging/appimage/**` + macOS wx cache bust (`17684823b`).
- Verbose / dead config trim (`2d16fa512`, −38 lines): three upload-artifact steps consolidated into one matrix-driven step, unused Install bubblewrap step dropped.
- Deprecated `save-always` removed from Windows bootstrap cache (`c2a25fa1a`).
- Stray root-level build artifacts cleaned (`b0d717409`).

jijinbei:
## Tested builds from GitHub Actions

Verified that the artifacts generated by CI work correctly.

image
Screenshot From 2026-05-21 22-35-17

jijinbei:
## AppImage launch reliability

conda-forge libs (webkit2gtk `LIBEXECDIR`, libcurl `OPENSSLDIR`, fontconfig defaults, DT_RPATH) bake the build-host `$CONDA_PREFIX` as absolute strings and ignore env-var overrides, so the Linux AppImage failed at startup on user machines with Fontconfig / libcurl / WebKitNetworkProcess errors. Fixed by wrapping `bambu-studio` in bwrap and bind-mounting `AppDir/usr` onto the recorded build prefix (`79346b03f`), with a `$ORIGIN`-relative RPATH (`a48f3739c`) + patchelf rewrite (`a90c3d456`) as the supporting layer. A late fix tmpfs's the missing build-prefix ancestor so `--bind` can `mkdir` the parent dirs on machines where `/home/runner/...` doesn't exist (`767949a21`).

Verified locally on Ubuntu 26.04 with `apparmor_restrict_unprivileged_userns=1`.

## Known platform gaps

Out of scope for this PR; follow-ups welcome.

- **Intel Mac (x86_64) — not yet supported.** 
- **Arch-family Linux — not yet exercised.**

jijinbei:
Build instructions are documented in [`doc/How to build - Pixi.md`](https://github.com/jijinbei/BambuStudio/blob/modernize/pixi-poc/doc/How%20to%20build%20-%20Pixi.md).

Quick start:
1. Install [Pixi](https://pixi.sh)
2. `pixi install`
3. `pixi run bootstrap`
4. `pixi run build`

jijinbei:
Looking back, I'm wondering whether a phased migration would have been better here.

lanewei120:
Hello @jijinbei

Thanks a lot for the effort put into this work. 
We appreciate the goal of making the build environment easier to reproduce and lowering the barrier for contributors. Pixi seems to provide a nice developer experience, especially for first-time contributors.

That said, at the moment we would prefer to keep our existing build infrastructure as the primary and officially supported path. For release builds and long-term maintenance, we tend to favor solutions with fewer external runtime dependencies and a workflow that remains compatible with our current CI, packaging, and internal environments.

For this reason, we are not planning to migrate to a Pixi-only build system in the near future. We would be more comfortable treating Pixi as an optional or experimental build method alongside the existing build scripts, rather than replacing them entirely.

Thanks again for the contribution and for exploring this direction. We believe reproducible environments are valuable, but we'd like to approach such a migration incrementally and evaluate it over time.

PR #10712, opened 2026-05-14
FIX: Add Linux ARM64 support to new nodejs dep

hadess:


src/slic3r/GUI/DeviceWeb/CMakeLists.txt
```diff
@@ -44,22 +44,37 @@ function(prepare_node_env)
...
-        set(NODE_URL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz")
-        set(NODE_ARCHIVE_NAME "node-v${NODE_VERSION}-linux-x64.tar.xz")
-        set(NODE_BIN_DIR "node-v${NODE_VERSION}-linux-x64/bin")
-        set(NODE_EXEC "node")
-        set(NODE_HASH "88fd1ce767091fd8d4a99fdb2356e98c819f93f3b1f8663853a2dee9b438068a")
-
-        set(PNPM_URL "https://github.com/pnpm/pnpm/releases/download/v${PNPM_VERSION}/pnpm-linux-x64")
-        set(PNPM_EXEC_NAME "pnpm")
-        set(PNPM_HASH "eb2dc1f109bca046ce734d062c8dd8f34db2b58a115992f9b086456efd7b2305")
+        if(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
```
hadess: I haven't yet been able to verify whether this is correct.

BenJule:
Heads-up: both failures here come from `master`, not from your nodejs change.

1. **Ubuntu/Windows `Build BambuStudio`** fails with `'class MachineObject' has no member named 'get_extruder_id_by_ams_id'` (AMSMaterialsSetting.cpp:1168). This was already fixed on `master` (it now uses `GetFilaSystem()->GetExtruderIdByAmsId`), so a **rebase onto current master** clears it.

2. **macOS `Build Deps`** fails while building `dep_Assimp` — its bundled `contrib/zlib` no longer parses against the macOS 15+/26 SDK. A one-line fix is proposed in #11051; once it lands, a re-run goes green.

So: rebase + #11051 and this should build clean.

src/slic3r/GUI/DeviceWeb/CMakeLists.txt
hadess: It's actually `aarch64`.

JingTanbbl (approved):

src/slic3r/GUI/DeviceWeb/CMakeLists.txt
```diff
@@ -208,11 +223,17 @@ set(DEVICE_PAGE_BUILD_INPUTS
 
 add_custom_command(
     OUTPUT "${DEVICE_PAGE_BUILD_STAMP}"
-    COMMAND ${CMAKE_COMMAND} -E env "PATH=${DEVICE_PAGE_PATH}" "PNPM_STORE_DIR=${PNPM_STORE_DIR}" ${PNPM_EXECUTABLE} install
-    COMMAND ${CMAKE_COMMAND} -E env "PATH=${DEVICE_PAGE_PATH}" "PNPM_STORE_DIR=${PNPM_STORE_DIR}" ${PNPM_EXECUTABLE} run build
+    COMMAND echo "Running PNPM install"
+    COMMAND ${CMAKE_COMMAND} -E env "PATH=${DEVICE_PAGE_PATH}" ${PNPM_EXECUTABLE} install --store-dir ${PNPM_STORE_DIR}
```
JingTanbbl, line 227: someone mentioned the --store-dir CLI argument has a bug on pnpm 11.x + Windows where it gets ignored：[https://github.com/pnpm/pnpm/issues/12013](https://github.com/pnpm/pnpm/issues/12013). Consider keeping the original approach, or adding a comment near the version number to remind future maintainers.

hadess: The original approach? Setting an environment variable does nothing, which means that it completely ignores the store dir.

In this case, you should probably add a comment above the pnpm downloads earlier in the code if some versions are broken, seeing as this `CMakeLists.txt` the ones downloading pnpm, making broken versions easy to avoid.

JingTanbbl (approved):

src/slic3r/GUI/DeviceWeb/CMakeLists.txt
JingTanbbl: you are right

JingTanbbl: Please squash these four commits into one — it would help us a lot.

hadess: You can do that by selecting the "squash" option in the merge type drop-down.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github

PR #10717, opened 2026-05-14
Fix macOS build with CMake 4.x and Xcode 16+

karlingen:
## Problem

Building BambuStudio on macOS fails with **CMake ≥ 4.0** and **Xcode 16+** (AppleClang 18+) due to several issues:

1. **CMake 4.x removed compatibility with `cmake_minimum_required` < 3.5**, causing multiple dependency sub-builds (Boost, OpenSSL, wxWidgets cotire, etc.) to fail during configuration.

2. **Assimp's bundled zlib** defines `#define fdopen(fd,mode) NULL` in `zutil.h`, which conflicts with the macOS SDK's `fdopen` declaration in `_stdio.h` starting with Xcode 16, producing:
   ```
   error: expected identifier or '('
   FILE    *fdopen(int, const char *) ...
   ```

3. **Clipper2's `-Werror`** causes the build to fail because the parent `CMakeLists.txt` adds `-Wno-error=enum-constexpr-conversion` (line 286), which is an unrecognized warning option on some AppleClang versions. Combined with Clipper2's `-Werror`, this triggers `-Werror,-Wunknown-warning-option`.

4. **`constexpr` enum casts** in `wxMediaCtrl2.h` and `MediaPlayCtrl.h` fail with newer Clang, which enforces stricter rules for `constexpr` with unscoped enum types:
   ```
   error: constexpr variable 'MEDIASTATE_BUFFERING' must be initialized by a constant expression
   ```

5. **`AMSMaterialsSetting.cpp`** calls `obj->get_extruder_id_by_ams_id()`, a method that does not exist on `MachineObject`. This is a code bug that fails on any compiler.

### `deps/deps-macos.cmake`
Add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to `DEP_CMAKE_OPTS` so it propagates to all `ExternalProject_Add` sub-builds via `bambustudio_add_cmake_project`.

> **Note:** Builders using CMake 4.x also need to set `CMAKE_POLICY_VERSION_MINIMUM=3.5` as an **environment variable** during the build to cover `try_compile()` calls (e.g., wxWidgets' cotire test). This should be documented in the wiki.

### `deps/Assimp/Assimp.cmake`
Change `ASSIMP_BUILD_ZLIB` from `ON` to `OFF`. This uses the system zlib instead of Assimp's bundled copy, which is the standard approach on macOS where system zlib is always available.

### `src/clipper2/CMakeLists.txt`
Add `-Wno-unknown-warning-option` alongside `-Werror` for non-Windows builds. This prevents the build from failing when the parent project adds warning flags that the current compiler version doesn't recognize.

### `src/slic3r/GUI/wxMediaCtrl2.h` and `src/slic3r/GUI/MediaPlayCtrl.h`
Change `static constexpr wxMediaState` to `static const wxMediaState` for custom media state constants. Casting integer values to an unscoped enum in a `constexpr` context is rejected by Clang 18+ (Xcode 16+). Using `const` preserves identical runtime behavior.

### `src/slic3r/GUI/AMSMaterialsSetting.cpp`
Replace call to non-existent `obj->get_extruder_id_by_ams_id()` with `obj->GetFilaSystem()->GetExtruderIdByAmsId()`, which is the correct API used elsewhere in the same file (e.g., lines 597 and 1025).

## Build environment tested

- **macOS**: arm64 (Apple Silicon)
- **Xcode**: 16.4.1
- **CMake**: 4.3.2 (Homebrew)
- **Deployment target**: 10.15

## Wiki update suggestion

The [Mac Compile Guide](https://github.com/bambulab/BambuStudio/wiki/Mac-Compile-Guide) should note that **CMake 4.x** users need to prefix build commands with:
```bash
export CMAKE_POLICY_VERSION_MINIMUM=3.5
```
or use CMake 3.x until upstream dependencies update their `cmake_minimum_required()` declarations.


adele-with-a-b:
+1 — independently hit the same five failures while building locally on macOS (Apple Silicon, Apple Clang 21, macOS SDK 26.5, CMake 4.x). Each fix in this PR matches what I ended up doing in my fork to get a green build:

- **Assimp bundled zlib** — `-DASSIMP_BUILD_ZLIB=OFF` in `deps/Assimp/Assimp.cmake`. Same `zutil.h:147` `fdopen → NULL` collision against `_stdio.h:318`. Confirmed needed.
- **CMake 4.x `cmake_minimum_required` < 3.5** — needed `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to get past `FindOpenVDB.cmake:126` and several other vendored sub-builds.
- **Clipper2 `-Werror` + unknown-warning-option** — without `-Wno-unknown-warning-option`, Apple Clang 21 errored on the parent's `-Wno-error=enum-constexpr-conversion` (the warning name was dropped upstream of clang 21).
- **`wxMediaState constexpr → const`** — same `error: constexpr variable 'MEDIASTATE_BUFFERING' must be initialized by a constant expression` against newer Clang.
- **`AMSMaterialsSetting` non-existent method** — confirmed `obj->GetFilaSystem()->GetExtruderIdByAmsId(...)` is the correct call (also used at `AMSMaterialsSetting.cpp:597, 1025` and across `AMSDryControl`, `CalibrationWizardPresetPage`).

Worth bumping for review — every macOS-touching open PR's CI is currently red on (1) alone, including unrelated ones like #10712 (Linux ARM64 nodejs) and my own #10767 (LAN MQTT recovery).

PR #10746, opened 2026-05-16
fix: replace non-existent get_extruder_id_by_ams_id with existing GetFilaSystem()->GetExtruderIdByAmsId

### LINKED ISSUE #10745: AMSMaterialsSetting.cpp:1168 calls undefined method get_extruder_id_by_ams_id
## Build Error

Compilation fails with:

```
src/slic3r/GUI/AMSMaterialsSetting.cpp:1168: error: 'class Slic3r::MachineObject' has no member named 'get_extruder_id_by_ams_id'
```

## Root Cause

Line 1168 of `AMSMaterialsSetting.cpp` calls a non-existent method directly on `MachineObject`:

```cpp
int ext_id = obj->get_extruder_id_by_ams_id(std::to_string(ams_id));
```

This method was never defined anywhere in the codebase — not on `MachineObject` in `DeviceManager.hpp` or anywhere else.

## Evidence

The same file already uses the correct equivalent elsewhere:

- **Line 597**: `obj->GetFilaSystem()->GetExtruderIdByAmsId(std::to_string(ams_id))`
- **Line 1025**: `obj->GetFilaSystem()->GetExtruderIdByAmsId(std::to_string(ams_id))`

The method `GetExtruderIdByAmsId` is defined on `DevFilaSystem` (`DeviceCore/DevFilaSystem.h:288`, `.cpp:411`) and returns `int`. It correctly looks up the extruder ID from the AMS list, or returns `MAIN_EXTRUDER_ID` / `DEPUTY_EXTRUDER_ID` for virtual trays, or `-1` if not found.

## Fix

Replace line 1168:

```diff
- int ext_id = obj->get_extruder_id_by_ams_id(std::to_string(ams_id));
+ int ext_id = obj->GetFilaSystem()->GetExtruderIdByAmsId(std::to_string(ams_id));
```

It appears a convenience wrapper `get_extruder_id_by_ams_id` was intended for `MachineObject` but the header change was never committed.

kooscode:
## Summary
- Fixes compilation error caused by calling a non-existent method `get_extruder_id_by_ams_id` on `MachineObject`
- Replaces it with the correct equivalent `GetFilaSystem()->GetExtruderIdByAmsId()` which already exists in `DevFilaSystem` and is used elsewhere in the same file (lines 597, 1025)

## Fix
```diff
- int ext_id = obj->get_extruder_id_by_ams_id(std::to_string(ams_id));
+ int ext_id = obj->GetFilaSystem()->GetExtruderIdByAmsId(std::to_string(ams_id));
```

## Related
Fixes #10745


XinZhangBambu:
Thanks for the contribution, @kooscode! 🙏

This compilation issue has already been fixed, so this change is no longer needed and we'll be closing the PR. We really appreciate you catching it and taking the time to submit a fix — thanks again for helping improve Bambu Studio!

PR #10767, opened 2026-05-18
fix(devmgr): auto-reconnect stale LAN MQTT and recover startup race

### LINKED ISSUE #9445: Bambu Studio completely forgets LAN only printer between sessions
### Bambu Studio Version

2.4.0.70

Bambu Lab Official website

### OS version

Windows 11 and NixOS

_No response_

### Printer

Bambu Lab P2S

### Printer Firmware Version

01.00.05.00

### How to reproduce

1. P2S in Lan Only Mode with developer mode enabled  
2. Open Bambu Studio
3. Go to the "Device" tab
4. Click "Connect with access code"
5. Enter Printer IP and access code => Printer connects and can be used
6. Close Bambu Studio
7. Open Bambu Studio

Now, there is no device under the "Device" tab and the whole procedure has to be done again.

### Actual results

I can connect to the printer, but the connection is not remembered after closing Bambu Studio. Upon launching it again, the printer has to be bound again from the start by entering its IP and access code.

### Expected results

When re-opening Bambu Studio on the same machine after the printer was bound before, Bambu Studio should remember the printer - rebinding it every time should not be necessary.

I've tested this on a desktop PC and a Laptop on Win 11, NixOS (native), and NixOS (flatpack). The issue persists everywhere.

- [ ] Log file
- [ ] Project file

adele-with-a-b:
## Summary

LAN-only users (no Bambu cloud account) routinely hit two reconnection failures that leave the printer effectively unusable until the user manually clicks the printer in the Devices tab. This PR fixes both with a single bundled commit, since they share the same code surface (`DeviceManager` + `TryLoadLastMachine`), affect the same users, reproduce on the same workflow, and interact at runtime.

Closes #9445 (still actively reported in May 2026 on macOS — see [comment from harmstorf](https://github.com/bambulab/BambuStudio/issues/9445#issuecomment-4471441305)).

## Failure 1 — Stale MQTT socket after idle

**Repro:** Open BambuStudio, confirm the printer is connected. Walk away for 30+ minutes. Come back, click Send to Printer. Result: `BAMBU_NETWORK_ERR_SEND_MSG_FAILED` (`-4`).

**Root cause:** macOS App Nap, the network stack, or the printer-side keepalive can silently tear down the MQTT-over-TLS socket while the app is idle. The next `publish_gcode()` writes to a dead socket. The 1Hz `DeviceManagerRefresher::on_timer` exists to keep MQTT alive but its `check_pushing()` / `refresh_connection()` calls are gated on `is_user_login()`. For LAN-only users that flag is never set true, so keep-alive never runs.

**Fix:** Add a parallel branch in `on_timer` that detects the stale-LAN-MQTT condition (`is_lan_mode_printer && has_access_right && is_studio_active && !is_connected`) and, when all four hold, re-selects the same machine ID. That triggers the same-id-LAN branch in `set_selected_machine` (`DevManager.cpp:478-492`), which runs `disconnect_printer → reset → connect` — the exact path the manual workaround takes. Throttled to one attempt per 10 s to avoid hammering on persistent failures (powered-off printer).

## Failure 2 — Stale `user_access_dev_ip` on cold start

**Repro:** Restart BambuStudio. Result: printer doesn't show as selected, even though it's on the network and was bound before.

**Root cause:** `TryLoadLastMachine::InnerLoad` runs at startup (right after `agent->start()`) and tries `bind_detect(decoded_user_access_dev_ip, ...)`. If the cached IP is stale — slicer_uuid was rotated since pairing so the encoded IP can't decode, OR the printer's DHCP IP changed — `bind_detect` fails immediately with `-2`, erases `user_access_dev_ip`, and bails. The cloud fallback then runs but also fails because the LAN printer isn't in `localMachineList` yet. SSDP populates the local list 1-3 seconds later, but no further `InnerLoad` retry happens. The printer sits in `localMachineList` but is never selected.

**Fix:** Add `GUI_App::try_load_last_machine_on_alive(dev_id)` and call it from `DeviceManager::on_machine_alive` whenever an SSDP packet announces a previously-paired printer. The method self-filters on `dev_id == get_user_last_machine()` and no-ops if a machine is already selected, so calling it on every SSDP packet is cheap. By the time SSDP arrives, the printer is in `localMachineList`, so the second `InnerLoad`'s `set_selected_machine` succeeds via the same-id-LAN branch.

## Test plan

**Stale-MQTT recovery:**
1. Pair a LAN-only printer, confirm connection in Devices tab.
2. Idle the app for 30+ minutes (lock screen, close laptop lid).
3. Wake the app. Click Send to Printer.
4. **Without patch:** returns `-4`. **With patch:** the next 1Hz refresher tick auto-reconnects (within ~1 s) and the send succeeds.
5. Verify log line `LAN auto-reconnect: stale MQTT socket detected for dev_id=...`.

**Startup-race recovery:**
1. Pair a LAN-only printer.
2. Either rotate `slicer_uuid` in `BambuStudio.conf` to invalidate the encoded `user_access_dev_ip`, OR power-cycle the router so SSDP is delayed at next launch.
3. Restart BambuStudio.
4. **Without patch:** printer is in the device list (post-SSDP) but never selected. **With patch:** the SSDP packet triggers `InnerLoad` retry; the printer is selected automatically.
5. Verify log lines `try_load_last_machine_on_alive: SSDP-triggered retry for ...` and `set_selected_machine: select new lan machine ...`.

## Limitations

After studio inactivity (>15 min), the first user action after wake may still see one failed send before stale-MQTT auto-reconnect runs on the next refresher tick. A subsequent send within 1 s succeeds.

## Tested on

- macOS 26.4.1, Apple Silicon, against a Bambu H2C in LAN-only mode (developer mode enabled).
- Both fixes confirmed via session log analysis on a real reproduction.

### DIFF
--- a/src/slic3r/GUI/DeviceCore/DevManager.cpp
+++ b/src/slic3r/GUI/DeviceCore/DevManager.cpp
@@ -282,6 +282,33 @@ namespace Slic3r
                 obj->erase_user_access_code();
                 obj->erase_user_access_dev_ip();
             }
+
+            // SSDP-driven retry of last-machine restore. The startup
+            // TryLoadLastMachine::InnerLoad path is racy for LAN-only users:
+            // if the cached user_access_dev_ip is stale (slicer_uuid rotated
+            // since pairing, or the printer is at a new IP) the initial
+            // bind_detect fails before SSDP can announce the printer's
+            // current IP, and the printer ends up in localMachineList but
+            // never selected.
+            //
+            // Recovery path: by the time SSDP populates localMachineList, the
+            // failed initial bind_detect has erased user_access_dev_ip. The
+            // retry's InnerLoad finds the encoded IP empty and falls through
+            // to the dev->get_my_machine(...) non-null branch (GUI_App.cpp
+            // around line 8429), which calls set_selected_machine directly.
+            // No second bind_detect is spawned.
+            //
+            // try_load_last_machine_on_alive runs on the wx UI thread (this
+            // function is dispatched via CallAfter from the SSDP listener),
+            // same thread as DeviceManagerRefresher::on_timer, so there is
+            // no concurrent invocation of set_selected_machine across the
+            // two LAN-recovery code paths in this commit.
+            //
+            // Called for every SSDP announcement (~5s/printer), not just
+            // first discovery. Cheap: try_load_last_machine_on_alive
+            // self-filters on dev_id == get_user_last_machine() and no-ops
+            // if a machine is already selected.
+            Slic3r::GUI::wxGetApp().try_load_last_machine_on_alive(dev_id);
         }
         catch (...) {
             ;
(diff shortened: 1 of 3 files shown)


adele-with-a-b:
The macOS / Linux / Windows CI failures here aren't caused by the changes in this PR. The diff under `deps/` is empty:

```
$ git diff --stat upstream/master..fix/lan-stale-mqtt
 src/slic3r/GUI/DeviceCore/DevManager.cpp | 90 +++++++++++++++++++++++++++++++
 src/slic3r/GUI/GUI_App.cpp               | 26 +++++++++
 src/slic3r/GUI/GUI_App.hpp               | 13 +++++
```

The failure is in `deps/build/.../dep_Assimp-prefix/src/dep_Assimp/contrib/zlib/zutil.c`, where `zutil.h:147` defines `fdopen(fd,mode) → NULL` and the substituted `NULL` collides with macOS SDK's `fdopen` declaration in `<_stdio.h>:318` (`__DARWIN_ALIAS_STARTING(...)` doesn't tolerate a macro-expanded parameter name). This is reproducible across every open macOS-touching PR — see for example #10712's CI, which fails the exact same way on completely unrelated code.

The umbrella fix is already filed as #10717 (`-DASSIMP_BUILD_ZLIB=OFF` plus four other related issues). I hit the same compile error locally (Apple Clang 21 + macOS SDK 26) and the `ASSIMP_BUILD_ZLIB=OFF` workaround landed cleanly — same fix from Karlingen's PR.

The changes in this PR are confined to `DeviceManager` and `TryLoadLastMachine`, so I'd expect them to build green once #10717 (or the equivalent) lands and CI is unblocked.

adele-with-a-b:
Update on the CI failures: the Linux failures (`Build BambuStudio` step on ubuntu-22.04 / 24.04) are also unrelated to this PR's diff. They fail at `AMSMaterialsSetting.cpp:1168` with `‘class Slic3r::MachineObject’ has no member named ‘get_extruder_id_by_ams_id’` — the same regression that's covered by item 5 of #10717 and is the focus of #10768. Easy to confirm by reading the same line in this PR's diff (it's not there — PR diff is 3 files in `src/slic3r/GUI/`, none of them `AMSMaterialsSetting.cpp`):

```
$ git diff --stat upstream/master..fix/lan-stale-mqtt
 src/slic3r/GUI/DeviceCore/DevManager.cpp | 90 ++++++++++++++++++++++++++++++++
 src/slic3r/GUI/GUI_App.cpp               | 26 +++++++++
 src/slic3r/GUI/GUI_App.hpp               | 13 +++++
```

The same error appears on every open PR's Linux build for the past 2+ weeks — see #10712's CI for an unrelated example. So the full CI picture for this PR's diff is: macOS and Windows red on the Assimp/zlib `fdopen → NULL` collision, Linux red on the `get_extruder_id_by_ams_id` regression, both upstream-wide. The PR's actual code in `DeviceManager` and `TryLoadLastMachine` should compile green once #10717 (or #10768) lands and CI is unblocked.

BenJule:
Hi @adele-with-a-b! The macOS CI failures in this PR are caused by a pre-existing infrastructure issue unrelated to your changes.

**Root cause:** Assimp 5.4.3's bundled zlib defines `#define fdopen(fd,mode) NULL` in `contrib/zlib/zutil.h:147`. On macOS 15+ the SDK's `_stdio.h` expands `fdopen` through `__DARWIN_ALIAS(fdopen)`, which hits the `NULL` macro and produces a parse error during the "Build Deps" step.

I've submitted a fix in #10849 that disables the bundled zlib on macOS (system zlib is always available there) while keeping it enabled on Windows. Once that merges, this PR should pass CI cleanly.

XinZhangBambu:
@adele-with-a-b Thanks for your contribution! We really appreciate the effort!  But just to let you know — we’ve already addressed this issue on our side, and the fix will be included in an upcoming release.

adele-with-a-b:
@XinZhangBambu Thanks for the heads up!

1. Is there a tracking handle (internal ticket reference, JIRA STUDIO-XXXXX, target version like 02.07.00.x, or anything I can grep against future commits) you can share so I can track the upcoming fix and avoid carrying my local patch longer than necessary?

2. Does the upstream fix cover both failure modes the PR addresses — (a) the post-idle stale-MQTT `-4` symptom AND (b) the cold-start `user_access_dev_ip` race where the printer doesn't show as selected after restart? Or only one of them?

3. Meta question for future PRs: is there a public signal I should be checking before opening a PR that would tell me "this is already being addressed internally"? I checked the obvious channels first (issue assignment, labels, milestones, commit cross-refs in `upstream/master` and `upstream/release/*`) and didn't find a way to know about an internal-but-not-yet-released fix. Happy to defer issues to internal triage if there's a signal I'm missing — I just don't want to keep submitting parallel PRs that you've already solved.

If both failure modes will be fixed in the upcoming release, happy to close this PR as duplicate as soon as I can confirm the fix shape. If only one, scoping this down to the remaining one is also fine. In the meantime I'm carrying it as a local patch on top of my fork, so no urgency on my end — just trying to keep the public record clean and build a better pre-PR check process for future contributions.

XinZhangBambu:
> **[adele-with-a-b](/adele-with-a-b) ** c

Hi @adele-with-a-b，

Thanks for the detailed follow-up.

The internal tracking tag for this issue is STUDIO‑18013, and the corresponding fix has already landed in our mainline.
You can reference the commit here:

https://github.com/bambulab/BambuStudio/commit/c3d42ce015267f9482542da0bcbfd8aa5f512bed

This internal fix covers the same problem space you mentioned, so your PR can safely rely on this upstream change.

Thanks again for keeping the public record clean and for the thoughtful questions — much appreciated.

adele-with-a-b:
Closing as duplicate of internal STUDIO-18013, which has now landed upstream as commit [c3d42ce015](https://github.com/bambulab/BambuStudio/commit/c3d42ce015267f9482542da0bcbfd8aa5f512bed) ("FIX: restore cached LAN printers"). Confirmed by @XinZhangBambu in [the comment above](https://github.com/bambulab/BambuStudio/pull/10767#issuecomment-3170218927).

Verified locally: the upstream commit is now an ancestor of `upstream/master` (`git merge-base --is-ancestor c3d42ce015 upstream/master` → 0).

Thanks @XinZhangBambu for the heads-up and tracking handle, and @BenJule for confirming the unrelated CI failure on Assimp/zlib — that infra issue was separately addressed by #10849.

Closing this PR. Local fork branch will be deleted; the work is preserved as an immortal git tag (`archive/fix_lan-stale-mqtt-20260608` on adele-with-a-b/BambuStudio) for any future reference.

PR #10794, opened 2026-05-19
Color Cutting, Dovetail, Non-planar cutting, sculpting, with color preservation

adamgasoft:
This pull request includes a colorcut addin, that does color preservation in both planar and dovetail, a warpcut test addin that allows a user to define an arbitrary cutting plane and adjust it, with color preservation, and a ballerase test addin that allows sculpting using a ball/sphere or a cube shape, resizable, that also preserves color... 

these addins were seperated from the core bambustudio addins so they could be tested independently.

https://github.com/adamgasoft/AGBStudioPublic has a released setup application so you can test all the functionality, as well as documentation (see root pdf)


Haidiye00:
Hello, there are too many changes to the document, the translation file does not need to be modified. Please submit the GIFs before and after the changes. Thank you.

lanewei120:
also there are lots of modifications to the preset files, no need in this patch

PR #10795, opened 2026-05-19
ENH: auto-insert all known LAN printers into device list on startup

Sebski123:
Previously, only the last-selected machine was reconnected at startup. Now, all LAN-only printers stored in the config are inserted into the device list.

The last-selected machine still receives the active connection. Other inserted machines are visible but idle until selected.

### DIFF
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -4520,6 +4520,7 @@ void GUI_App::request_user_logout()
     if (m_agent && m_agent->is_user_login()) {
         m_load_last_machine.is_list_ok = false;
         m_load_last_machine.is_mqtt_ok = false;
+        m_load_last_machine.launched_dev_ids.clear();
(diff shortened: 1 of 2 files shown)


XinZhangBambu:
@Sebski123 Thanks for your contribution! We really appreciate the effort. But just to let you know — we’ve already addressed this issue on our side, and the fix will be included in an upcoming release.

Sebski123:
> @Sebski123 Thanks for your contribution! We really appreciate the effort. But just to let you know — we’ve already addressed this issue on our side, and the fix will be included in an upcoming release.

Perfect, as long as it's added I'm happy

XinZhangBambu:
Hi @Sebski123 

Thanks for the follow-up.

The internal tracking tag for this issue is STUDIO‑18013, and the corresponding fix has already landed in our mainline.
You can reference the commit here:

https://github.com/bambulab/BambuStudio/commit/c3d42ce015267f9482542da0bcbfd8aa5f512bed

Issue #10165, opened 2026-04-08
Please add the ability to read makerworld comments


nixit302-spec:
It would be really nice to be able to view other users comments from Bambu studio like we can do in the handy app or website. There is valuable information in there and all that is available is the 1-5 Star ratings. Forcing us to open the model in the app or website before we print just to make sure there are no gotchas on recommendations.

Otherwise love the program. Works great and looks amazing. Keep up the good work!

Nick Mitchell
TheRealNickMitchell
nick@nickmitchell.org

Issue #9321, opened 2026-01-08
Render hyperlinks in Makerworld descriptions


dwballance:
**Is your feature request related to a problem? Please describe.**
When I open a Makerworld model in BambuStudio, hyperlinks embedded in the description by the author do not render. 

**Describe the solution you'd like**
I would like to see the hyperlink with an alternate font handling (blue, underline, etc) and when clicking on it, either launch a web browser to that link or navigate to the linked model in BS.

**Describe alternatives you've considered**
The only current alternative is to open the MW model page directly in a web browser so that the hyperlinks can be accessed. It would be useful to have an "open in browser" button - this way other non-rendered content could also be more easily accessed (user comments, STL download option).

**Additional context**
This would particularly help when a maker references another model that supports or extends the current model.

dwballance:
I discovered that if I look at the "Project info" tab of a project that I have loaded into the slicer (versus looking at the details in the Makerworld browser in the Home tab), I do see they hyperlinks. So that is a partial solution. But it would be great to see the links in the browser as well.

Issue #5755, opened 2025-01-11
When Viewing MakerWorld: Comments are hidden and can't copy links


JonLaliberte:
**Is your feature request related to a problem? Please describe.**
I find using MakerWorld via the app to often be not useful. I depend on seeing the comments others have left to decide if the model is going to work for what I need.

There is also no easy way to copy a link to the model, nor open it in my standard browser.
So, I end up having to instead go to the site and search for the model again.

**Describe the solution you'd like**
Show comments.
Let us copy a link to the model on MakerWorld somehow (right click, button, whatever).

**Describe alternatives you've considered**
Not using the app for viewing models at all.

Issue #5076, opened 2024-10-25
When browsing makerworld in Bambu Studio, I'd like to be able to click to the designer's other designs


twelch24:
**Is your feature request related to a problem? Please describe.**
Not really a problem but an inconvenience for sure. 

**Describe the solution you'd like**
When browsing maker world on the Bambu Handy or Bambu Studio app, I can see the designer's name. There's no simple way (that I see) to click on their name or otherwise see their other designs. I'd like to make their name a hyperlink or something similar to be able to see their other designs. This works when browsing the maker world website, but not in handy or studio. 

**Describe alternatives you've considered**
Using the website, i suppose. 

**Additional context**
<img width="1190" alt="bambustudio" src="https://github.com/user-attachments/assets/cc9f7188-ab65-48f5-a969-8f24fa677014">

SketchyGuySketches:
This is infuriating when browsing uploaded content. The inability to support hyperlinks in the text descriptions is also poor user experience.

PR #10798, opened 2026-05-20
feat(webview): translate Online Models URLs to public web + add Share menu

adele-with-a-b:
## Summary

Two related improvements to the Online Models tab, bundled because they share the same code surface (`WebViewPanel` toolbar + URL handling) and cooperate at runtime:

1. **URL translation:** the existing toolbar "Open in browser" button used to launch the slicer-embedded URL (`/studio/webview?modelid=N`), which serves a stripped-down view designed for the in-app iframe — comments hidden, designer profile hidden, hyperlinks in descriptions not rendered, version selector abbreviated. Translate that to the public-web URL (`/<lang>/models/N`) so the user lands on the actual MakerWorld page they expect.

2. **Share menu:** replace the single-action "Open in browser" toolbar button with a Share button (Bambu green) that pops a 3-item menu — `Open Link`, `Copy Link`, `Share...`. The previous open-in-browser flow is preserved as the menu's first item; copy and system-share are now reachable from the same surface. `Share...` opens the OS-native share UI (`NSSharingServicePicker` on macOS, `DataTransferManager` on Windows) so the user can send the link via Messages, Mail, AirDrop, Notes, or any installed share extension (Slack, Telegram, etc.). On Linux (and on Windows builds without the Win10 SDK) the menu falls back to Open Link / Copy Link only.

Addresses #5755 (comments hidden, can't copy links) and helps #10165 (read makerworld comments), #9321 (render hyperlinks), #5076 (related discoverability gaps).

## Architecture

The cross-platform interface lives in `src/slic3r/GUI/SystemShare.{hpp,cpp}` plus three platform implementations selected by CMake:

- `SystemShare_mac.mm` — `NSSharingServicePicker` anchored to the wxWindow's NSView via `wxWindow::GetHandle()`.
- `SystemShare_win.cpp` — `Windows.ApplicationModel.DataTransfer.DataTransferManager` driven via `IDataTransferManagerInterop` so a Win32 HWND can talk to what's normally a UWP-only API. Activation pattern follows the existing `Utils/FixModelByWin10.cpp`: dynamic-load `ComBase.dll` and look up `RoInitialize` / `RoGetActivationFactory` / `WindowsCreateString` at runtime via `GetProcAddress`. Gated on `HAS_WIN10SDK`, same gate `FixModelByWin10.cpp` uses.
- `SystemShare_fallback.cpp` — Linux, plus Windows builds without `HAS_WIN10SDK`. `HasNativePicker()` returns false; the menu omits the `Share...` item.

`HasNativePicker()` on Windows additionally consults the lazy WinRT loader so older Windows installs lacking ComBase.dll (or the specific symbols we depend on) hide the menu item entirely instead of advertising a feature that would silently no-op when invoked.

`add_DataRequested` is additive on the per-HWND `DataTransferManager` singleton, so the handler must be unregistered explicitly. We capture the cookie returned by `add_DataRequested` via `shared_ptr<EventRegistrationToken>` into the handler's closure and the handler removes itself on first fire; explicit cleanup at the call site if `ShowShareUIForWindow` returns an error before invoking the data callback. macOS picker is autoreleased (project default is MRR, no `-fobjc-arc`) so each Share... invocation doesn't permanently retain a picker.

## Test plan

**URL translation:**
1. Open Online Models tab.
2. Click any model card.
3. Click the new toolbar Share button → choose Open Link.
4. **Without patch:** browser opens `/studio/webview?modelid=N` (slicer-embedded view, comments hidden). **With patch:** browser opens `/<lang>/models/N` (real MakerWorld page).

**Share menu:**
1. Open Online Models tab; click any model card.
2. Click the toolbar Share button.
3. Verify menu items: `Open Link`, `Copy Link`, separator, `Share...` (macOS / Windows only).
4. Choose `Copy Link` → paste anywhere; verify the URL is the public `/<lang>/models/N` form.
5. (macOS / Windows) Choose `Share...` → verify the system share picker appears anchored at the toolbar button, listing Messages, Mail, AirDrop, etc.

## Tested

- macOS 26.4.1, Apple Silicon (M5 Max), Xcode 26.5 SDK. Open Link, Copy Link, and Share... (NSSharingServicePicker) all verified end-to-end.
- Windows: implementation written against the WinRT docs and the existing `Utils/FixModelByWin10.cpp` activation pattern in this codebase. Compiles per Bambu Windows CI; runtime-untested by author (no Windows test machine). Happy to iterate on feedback from a maintainer or Windows-using contributor.
- Linux: fallback path (Open Link / Copy Link only) compiles; the macOS-tested URL translation logic is platform-independent so the Linux user-visible behavior is the same as macOS minus the `Share...` item.

### DIFF
--- a/resources/images/share.svg
+++ b/resources/images/share.svg
@@ -1,0 +1,14 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <!-- 3-dots-and-lines share glyph, OS-agnostic. Sized to visually
+       match the weight and extent of the refresh icon next to it in
+       the online-models toolbar. -->
+
+  <!-- Connector lines: drawn first so the dots overpaint the joins. -->
+  <path d="M3.2 8 L12.8 13.7 M12.8 2.3 L3.2 8"
+        stroke="#00AE42" stroke-width="1.9" stroke-linecap="round"/>
+
+  <!-- Three dots, radius 2.7, edges of the box. -->
+  <circle cx="13.3" cy="2.3"  r="2.7" fill="#00AE42"/>
+  <circle cx="2.7"  cy="8"    r="2.7" fill="#00AE42"/>
+  <circle cx="13.3" cy="13.7" r="2.7" fill="#00AE42"/>
+</svg>
--- a/resources/images/share_copy_link.svg
+++ b/resources/images/share_copy_link.svg
@@ -1,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect x="4.5" y="4.5" width="8" height="9.5" rx="1.5" stroke="#262E30" stroke-width="1.2" fill="none"/>
+<path d="M3.5 11.5h-0.5a1.5 1.5 0 0 1-1.5-1.5v-7a1.5 1.5 0 0 1 1.5-1.5h6a1.5 1.5 0 0 1 1.5 1.5v0.5" stroke="#262E30" stroke-width="1.2" fill="none"/>
+</svg>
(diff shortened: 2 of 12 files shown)


ubidefeo:
I have been wanting this feature for soooo long 🥲

adele-with-a-b:
@lanewei120 gentle nudge on this one — it's been open a while and I don't think it's reached anyone's review queue yet.

Current state: green across all seven build targets, including the macOS jobs that were failing on the unrelated Assimp deps issue earlier in the summer, and GitHub reports it as mergeable against current `master`.

Is there anything needed from me to move it along? If the scope is the awkward part, I'm happy to split it into two smaller PRs — the Online Models URL translation and the Share menu are independent of each other and could be reviewed separately.

PR #10824, opened 2026-05-20
fix: persist GCode view type (color scheme) across sessions

### LINKED ISSUE #7131: Color Scheme now defaults to FILAMENT (used to be LINE TYPE) -> PUT THIS IN SETTINGS?!?!
### Bambu Studio Version

2.1.0.5.9

Bambu Lab Official website

### OS version

MacOS Somona

M1Ultra

### Printer

N/A

### How to reproduce

Slice any part

### Actual results

Whatever I slice, the default color scheme is anything but LINE STYLE like it used to be.

### Expected results

Default to LINE STYLE *OR* make it user selectable on settings menu

N/A

- [x] Log file
- [x] Project file

BenJule:
## Summary

The selected GCode color scheme (Line Type, Filament, Feedrate, etc.) was not remembered across sessions. The save/restore code existed but was gated behind an `enable_record_gcodeviewer_option_item` config flag that defaults to `false`.

Removed the guard so the last selection is always saved to `gcodeviewer_option_item` in app config and restored on next startup.

## Test plan

- [ ] Slice a model
- [ ] Switch the GCode view color scheme to "Filament" (or any non-default)
- [ ] Restart BambuStudio
- [ ] Open the same or any other sliced model → "Filament" color scheme should still be selected

Closes #7131

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
      - master
-     - main
     paths:
      - 'deps/**'
      - 'src/**'
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -3,6 +3,7 @@ name: Build all
 on:
   push:
     branches:
+     - master
      - main
     paths:
      - 'deps/**'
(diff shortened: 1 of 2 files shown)


Haidiye00:
Hello, this change is meaningless. Just check it in the preference settings UI, thank you.
image

PR #10825, opened 2026-05-20
fix: add transparent/clear color to filament color picker presets

### LINKED ISSUE #10574: No "clear" option in filament color.
I have clear TPU. When I loaded it into the filament manger, ther is no option for "clear". I would like you to update the color selection to include a "clear" color option. Also, would be nice to have "glow" and "translucent" as "check boxes" when selecting the filament color. 

Thank you!

BenJule:
## Summary

Users with clear or transparent filament had no way to select that color in the filament color picker, there was no transparent swatch in the preset list.

Added a fully-transparent entry (alpha=0, `wxColour(0, 0, 0, 0)`) as the first preset swatch in `ColorPickerPopup`. The existing checkered transparent bitmap rendering in `ColorPicker::doRender` and the alpha=0 handling in `set_def_colour` already support this case correctly.

## Test plan

- [ ] Open the filament color picker for any slot
- [ ] The first preset swatch should show a checkered transparent pattern
- [ ] Clicking it should set the filament color to transparent/clear

Closes #10574

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
      - master
-     - main
     paths:
      - 'deps/**'
      - 'src/**'
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -3,6 +3,7 @@ name: Build all
 on:
   push:
     branches:
+     - master
      - main
     paths:
      - 'deps/**'
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -2505,7 +2505,9 @@ void ColorPicker::doRender(wxDC& dc)
 
     auto draw_state = [&]() {
         if (m_selected) {
-            dc.SetPen(wxPen(m_colour));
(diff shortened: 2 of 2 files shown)


BenJule:

XinZhangBambu:
> Thanks for the contribution. I tested the current implementation, but the UX feels unclear: the transparent option looks too similar to a regular color swatch and its placement as the first preset is not intuitive. Could we give it a clear checkerboard/transparent icon with a tooltip such as “Transparent / Clear”, and visually separate it from the normal colors?
image

BenJule:
Thanks for the feedback. I’ve updated the transparent/clear color option accordingly:

- moved it into a dedicated section, separate from the regular color presets
- kept the existing checkerboard transparency icon
- added a “Transparent / Clear” tooltip
- added a visible selection outline for the transparent option

The PR has also been cleaned up to contain only this change as a single signed commit. I verified the updated source with a targeted C++ syntax check.

Please take another look when you have a chance. Thanks!

XinZhangBambu:
Thanks for the contribution, and for reworking it based on the feedback

Before I take another look, could you share a **screenshot or short video of how it works now** — Seeing the updated UX in action would make it much easier to confirm. Thanks again!

PR #10833, opened 2026-05-20
fix(macos): Cmd+H hide-app shortcut blocked by missing evt.Skip()

### LINKED ISSUE #10740: Cannot Hide Bambu Studio desktop on MacOS using keyboard shortcut
### Bambu Studio Version

2.6.0.51

Bambu Lab Official website

### OS version

MacOS 15.7.4

Apple Silicon M2, 16GB

### Printer

Bambu Lab X1C

### How to reproduce

Press Command + h on keyboard to hide application
nothing happens

### Actual results

Nothing

### Expected results

Application should be hidden.  

Have to use mouse and click Bambu Studio in menu bar and drag down to "Hide Bambu Studio" option

No project file related

- [ ] Log file
- [ ] Project file

BenJule:
## Problem

On macOS, pressing **Cmd+H** to hide BambuStudio has no effect — users have to go to the menu bar and click "Hide BambuStudio" manually (#10740).

## Root cause

`MainFrame.cpp` has a `wxEVT_CHAR_HOOK` handler that special-cases Cmd+H on macOS with a comment "call parent_menu hide behavior". But the implementation just calls `return` without first calling `evt.Skip()`. In wxWidgets, returning from a `wxEVT_CHAR_HOOK` handler without `Skip()` **consumes** the event — macOS never sees it, so the native "Hide Application" behaviour never fires.

```cpp
// Before (broken):
if (evt.CmdDown() && (evt.GetKeyCode() == 'H')) {
    //call parent_menu hide behavior
    return;}   // ← consumes the event, macOS never sees Cmd+H
```

## Fix

Call `evt.Skip()` before returning, so the event propagates through wxWidgets to the macOS application menu where the standard hide action lives.

```cpp
// After:
if (evt.CmdDown() && (evt.GetKeyCode() == 'H')) {
    evt.Skip();  // pass to macOS "Hide BambuStudio"
    return;
}
```

## Test plan

- [ ] macOS: press Cmd+H → BambuStudio window is hidden
- [ ] macOS: click "BambuStudio" in menu bar → "Hide BambuStudio" entry works the same way
- [ ] No regression on Windows/Linux (change is inside `#ifdef __APPLE__`)

Closes #10740

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
      - master
-     - main
     paths:
      - 'deps/**'
      - 'src/**'
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -3,6 +3,7 @@ name: Build all
 on:
   push:
     branches:
+     - master
      - main
     paths:
      - 'deps/**'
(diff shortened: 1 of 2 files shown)


tonghao-bbl:
LGTM @lanewei120

PR #10836, opened 2026-05-20
fix: filament weight resets to 1000g after 'push local to cloud'

### LINKED ISSUE #10766: AMS synced filaments in filament manager reset to 1000 grams
### Bambu Studio Version

2.7.0.55

Bambu Lab Official website

### OS version

Windows 11

CPU: AMD Ryzen 7 5825U
Memory: 16 GB
GPU: AMD Radeon Graphics

### Printer

Bambu Lab A1 Mini

### How to reproduce

1. Go to Filament Manager
2. Click Add Filament
3. Click read from AMS
4. Select a filament and click add
5. Click on the spool detail button on the filament
6. Click Edit Filament Info
7. Change the current net weight to something other than 1000 and click save
8. Press the push local to cloud button(if printer is on) or turn on the printer

### Actual results

Filament gets resets to 1000 g instead of the filament you set.

https://github.com/user-attachments/assets/643505c2-ea29-443d-b565-b00bd7d57959

### Expected results

The filament amount should be the set amount.

[Untitled.zip](https://github.com/user-attachments/files/27941250/Untitled.zip)

- [ ] Log file
- [x] Project file

BenJule:
## Summary

- After manually editing a filament's **current net weight** in the Filament Manager and pressing **Push local to cloud**, the weight was immediately reset back to 1000 g (the AMS default for a full spool).
- Root cause: the `on_push_done` callback triggered an unconditional `enqueue_pull()` after every successful push. Due to read-after-write lag in the cloud backend, the subsequent `GET /filaments` could return the **stale** `netWeight` (before the push was committed) and overwrite the user's edited value in the local store.

## Fix

`enqueue_pull()` is now only called after **CREATE** operations, where we genuinely need the server-assigned spool id. For **UPDATE** and **DELETE** pushes the local store is already consistent with what was sent, so triggering a pull introduces a race without any benefit.

## How to reproduce

1. Filament Manager → Add Filament → Read from AMS → select a spool → Add.
2. Click the spool detail icon → Edit Filament Info.
3. Change **Current Net Weight** to any value other than 1000 g and save.
4. Press **Push local to cloud** (printer must be on, or turn it on after saving).
5. Observe: weight is reset to 1000 g.

## Test plan

- [ ] Reproduce steps 1-4 above, weight should now stay at the edited value after pushing.
- [ ] Create a new spool manually and push, the pull should still happen (to get the server-assigned id).
- [ ] Delete a spool and push, no spurious pull should be triggered.

Fixes #10766

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
      - master
-     - main
     paths:
      - 'deps/**'
      - 'src/**'
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -3,6 +3,7 @@ name: Build all
 on:
   push:
     branches:
+     - master
      - main
     paths:
      - 'deps/**'
--- a/src/slic3r/GUI/DeviceWeb/FilaManagerVM.cpp
+++ b/src/slic3r/GUI/DeviceWeb/FilaManagerVM.cpp
@@ -77,24 +77,33 @@ FilaManagerVM::FilaManagerVM()
                                         const std::string& msg) {
             publish_push_failed(id, op, code, msg);
         });
-        // Cloud is the source of truth: whenever a local push succeeds we
-        // immediately enqueue a pull so the next spool list the web sees is
-        // whatever the cloud just acknowledged. The dispatcher is single-slot
-        // so the pull runs right after this push op returns to idle.
+        // After a successful push we refresh the spool list so the UI reflects
+        // the latest local state immediately.  A follow-up cloud pull is only
+        // necessary for CREATE operations — we need the server-assigned spool
+        // id before advertising the new row to the web UI.  For UPDATE and
+        // DELETE we already know the outcome and triggering a pull introduces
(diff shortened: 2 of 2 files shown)


BenJule:

JingyuSunBambu:
We appreciate your valuable input and the suggested fixes for Bambu Studio. We’ve reviewed the issue with our product team and decided to lock the weight editing for RFID filaments, as the device auto-updates this data and manual changes would cause conflicts. As for the non-RFID spools resetting to 1000g on the cloud, we could not replicate that issue on our end. Please let us know if you have any other suggestions!

BenJule:
Thanks for the clarification.

The original report reproduces the reset after using “Read from AMS”, editing the current net weight, and pushing the local data to the cloud, but it does not clearly state whether the selected spool was RFID or non-RFID.

To make sure I understand the intended solution correctly:

1. For RFID filaments, should I revise this PR so that the current net weight cannot be edited manually?
2. For non-RFID filaments, should the weight remain editable, with any reset issue handled separately once a reproducible case and logs are available?

I can retest the non-RFID case and provide a recording and debug logs if I am able to reproduce it.

Issue #10642, opened 2026-05-09
Add a version number somewhere in homepage


Astatyr:
**Is your feature request related to a problem? Please describe.**
I am unable to find the version number without checking the folders

**Describe the solution you'd like**
Add a version number somewhere in the homepage

**Describe alternatives you've considered**
Add a version number in the title

Mujin-code:
Hello, thank you for your feedback. You can check the version number here
<img width="598" height="399" alt="Image" src="https://github.com/user-attachments/assets/d54e4d0e-bf7e-45fe-bb27-4aae6efac337" />

LJKraus:
_Help→ About_ is the standard location to find any program's version number, at least in Windows.

The version number is also displayed when Studio is launched.

BenJule:
The `update_title()` method was a no-op. It now sets the topbar label to `Bambu Studio X.Y.Z` on Windows/Linux so the version is always visible without opening the About dialog.

Pull request: https://github.com/bambulab/BambuStudio/pull/10837

tonghao-bbl:
Hi @Astatyr Since `Help→ About ` is the most comon way to show version on most platforms, we won't implement this feature.

Thanks for your attention!

PR #10837, opened 2026-05-20
feat: show unsaved-changes indicator (*) in the window title on all platforms

### LINKED ISSUE #9987: The filename in the window title Doesn't Show '*' when the file is modified
### Bambu Studio Version

2.5.0.66

Bambu Lab Official website

### OS version

Tahoe 16.3.1

_No response_

### Printer

Bambu Lab H2S

### Printer Firmware Version

01.01.02.00

### How to reproduce

1. Open a new or existing project in Bambu Studio.
2. Observe that there is a title of the project in the Window Title. It is "Untitled" for a new project.
3. Make a change, such as importing a new part or moving an existing part.

### Actual results

4. Notice that the title does not include a trailing '*' to indicate that the file is unsaved.

### Expected results

On Apple Macs, a standard interface feature is that when a file is changed, the window title changes, adding a '*' to show the user that the file contains unsaved changes. This is standard and critical because otherwise the user has now way of knowing that the file has unsaved work. I the program quits or is killed, the user will lose work.

Note: In many modern programs, including web-based apps, this feature is unnecessary because all changes are immediately saved to disk. However, Bambu Studio doesn't work this way, so needs to inform the user that they have unsaved work. Appending the '*' is the standard way to do this. 

none

- [ ] Log file
- [ ] Project file

BenJule:
## Summary

Reduced to **only the unsaved-changes indicator**, as requested in review.

On macOS the native window title already shows the project name (set in `Plater::priv::set_project_name`). This change prepends `* ` to that title while the project has unsaved changes (`is_project_dirty()`), matching the standard edited-document hint, and removes it again after saving. The `*` is refreshed from the existing idle handler as the dirty state changes.

**Windows/Linux behaviour is unchanged.**

## Notes
- Dropped from the earlier version of this PR (per review feedback): the always-visible version string (#10642) and the Windows/Linux title changes.
- The earlier implementation was guarded by `#ifndef __APPLE__`, so it never touched the macOS title — even though #9987 is specifically about macOS. The logic now lives on the macOS title path so it actually fixes the reported issue.
- Reverted an unrelated `build_all.yml` trigger change that had slipped into this branch.

## Test plan (macOS)
- [ ] Open a project: title shows the project name.
- [ ] Make a change (e.g. move an object): title becomes `* ProjectName`.
- [ ] Save: the `*` disappears.

Closes #9987

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -787,6 +787,9 @@ DPIFrame(NULL, wxID_ANY, "", wxDefaultPosition, wxDefaultSize, BORDERLESS_FRAME_
                  m_topbar->EnableUndoItem(m_plater->can_undo());
                  m_topbar->EnableRedoItem(m_plater->can_redo());
              }
+             // Keep the unsaved-changes "*" in the title in sync on all platforms (#9987).
+             if (m_plater)
+                 update_title();
          }));
 #ifdef _MSW_DARK_MODE
     wxGetApp().UpdateDarkUIWin(this);
@@ -1132,7 +1135,29 @@ void MainFrame::update_filament_tab_ui()
 
 void MainFrame::update_title()
 {
-    return;
+    if (!m_plater)
+        return;
+    // Prepend "* " while the project has unsaved changes (#9987), on top of the project name
+    // that is already shown: in the custom topbar on Windows, and in the native window title
+    // on macOS/Linux (both set by Plater::priv::set_project_name).
+    const wxString name  = m_plater->get_project_name();
+    const wxString title = (m_plater->is_project_dirty() && !name.IsEmpty()) ? ("* " + name) : name;
+    if (title == m_title_cache)
+        return;
+    m_title_cache = title;
+#ifdef __WINDOWS__
+    if (m_topbar)
+        m_topbar->SetTitle(title);
+    // Also reflect the "*" in the window/taskbar title, which set_project_name builds
+    // as "<name> - BambuStudio".
+    SetTitle(title + " - BambuStudio");
+#else
+    SetTitle(title);
+#ifdef __APPLE__
+    if (!title.IsEmpty())
+        update_title_colour_after_set_title();
+#endif
+#endif
 }
 
 void MainFrame::show_calibration_button(bool show, bool is_BBL)
--- a/src/slic3r/GUI/MainFrame.hpp
+++ b/src/slic3r/GUI/MainFrame.hpp
@@ -96,6 +96,7 @@ class MainFrame : public DPIFrame
 #endif
     bool     m_loaded {false};
     wxTimer* m_reset_title_text_colour_timer{ nullptr };
+    wxString m_title_cache;  // last value applied by update_title(), avoids redundant SetTitle
 
     wxString    m_qs_last_input_file = wxEmptyString;
     wxString    m_qs_last_output_file = wxEmptyString;
--- a/src/slic3r/GUI/Plater.cpp
+++ b/src/slic3r/GUI/Plater.cpp
@@ -15127,15 +15127,9 @@ void Plater::priv::set_project_name(const wxString& project_name)
 {
     BOOST_LOG_TRIVIAL(trace) << __FUNCTION__ << __LINE__ << " project is:" << project_name;
     m_project_name = project_name;
-    //update topbar title
-#ifdef __WINDOWS__
-    wxGetApp().mainframe->SetTitle(m_project_name + " - BambuStudio");
-    wxGetApp().mainframe->topbar()->SetTitle(m_project_name);
-#else
-    wxGetApp().mainframe->SetTitle(m_project_name);
-    if (!m_project_name.IsEmpty())
-        wxGetApp().mainframe->update_title_colour_after_set_title();
-#endif
+    // Update the window/topbar title. The platform-specific logic (and the
+    // unsaved-changes "*" prefix) lives in MainFrame::update_title().
+    wxGetApp().mainframe->update_title();
 }
 
 void Plater::priv::set_project_filename(const wxString& filename)


BenJule:

tonghao-bbl:
Hi @BenJule Showing `*` sounds good, and is important for Mac users as mentioned #9987 , but the other two feature seems not so good for users

Can you split this PR to just add a `*`?

src/slic3r/GUI/MainFrame.cpp
```diff
@@ -1135,7 +1133,20 @@ void MainFrame::update_filament_tab_ui()
...
+#ifndef __APPLE__
+    if (!m_topbar) return;
+    wxString title = wxString(SLIC3R_APP_FULL_NAME) + " " + SLIC3R_VERSION;
+    if (m_plater) {
+        wxString proj = m_plater->get_project_filename();
+        if (!proj.IsEmpty()) {
+            wxFileName fn(proj);
+            title = fn.GetName() + " - " + title;
+        }
+        if (m_plater->is_project_dirty())
```
tonghao-bbl: as discussed in the comment, only * is needed for now

BenJule:
Thanks @tonghao-bbl, done, I've trimmed this down to just the `*`.

While doing so I noticed the original implementation was guarded by `#ifndef __APPLE__`, so it only affected Windows/Linux and never touched the macOS title, even though #9987 (and your note that it matters for Mac users) is specifically about the macOS window title. So I moved the change onto the macOS path instead:

- On macOS the native window title already shows the project name (set in `Plater::priv::set_project_name`). `update_title()` now prepends `* ` to it while `is_project_dirty()` is true, and the existing idle handler keeps it in sync as the dirty state changes; it's removed again after saving.
- Dropped the always-visible version string (#10642) and the Windows/Linux title changes, Windows/Linux behaviour is unchanged.
- Reverted the unrelated `build_all.yml` trigger change that had slipped into this branch.

The PR is now a single commit touching only `src/slic3r/GUI/MainFrame.cpp` (+20/-1), and it closes only #9987. I don't have a Mac to build on, so a quick check on macOS would be appreciated.

tonghao-bbl:
Thanks @BenJule I think keep a `*`  on all platform is good, sorry for the misleading and I will test this when I'm availabel.

BenJule:
Thanks @tonghao-bbl, no worries, and done. I've extended the `*` to all platforms:

- `update_title()` now prepends `* ` to the project title while the project is dirty, on **Windows** (the custom topbar), **macOS** and **Linux** (the native window title), all of which already show the project name via `Plater::priv::set_project_name`.
- The existing idle handler keeps it in sync as the dirty state changes; a cached last-applied title avoids redundant title updates.

Still 2 files (`MainFrame.cpp` + `MainFrame.hpp`), and it closes #9987. I can't build it locally (no Windows/Mac here), so it's ready for your test whenever you're available, thanks for taking a look!

BenJule:
Quick follow-up: on Windows the `*` now also appears in the window/taskbar title (`* <name> - BambuStudio`), not just the topbar, so it's consistent across topbar, taskbar and the macOS/Linux titles. Same two files.

src/slic3r/GUI/MainFrame.cpp
```diff
@@ -1135,7 +1135,29 @@ void MainFrame::update_filament_tab_ui()
 
 void MainFrame::update_title()
```
tonghao-bbl, line 1136: Hi @BenJule while testing I found this piece of code is most duplicated with `plater::priv::set_project_name` below the comment `update topbar title`.

Can you move the code in plater.cpp into here btw? Thank you!

BenJule:
@tonghao-bbl Done, thanks for catching that. `Plater::priv::set_project_name` no longer duplicates the title logic; it now just stores `m_project_name` and calls `MainFrame::update_title()`, so the window/topbar title (including the unsaved-changes `*`) is built in a single place, consistent with how the rest of the code already refreshes the title. Behaviour is unchanged: `update_title_colour_after_set_title()` is preserved and is already `#ifdef __APPLE__` internally (no-op on Windows/Linux).

tonghao-bbl:
LGTM @lanewei120

lanewei120:
@BenJule 

Thanks for the cleanup. 

Centralizing the title update logic in MainFrame::update_title() is a good direction, and showing the unsaved-project marker in the native title bar makes sense.

However, I think this needs a little more validation before merging.

Although the PR title mentions macOS, the new logic also changes Windows behavior: both the custom topbar title and the native/taskbar title may now include the "*" prefix. Please confirm whether this is intentional.

I am also a bit concerned about m_title_cache. Since the cache only stores the computed project title, it may skip updates in cases where the topbar is recreated, the platform-specific suffix needs to be refreshed, or other title-related UI state changes without the project name changing.

Could you please verify the following flows on macOS and Windows:
1. new project → modify → save → modify again
2. open existing project → modify → save as
3. undo/redo dirty state changes
4. close project / empty project name
5. title behavior in Windows custom topbar and taskbar

If these cases are covered, I think the change is reasonable.

BenJule:
Thanks for taking a close look.

On Windows: yes, that's on purpose. tonghao asked earlier in this PR to keep the `*` on all platforms, so now both the topbar title and the native/taskbar title show it on Windows too. I went ahead and fixed the PR title since it still said "macOS".

About `m_title_cache` - it's really just there to keep the idle path cheap. The idle handler calls `update_title()` every tick now, and without the guard that means a `SetTitle()` on every frame, plus on macOS it would re-arm the title-colour timer each time, which flickers. The cache only kicks in when the new title is exactly the same string that's already shown.

For the three cases you were worried about:

- Topbar being recreated: it isn't. `m_topbar` is only `new`'d once in the MainFrame constructor (around line 257) and never replaced. Theme/dark-mode changes restyle the same topbar, they don't make a new one.
- Suffix: the only suffix is the literal `" - BambuStudio"`, so there's nothing that can change there on its own.
- Some other title state changing while the name stays the same: the title is just a function of the project name and the dirty flag, nothing else. Both go into the cached string, so if either one changes the cache misses and we update.

I don't have a Mac or a Windows box to run the GUI on, so I can't claim I clicked through all five - I built and ran on Linux and traced the rest through the code:

1. new → modify → save → modify again: empty/clean shows `name`, dirty shows `* name`, save drops the star, dirty again brings it back. Each step changes one of the two inputs so it never gets stuck on the cache.
2. open → modify → save as: `Existing` → `* Existing`, then Save As changes the name and clears dirty → `NewName`. Cache had `* Existing` so it updates.
3. undo/redo: those go through ProjectDirtyStateManager (lines 23/49/58) and the idle poll picks up anything that slips past, so each dirty flip recomputes.
4. close / empty name: with an empty name the title is just `""` (the `!name.IsEmpty()` check avoids a lonely `"* "`), same as before - Windows ends up with `" - BambuStudio"`, macOS gets an empty title and skips the colour refresh.
5. Windows topbar + taskbar: topbar gets `* name`, the window/taskbar gets `* name - BambuStudio`, both with the star.

If you'd rather not have the cache at all I can just drop it and call `SetTitle` every time - I'd only keep the macOS colour-timer guard so it doesn't flicker on idle. Whatever you prefer.

lanewei120:
thanks @BenJule

PR #10839, opened 2026-05-20
feat: add 'Don't show this warning again' to cloud connection failure dialog

### LINKED ISSUE #10589: Bambu Studio İnformation - Internet connection Warning
When using Bambu Studio on a PC without an internet connection, the application repeatedly shows “failed to connect to the cloud device server. please check yout network and firewall” warnings. This becomes disruptive, especially in offline workflows where cloud features are intentionally not used. The notifications appear frequently and interrupt the user experience during slicing and printer preparation.

Add an option to disable internet connectivity checks and related warnings entirely. This could be implemented as a simple toggle such as “Offline Mode” or “Disable network alerts” in the settings. When enabled, the application should stop attempting to connect to external services and suppress all related notifications.

Current workarounds include blocking the application via firewall or using LAN-only mode. However, these are not ideal:

* Firewall blocking is a system-level workaround and not user-friendly for most users.
* LAN mode still does not fully eliminate the warnings in some cases.

**Additional context**
Many users operate their printers in isolated or controlled environments where internet access is restricted or unnecessary. Providing a built-in offline mode would improve usability and reduce unnecessary interruptions in such scenarios.

BenJule:
## Summary

The "Failed to connect to the cloud device server" dialog previously reappeared on every MQTT connection failure, which interrupted LAN-only and offline users repeatedly.

**Fix:** replace `MessageDialog` with `RichMessageDialog` and add a **"Don't show this warning again"** checkbox. When the user checks it, the config key `suppress_cloud_warnings` is set to `"1"` and the dialog is permanently suppressed for future sessions.

The existing `is_showing` re-entrancy guard is preserved, and an early-return reads the config key before showing the dialog.

## Test plan

- [ ] With cloud connection failing, the dialog appears and shows the checkbox.
- [ ] Check the box and click OK: dialog no longer appears on subsequent connection failures.
- [ ] Manually remove `suppress_cloud_warnings=1` from app config: dialog reappears.

Closes #10589

### DIFF
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -2088,10 +2088,14 @@ void GUI_App::init_networking_callbacks()
 
                     static bool s_mqtt_connect_failed_dialog_shown = false;
                     if (s_mqtt_connect_failed_dialog_shown) return;
+                    if (app_config && app_config->get("suppress_cloud_device_server_warning") == "1") return;
                     s_mqtt_connect_failed_dialog_shown = true;
                     BOOST_LOG_TRIVIAL(trace) << "static: server connection failed";
-                    MessageDialog msg_dlg(nullptr, _L("Failed to connect to the cloud device server. Please check your network and firewall."), "", wxOK);
+                    RichMessageDialog msg_dlg(nullptr, _L("Failed to connect to the cloud device server. Please check your network and firewall."), "", wxOK);
+                    msg_dlg.ShowCheckBox(_L("Don't show this warning again"));
                     msg_dlg.ShowModal();
+                    if (msg_dlg.IsCheckBoxChecked() && app_config)
+                        app_config->set("suppress_cloud_device_server_warning", "1");
                 });
                 return;
             }
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -2090,10 +2090,17 @@ void GUI_App::init_networking_callbacks()
 
                     static bool s_mqtt_connect_failed_dialog_shown = false;
                     if (s_mqtt_connect_failed_dialog_shown) return;
+                    if (app_config && app_config->get("suppress_cloud_device_server_warning") == "1") return;
                     s_mqtt_connect_failed_dialog_shown = true;
                     BOOST_LOG_TRIVIAL(trace) << "static: server connection failed";
-                    MessageDialog msg_dlg(nullptr, _L("Failed to connect to the cloud device server. Please check your network and firewall."), "", wxOK);
+                    RichMessageDialog msg_dlg(nullptr, _L("Failed to connect to the cloud device server. Please check your network and firewall."), "", wxOK);
+                    msg_dlg.ShowCheckBox(_L("Don't show this warning again"));
                     msg_dlg.ShowModal();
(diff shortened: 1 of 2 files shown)


cafepromenade:
Or as a in app notification that appears like a dialog that doesn’t affect click throughs

BenJule:
This clean-up is exactly what the codebase needed after the limitations of the previous patch. Extending the robust focus evaluation down the hierarchy prevents various shortcuts from intercepting key strokes globally when custom layout wrappers have native controls embedded inside.

My fix in #10826 tackled the immediate breakdown with the 'Z' key in composite text fields, and having this broader refactoring in #10839 ensures we won't see similar regressions with other shortcuts across different OS platforms. 

Fully backing this change let's get this merged to finally keep the global hook out of the user's input fields!

XinZhangBambu:
Thanks for the contribution! I have a few small suggestions:

1. **Consider renaming `suppress_cloud_warnings`**  
   The current name is a bit too broad. A more specific key name would make the intent clearer and avoid confusion in future features.

2. **Allow re‑enabling the dialog in Preferences**  
   Since users may want to restore this warning later, it would be helpful to expose a toggle in the Preferences page to re‑enable the popup.

3. **Remove unrelated changes**  
   For example, updates in `build_all.yml` seem unrelated to this feature and should be excluded to keep the patch clean and focused.

Thanks again for improving the user experience!

BenJule:
Thanks for the detailed review, @XinZhangBambu! All three points are addressed in the latest push:

**1. Config key renamed**
`suppress_cloud_warnings` → `suppress_cloud_device_server_warning` — now scoped specifically to this dialog so it cannot accidentally suppress unrelated warnings in the future.

**2. Preferences toggle added**
`Preferences → Project → "Don't show cloud device server connection warnings"` — unchecking it clears the suppression immediately, so users can restore the warning without touching config files.

**3. Unrelated CI changes removed**
The `build_all.yml` commits have been dropped. The PR now contains exactly one commit touching only `GUI_App.cpp` and `Preferences.cpp`.

BenJule:

BenJule:
Thanks @XinZhangBambu, all three points are addressed in the current revision:

1. **Renamed** the key to `suppress_cloud_device_server_warning`, scoped to this specific dialog instead of the broad `suppress_cloud_warnings`.
2. **Added a Preferences → General toggle** ("Don't show cloud device server connection warnings") so users can re-enable the dialog later.
3. **Dropped the unrelated `build_all.yml` change**, the PR now only touches `GUI_App.cpp` and `Preferences.cpp`.

BenJule:
Thanks for the suggestions @XinZhangBambu - all three are addressed in the current version:

1. **Renamed** the key to `suppress_cloud_device_server_warning` (specific to this dialog instead of the broad `suppress_cloud_warnings`).
2. **Re-enable in Preferences**: added a "Don't show cloud device server connection warnings" checkbox on the Preferences page, with a tooltip explaining that unchecking it restores the warning.
3. **Unrelated changes removed**: the diff is now just `GUI_App.cpp` + `Preferences.cpp`, no `build_all.yml`.

Ready for another look, thanks!

BenJule:
Rebased onto latest master. master meanwhile added its own once-per-session guard for this dialog, so I merged the two: kept that guard plus the connection-status update, and layered the persistent 'don't show again' opt-in on top. Dropped my old is_showing re-entry lock since the session guard already covers that.

PR #10844, opened 2026-05-20
fix: AMS slot labels always show 'A' when selecting a non-first AMS unit

### LINKED ISSUE #10512: Add Filament in Filament Manager - AMS unit identifier does not reflect any device but "A"
### Bambu Studio Version

Beta 2.6.1.55

Bambu Lab Official website

### OS version

Windows 11

_No response_

### Printer

bambu H2C (H2D Upgraded)

### How to reproduce

1. Go to Filament Manager
2. Click Add Filament
3. Click Read From AMS
4. Click the AMS Icons.

### Actual results

A1, A2, A2, A4 descriptors stay the same no matter which ams unit icon is chosen.

### Expected results

AMS callout should represent the correct AMS unit chosen

No Crash or Project file required for this to happen.

- [ ] Log file
- [ ] Project file

BenJule:
## Summary

- In the **Add Filament → Read from AMS** dialog, the slot labels (A1-A4) were hardcoded and never updated when the user switched to a second or third AMS unit.
- With multi-AMS setups, selecting AMS B would still display slot labels A1, A2, A3, A4, matching none of the identifiers shown everywhere else in the UI (filament grouping, bound_ams_id field, etc.).
- Fix: derive the unit letter from `currentUnit.ams_id` (`'0'` = A, `'1'` = B, …) and construct `slotLabels` dynamically so they always reflect the currently selected AMS unit.

## Change

One-line change in `AddEditDialog.tsx`:

```tsx
// Before
const slotLabels = ['A1', 'A2', 'A3', 'A4'];

// After
const unitLetter = currentUnit ? String.fromCharCode('A'.charCodeAt(0) + parseInt(currentUnit.ams_id, 10)) : 'A';
const slotLabels = [1, 2, 3, 4].map((n) => `${unitLetter}${n}`);
```

## Test plan

- [ ] Single AMS: slot labels still show A1-A4
- [ ] Two AMS units connected: switching to unit 2 shows B1-B4; switching back shows A1-A4
- [ ] Three+ AMS units: C1-C4 for unit 3, etc.

Fixes #10512

### DIFF
--- a/src/slic3r/GUI/DeviceWeb/device_page/src/features/filament-manager/AddEditDialog.tsx
+++ b/src/slic3r/GUI/DeviceWeb/device_page/src/features/filament-manager/AddEditDialog.tsx
@@ -1846,7 +1846,8 @@ export function AddEditDialog({
   if (!open) return null;
 
   const currentUnit = amsData?.ams_units.find((u) => u.ams_id === selectedUnit);
-  const slotLabels = ['A1', 'A2', 'A3', 'A4'];
+  const unitLetter = currentUnit ? String.fromCharCode('A'.charCodeAt(0) + parseInt(currentUnit.ams_id, 10)) : 'A';
+  const slotLabels = [1, 2, 3, 4].map((n) => `${unitLetter}${n}`);
   // STUDIO-18344: form lock + form visibility derive from the new selection
   // model. Multi-select (>=2) hides the form entirely, so the lock flags
   // only matter in single-select mode.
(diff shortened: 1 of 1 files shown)


BenJule:

BenJule:
Rebuilt this branch cleanly on top of master. It had picked up some unrelated commits and stale device_page build artifacts along the way, so I reduced it back to just the actual fix. master also moved the source to features/filament-manager/, so the change now lands there (the hardcoded slotLabels bug is still present at that line on master). This only touches the .tsx source; the device_page bundle should be regenerated by the build.

JingyuSunBambu:
Thank you for identifying the AMS slot numbering issue in the 'Add Filament' menu and for suggesting a fix. Since different types of AMS follow distinct naming rules on the Device page, this problem requires a more nuanced approach than a simple fix. We are working on a resolution. Thanks again for your support of Bambu Studio!

BenJule:
Thank you for the clarification.

I understand that the different AMS types require distinct naming rules and that the issue therefore needs a broader solution than the approach used in this PR.

Since you are already working on a comprehensive fix, I will close this PR to avoid conflicting with or duplicating that implementation.

Thanks again for reviewing the contribution.

Issue #9418, opened 2026-01-16
Leading Spaces Crash


taylorjsam:
### Bambu Studio Version

2.4.0.70

### Where is the application from?

Bambu Lab Official website

### OS version

MacOS 15.7.2 (24G325)

### Additional system information

'24 Apple M4 MacBook Pro 14 Core 48GB ram

### Printer

P2S

### Printer Firmware Version

01.01.01.00

### How to reproduce

Added snap in name tag for a planter, tried to customize my adding the text "Lettuce" to the surface. Object was faced down with text 0.2mm thick, embedded .2mm into surface at a .2mm layer height. 

<img width="1152" height="853" alt="Image" src="https://github.com/user-attachments/assets/cc99ee57-1eee-4f4d-afba-9123f1f32401" />

### Actual results

Tried to add leading spaces to offset the text to the correct area. After 4-5 leading spaces BambuStudio crashes

### Expected results

BambuStudio doesn't crash

### Project file & Debug log uploads

[Plant Tag.3mf.zip](https://github.com/user-attachments/files/24679442/Plant.Tag.3mf.zip)

### Checklist of files to include

- [ ] Log file
- [x] Project file

fahaaaxiki:
Sorry, I didn't reproduce your issue in my computer. Can you reproduce this issue again and could send the crash report to us to analyze it？

Haidiye00:
Hello, can you download QuickRecorder software and record a GIF of the operation video and upload it? Thank you                         [lihaoyun6/QuickRecorder: A lightweight screen recorder based on ScreenCapture Kit for macOS / 基于 ScreenCapture Kit 的轻量化多功能 macOS 录屏工具](https://github.com/lihaoyun6/QuickRecorder/releases/tag/1.6.5)

adele-with-a-b:
Hit the same crash and tracked it down. Filed a candidate fix at #10847.

[...]

Issue #7782, opened 2025-08-07
Cannot Add SVG to Surface of Object


kimcharise:
### Bambu Studio Version

2.1.1.52

### Where is the application from?

Bambu Lab Official website

### OS version

macOS 15.6 and Windows 11

### Additional system information

Windows: CPU: AMD Ryzen 3 2.4 Ghz
                  RAM: 8.0 GB
                  Display: 1920 X 1080 60 hz

Mac: CPU: M4
          RAM: 32 GB
          Display:  4480 x 2250 Retina

### Printer

Bambu X1C and Bambu H2D

### How to reproduce

1. Go to "Add SVG"
2. Check "Use Surface"
3. It crashes in macOS and hangs up in Windows

### Actual results

Full Report
-----------

{"app_name":"BambuStudio","timestamp":"2025-08-07 18:35:01.00 -0400","app_version":"02.01.01.52","slice_uuid":"b60d0c97-0a4a-3044-a1bb-ca6869105e32","build_version":"","platform":1,"bundleID":"com.bambulab.bambu-studio","share_with_app_devs":0,"is_first_party":0,"bug_type":"309","os_version":"macOS 15.6 (24G84)","roots_installed":0,"name":"BambuStudio","incident_id":"985FD1AE-529E-4FB1-9CCC-D52BD953E04B"}
{
  "uptime" : 37000,
  "procRole" : "Foreground",
  "version" : 2,
  "userID" : 501,
  "deployVersion" : 210,
  "modelCode" : "Mac16,3",
  "coalitionID" : 1335,
  "osVersion" : {
    "train" : "macOS 15.6",
    "build" : "24G84",
    "releaseType" : "User"
  },
  "captureTime" : "2025-08-07 18:35:01.5453 -0400",
  "codeSigningMonitor" : 2,
  "incident" : "985FD1AE-529E-4FB1-9CCC-D52BD953E04B",
  "pid" : 12258,
  "translated" : false,
  "cpuType" : "ARM-64",
  "roots_installed" : 0,
  "bug_type" : "309",
  "procLaunch" : "2025-08-07 18:30:36.9436 -0400",
  "procStartAbsTime" : 900021892009,
  "procExitAbsTime" : 906372010500,
  "procName" : "BambuStudio",
  "procPath" : "\/Applications\/BambuStudio.app\/Contents\/MacOS\/BambuStudio",
[...]

Issue #5995, opened 2025-02-24
Bambu studio crashes when adding negative part modifier SVG to model


Qjeezy:
### Bambu Studio Version

1.10.1.50

### Where is the application from?

Bambu Lab Official website

### OS version

macOS Sonoma 14.6.1

### Additional system information

Chip: Apple M2 Pro
Memory: 16GB

### Printer

A1 Mini, P1S, and X1-C

### How to reproduce

1.) Add/open model to plate in Bambu studio
2.) Right click model
3.) Click “Add Negative part”
4.) Click “SVG” at the bottom
5.) Choose SVG you want to add & click open
6.) Adjust size to your liking
7.) Click box next to “use surface” to enable use surface. 

### Actual results

Bambu studio will crash and close. Sometimes it happens when adjusting the sizing parameters as well. I have not once been able to successfully use this feature. It crashes every single time. 

### Expected results

The SVG should snap to the objects surface and finish the operation. 

### Checklist of files to include

- [ ] Log file
- [ ] Project file

Haidiye00:
![Image](https://github.com/user-attachments/assets/2580cd7e-fe0c-41b0-a280-d92dce99316b)
Hello, can you provide SVG or 3mf? I tested it briefly and it didn't crash

Qjeezy:
> ![Image](https://github.com/user-attachments/assets/2580cd7e-fe0c-41b0-a280-d92dce99316b) Hello, can you provide SVG or 3mf? I tested it briefly and it didn't crash

![Image](https://github.com/user-attachments/assets/5c1925aa-490d-440d-aa21-3d9d68a72257)

I cannot seem to attach a 3mf or the step file I am using. The model I am trying to attach the SVG to is 74,856 triangles and is approximately 67x67x150 in size. 

It seems to work fine with smaller more basic objects. Is there another way to attach the 3mf here?

Haidiye00:
![Image](https://github.com/user-attachments/assets/81a1c00d-3917-4ef2-ae13-f754859f32af)
Hello, can you download ScreenToGif software and record a GIF of the operation video and upload it? Thank you
[Release ScreenToGif 2.41.1 · NickeManarin/ScreenToGif](https://github.com/NickeManarin/ScreenToGif/releases/tag/2.41.1)

Haidiye00:
Hello, I suggest you upload 3mf. Thank you

Qjeezy:
> Hello, I suggest you upload 3mf. Thank you

[...]

PR #10847, opened 2026-05-21
fix(threads): bump worker thread stack to 16MB to survive CGAL emboss

adele-with-a-b:
## Summary

Fixes a `SIGBUS` crash in the Emboss text-cut workflow when the worker thread overflows its 4 MB default stack. CGAL's `Polygon_mesh_processing::corefine` falls back from filtered interval arithmetic to exact rational arithmetic (mpq_class) on near-degenerate input, and the constrained 2D triangulation walker `Triangulation_2::march_locate_2D` recurses deeply enough to exceed 4 MB on real models. The fault address sits exactly inside the next thread's stack guard page — the textbook macOS stack-overflow signature.

This appears to be the same crash class as several already-open issues:

- #7782 — "Cannot Add SVG to Surface of Object" — SVG → "Use Surface" crashes on macOS. Same `Slic3r::cut_surface` → `Polygon_mesh_processing::corefine` path.
- #5995 — "Bambu studio crashes when adding negative part modifier SVG to model" — Negative-part SVG modifier with "Use Surface". Same code path.
- #9418 — "Leading Spaces Crash" — Text emboss at 0.2 mm embed depth on macOS M4. Same code path. macOS-specific symptoms match the stack-overflow fingerprint exactly.

Reporters of those issues and a maintainer with the original repro models can verify the fix end-to-end.

Bump `Slic3r::create_thread`'s default stack size from 4 MB to 16 MB on 64-bit. Resident memory remains proportional to actual stack depth on every target (defer-commit on macOS / Linux mmap'd stacks; `STACK_SIZE_PARAM_IS_A_RESERVATION` on Boost.Thread for Win32 means the reserve does not pre-commit).

## Reproducer

- BambuStudio Dev v02.07.00.55 on macOS 26.4.1 / Apple Silicon
- Emboss → text → embed depth ~0.6 mm into a real model with non-trivial topology under the text region (curved seam, sharp edge, or thin font stems all suffice to trigger the exact-arithmetic fallback)
- Crash report excerpt:

```
Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x00000001712ac000
Bad access in stack guard region for thread 45 but crash was associated
with thread 44 - possible stray access?

VM Region Info: 0x1712ac000 is in 0x1712ac000-0x1712b0000;
  Stack            170ea4000-1712ac000  [ 4128K] rw-/rwx  thread 44
--->STACK GUARD    1712ac000-1712b0000  [   16K] ---/rwx  stack guard for thread 45

Thread 44 Crashed:
  __gmpn_mul_1 / __gmpz_mul / __gmpq_mul
  CGAL::determinant<mpq_class>
  Projected_orientation_with_normal_3
  Filtered_predicate_with_state::operator()
  Triangulation_2<Projection_traits_3<Epeck>>::march_locate_2D
  Surface_intersection_visitor::triangulate_intersected_faces
  Polygon_mesh_processing::corefine
  Slic3r::cut_surface
  Emboss::cut_surface_to_its
  Emboss::GenerateTextJob::get_text_mesh
  PlaterWorker::PlaterJob::process
```

The thread's stack region is exactly 4128K — the default 4 MB plus a small TLS overhead.

## Why 16 MB and not a tighter number

We have one observed reproducer; the 16 MB value is chosen as 4× defensive headroom over that, not as a measured upper bound. CGAL's `march_locate_2D` recursion depth is data-dependent on input geometry, and we have no a-priori bound. Picking 16 MB rather than 8 MB trades a small amount of bounded virtual address space for a wider safety margin against future heavier emboss workloads.

## Why this scope and not a per-callsite override

Considered narrowing the bump to just the Emboss `GenerateTextJob` path by plumbing a stack-size override through `PlaterWorker` → `BoostThreadWorker(attribs, name)`. Rejected because:

1. The cost is bounded virtual address space on 64-bit (resident memory tracks actual depth on every supported OS).
2. The same exact-arithmetic recursion shape appears in OrcaSlicer issue trackers for STEP loading and other CGAL-heavy paths in this codebase, so a uniform bump preempts the same crash in adjacent code.
3. Per-callsite plumbing adds maintenance burden disproportionate to the saved address space.

## Cumulative cost

`Slic3r::create_thread` has 22 callsites across the codebase. Realistic peak concurrent live count is on the order of 10–15 workers (Plater UI worker, slicing process, FDM-support gizmo, STEP loader, network sync helpers, per-task sender threads in `TaskManager` up to `MaxSendingAtSameTime`, per-machine info threads in device-list dialogs). At 16 MB reserve × ~15 = ~240 MB of address-space commitment in the worst case, which is bounded on any 64-bit target.

## Test plan

- [x] Apple Silicon / macOS 26.4.1 / Apple Clang 21 / CMake 4.x — compiles clean. The original reproducer on the author's working model no longer crashes.
- [ ] Same build on a smaller-RAM Mac would also be useful to confirm there's no resident-memory regression. Author's expectation is "no" because pages are deferred-allocated on macOS, but a maintainer with diverse hardware could verify with the same emboss reproducer.
- [ ] Linux / Windows build verification by maintainers.

### DIFF
--- a/src/libslic3r/Thread.hpp
+++ b/src/libslic3r/Thread.hpp
@@ -46,11 +46,23 @@ void name_tbb_thread_pool_threads_set_locale();
 template<class Fn>
 inline boost::thread create_thread(boost::thread::attributes &attrs, Fn &&fn)
 {
-    // Duplicating the stack allocation size of Thread Building Block worker
-    // threads of the thread pool: allocate 4MB on a 64bit system, allocate 2MB
-    // on a 32bit system by default.
-    
-    attrs.set_stack_size((sizeof(void*) == 4) ? (2048 * 1024) : (4096 * 1024));
+    // Stack size for our worker threads. Originally duplicated TBB's pool
+    // default (4 MB), but the Emboss text-cut path calls into CGAL's
+    // Polygon_mesh_processing::corefine, which falls back from filtered
+    // interval arithmetic to exact rational arithmetic (mpq_class) on
+    // near-degenerate input, and the constrained 2D triangulation walker
+    // (Triangulation_2::march_locate_2D) can recurse deeply enough to
+    // exceed 4 MB on real models -- producing a SIGBUS at the next thread's
+    // stack guard page on macOS / Linux.
+    //
+    // 16 MB chosen as 4x defensive headroom over the observed crash
+    // threshold (n=1 reproducer at exactly 4 MB on macOS arm64). All three
+    // platforms defer-commit reserved stack pages: macOS / Linux mmap the
+    // stack and only fault in pages on touch; Boost.Thread on Win32 passes
+    // STACK_SIZE_PARAM_IS_A_RESERVATION to _beginthreadex, so the value is
+    // a reserve, not the initial commit. Resident memory therefore stays
+    // proportional to actual stack depth on every target.
+    attrs.set_stack_size(16 * 1024 * 1024);
     return boost::thread{attrs, std::forward<Fn>(fn)};
 }
 


Haidiye00:
approve

lanewei120:
thanks for doing this~

Qjeezy:
Thank you for digging into this and developing a fix! Hopefully they use it!

Haidiye00:
hello,can you download QuickRecorder software and record a tested GIF of the operation video and upload it? Thank you  
Our internal tests currently find that there is still a 50% chance of crashing.

Haidiye00:
After our evaluation, we temporarily rolled back the code. We are very sorry, and please provide a self-test GIF when you have time. Thank you.

PR #10859, opened 2026-05-21
fix: AMS slot edit icon click detection misses Y-bound check

### LINKED ISSUE #10858: AMS slot edit icon is triggered by clicks outside its Y bounds
## Description

In `AMSLib::on_left_down()` (`src/slic3r/GUI/Widgets/AMSItem.cpp`), the hit-test for the filament slot edit icon contains a bug: the upper-bound Y check compares `top <= bottom` instead of `pos.y <= bottom`.

## Steps to Reproduce

1. Open the AMS panel with at least one filament slot loaded.
2. Click **below** the edit icon (pencil/edit button) on an AMS slot.
3. The edit dialog still opens even though the click was not on the icon.

## Expected Behavior

The edit dialog should only open when the click is within the edit icon's bounding box (both X and Y axes).

## Root Cause

```cpp
// AMSItem.cpp — AMSLib::on_left_down()
// Buggy condition:
if (pos.x >= left && pos.x <= right && pos.y >= top && top <= bottom) {
```

`top <= bottom` is always `true` by construction (top is computed as `size.y - 20px - icon_height` and bottom as `size.y - 20px`), so the Y-upper-bound check never filters out clicks below the icon.

## Fix

```cpp
if (pos.x >= left && pos.x <= right && pos.y >= top && pos.y <= bottom) {
```

Replace `top <= bottom` with `pos.y <= bottom`.

Affects all AMS model variants: AMS, N3F, N3S, EXT_SPOOL, AMS_LITE.

BenJule:
## Summary

`AMSLib::on_left_down()` checks `top <= bottom` as the upper Y bound for the edit-icon hit-test, but `top` is always less than or equal to `bottom` by construction, so this condition is always `true`. Any click with the correct X range triggers the edit dialog regardless of how far below the icon the click lands.

**One-character fix:** replace `top <= bottom` with `pos.y <= bottom`.

Applies to all AMS model variants (AMS, N3F, N3S, EXT_SPOOL, AMS_LITE).

Closes #10858

## Changed file

- `src/slic3r/GUI/Widgets/AMSItem.cpp` — line 1039

## Test plan

- [ ] Click on the edit icon of an AMS slot → edit dialog opens
- [ ] Click just below the edit icon → dialog does NOT open
- [ ] Click above the edit icon (but still in X range) → dialog does NOT open

### DIFF
--- a/src/slic3r/GUI/Widgets/AMSItem.cpp
+++ b/src/slic3r/GUI/Widgets/AMSItem.cpp
@@ -1038,7 +1038,7 @@ void AMSLib::on_left_down(wxMouseEvent &evt)
                 bottom = size.y - FromDIP(20);
             }
 
-            if (pos.x >= left && pos.x <= right && pos.y >= top && top <= bottom) {
+            if (pos.x >= left && pos.x <= right && pos.y >= top && pos.y <= bottom) {
                 if (m_selected) {
                     if (m_info.material_state == AMSCanType::AMS_CAN_TYPE_VIRTUAL) {
                         post_event(wxCommandEvent(EVT_VAMS_ON_FILAMENT_EDIT));


BenJule:

tonghao-bbl:
LGTM @lanewei120

lanewei120:
Hi @BenJule 

could you rebase to the master base, and reduce the commit to 1?

thanks

BenJule:
@lanewei120 Done, rebased onto current master and squashed to a single commit. I also dropped the unrelated CI workflow change (`build_all.yml`) that had accidentally slipped in, so the PR now contains only the one-line `AMSItem.cpp` fix. Thanks!

lanewei120:
thanks @BenJule

PR #10860, opened 2026-05-21
fix: project profile name shows stale value after editor save

### LINKED ISSUE #10769: Bambu Studio [Profile Name] ‘s display content is wrong 【配置名称】显示内容不正确
### Bambu Studio Version

win-v02.06.00.51-20260417160415

Bambu Lab GitHub releases

### OS version

Win10

_No response_

### Printer

X2D

### How to reproduce

1. Click [Project]. 点击【项目】。
2. Click [Edit]. 点击【编辑】。
3. Change the [Profile Name] to anything you want and [Save] it. 任意修改【配置名称】并【保存】。
4. Check the result of your edit which is not correct. 查看【项目】的【配置名称】不是刚才修改并保存的内容。

### Actual results

”idea Lab“ which is the author of this project I guess. ”idea Lab“我猜是项目的作者。

### Expected results

Whatever you edit just now. 你刚才编辑的内容。

[FilamentClip.zip](https://github.com/user-attachments/files/27945402/FilamentClip.zip)

- [ ] Log file
- [x] Project file

BenJule:
## Summary

After saving in the Project editor and navigating back to the index page, the Profile Name displays the original (pre-edit) value instead of the newly saved one.

**Root cause:** `clear_model_info()` empties `m_last_payload` during the save flow. When `index.html` loads and immediately sends `request_3mf_info`, C++ does nothing because `m_last_payload` is empty. The background `on_reload()` thread that holds the fresh data fires its `RunScript` callback while the webview is still navigating; the script is dropped and the page never receives the updated model data.

**Fix:** When `request_3mf_info` arrives with an empty `m_last_payload`, call `update_model_data()` to trigger a fresh reload. The resulting `on_reload()` dispatches the current (post-save) model data directly to the now-fully-loaded page.

Closes #10769

## Changed file

- `src/slic3r/GUI/Project.cpp`, `OnScriptMessage` / `request_3mf_info` handler

## Test plan

- [ ] Open a project → click **Edit** → change Profile Name → click **Save**
- [ ] Profile Name on the index page reflects the newly saved value
- [ ] Opening and re-saving updates the displayed name each time
- [ ] Opening the editor after a save shows the correct (updated) Profile Name

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
      - master
-     - main
     paths:
      - 'deps/**'
      - 'src/**'
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -3,6 +3,7 @@ name: Build all
 on:
   push:
     branches:
+     - master
      - main
     paths:
      - 'deps/**'
--- a/src/slic3r/GUI/Project.cpp
+++ b/src/slic3r/GUI/Project.cpp
@@ -389,6 +389,12 @@ void ProjectPanel::OnScriptMessage(wxWebViewEvent& evt)
                 wxGetApp().CallAfter([this, strJS] {
                     RunScript(strJS.ToStdString());
                 });
+            } else {
+                // m_last_payload was cleared (e.g. after a save) and the background
+                // reload may not have finished yet — or its dispatch was dropped during
+                // webview navigation.  Trigger a fresh reload so the new page receives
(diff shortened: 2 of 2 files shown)


BenJule:

EricDou1:
Hi @BenJule,

When we tried to reproduce locally, we found that the Profile Name on the index page is always empty regardless of any save operation — it turns out TranslatePage() was wiping the <label id="profileName"> element entirely.

The stale value you observed might actually be caused by the same issue. Could you verify this on the latest version?

We'll fix this issue on our side. Thanks!

PR #10829, opened 2026-05-20
fix: 'Support for ABS' shown as 'ABS' in filament grouping dialog

BenJule:
## Problem

When a project loads both a regular filament (e.g. ABS) and its support variant (e.g. Support for ABS), the filament grouping dialog shows two identical **ABS** labels, making it impossible to distinguish the two slots.

## Root cause

Both preset types store `filament_type = "ABS"` in config. Two separate code paths read the raw values without checking `filament_is_support`:

1. `FilamentMapDialog.cpp` (`try_pop_up_before_slice`), dialog triggered before slicing
2. `Plater.cpp` (`open_filament_map_setting_dialog`), dialog opened manually

## Fix

**`FilamentMapDialog.cpp`:** Replace the direct option read with `get_filament_type()`, which returns the display string including the `Sup.` prefix for support filaments.

**`Plater.cpp`:** After reading raw `filament_type` values, prepend `"Sup."` for any slot whose preset has `filament_is_support = true`, matching the convention used by `get_filament_type()`.

## Test plan

- [ ] Load a project with ABS + Support for ABS slots
- [ ] Open the filament grouping dialog manually, verify **ABS** vs **Sup.ABS** labels
- [ ] Slice to trigger the dialog automatically, verify same distinct labels
- [ ] Verify pure-regular-filament projects are unaffected

Fixes #10773

PR #10861, opened 2026-05-21
fix: show 'Sup.' prefix for support filaments in filament grouping dialog

### LINKED ISSUE #10773: The name of Support for ABS in Filament Grouping is confusing
### Bambu Studio Version

win-v02.06.00.51-20260417160415

Bambu Lab GitHub releases

### OS version

win10

_No response_

### Printer

X2D

### How to reproduce

1. use ABS and Support for ABS
2. go to filament grouping
3. Two ABS there. Not ABS and Support for ABS

### Actual results

2 ABS

### Expected results

ABS and Support for ABS

Image

- [ ] Log file
- [ ] Project file

BenJule:
## Problem

When a project loads both a regular filament (e.g. ABS) and its support variant (e.g. Support for ABS), the filament grouping dialog (`FilamentMapDialog`) shows two identical **ABS** labels. This makes it impossible to tell which slot is the regular material and which is the support material.

## Root cause

Both preset types store `filament_type = "ABS"` in their config. The code in `Plater::open_filament_map_setting_dialog()` reads the raw `filament_type` option values without checking the `filament_is_support` flag, so both entries look the same.

## Fix

Before passing the type list to the dialog, iterate over the active filament presets and prepend `"Sup."` to any slot whose preset has `filament_is_support = true`. This is the same naming convention already used by `Preset::get_filament_type()` for display purposes elsewhere in the UI (yielding `"Sup.ABS"` for Support for ABS).

## Test plan

- [ ] Load a project with at least one ABS slot and one Support for ABS slot
- [ ] Open the filament grouping dialog
- [ ] Verify the two slots are labeled **ABS** and **Sup.ABS** respectively (not both **ABS**)
- [ ] Verify projects with only regular filaments are unaffected

Fixes #10773

### DIFF
--- a/src/slic3r/GUI/Plater.cpp
+++ b/src/slic3r/GUI/Plater.cpp
@@ -23146,6 +23146,17 @@ void Plater::open_filament_map_setting_dialog(wxCommandEvent &evt)
     const auto& project_config = wxGetApp().preset_bundle->project_config;
     auto filament_colors = config()->option<ConfigOptionStrings>("filament_colour")->values;
     auto filament_types = config()->option<ConfigOptionStrings>("filament_type")->values;
+    // Distinguish support filaments from their base material type so the grouping
+    // panel does not show identical labels (e.g. two "ABS" entries instead of
+    // "ABS" and "Sup.ABS" when one slot holds Support for ABS).
+    const auto& fila_preset_names = preset_bundle->filament_presets;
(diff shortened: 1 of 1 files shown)


BenJule:
Closing in favor of #10829 which covers both entry points to the dialog. The Plater.cpp fix has been merged into that branch.

PR #10811, opened 2026-05-20
fix: correct grammar and punctuation in English UI strings

BenJule:
## Summary

Four English grammar/spelling/punctuation errors corrected in UI strings displayed to users:

| File | Before | After |
|------|--------|-------|
| `src/slic3r/GUI/MediaPlayCtrl.cpp` | `"Problem occured."` | `"Problem occurred."` |
| `src/slic3r/GUI/SelectMachine.cpp` | `"When enable spiral vase mode…"` | `"When enabling spiral vase mode…"` |
| `src/slic3r/GUI/SyncAmsInfoDialog.cpp` | `"When enable spiral vase mode…"` | `"When enabling spiral vase mode…"` |
| `src/slic3r/GUI/Gizmos/GLGizmoAdvancedCut.cpp` | `"connectors is out of object"` | `"connectors are out of object"` |
| `src/slic3r/GUI/Gizmos/GLGizmoText.cpp` | `"Error:Detecting …outcomes.Please drag…"` | `"Error: Detecting …outcomes. Please drag…"` |

Source-string changes only, the translation catalogs (`bbl/i18n/*.po`, `BambuStudio.pot`) are left to the upstream gettext pipeline, per maintainer guidance on #10871.

Closes #10854.

PR #10809, opened 2026-05-20
fix: correct 'percision' typo to 'precision' in motion calibration UI strings

BenJule:
## Problem

Two printer status strings in `DeviceManager.cpp` (cases 36 and 37) used the misspelling **"percision"** instead of **"precision"**:

- `"Measuring motion percision"`
- `"Enhancing motion percision"`

These strings are shown to users in the device status panel during motion calibration.

## Fix

Correct both strings in `src/slic3r/GUI/DeviceManager.cpp`.

Source-string change only, the translation catalogs (`bbl/i18n/*.po`, `BambuStudio.pot`) are intentionally left untouched and regenerated by the upstream gettext pipeline, per maintainer guidance on #10871.

Closes #10856.

PR #10871, opened 2026-05-22
fix(devicemanager): percision -> precision in motion calibration

### LINKED ISSUE #10856: Misspelling 'percision' in motion calibration status messages
Two printer status strings in `DeviceManager.cpp` (cases 36 and 37) display the misspelling **"percision"** instead of **"precision"** during motion calibration:

- `"Measuring motion percision"` → should be `"Measuring motion precision"`
- `"Enhancing motion percision"` → should be `"Enhancing motion precision"`

These strings are shown in the device status panel while the printer performs motion calibration, so the typo is visible to all users.

mvanhorn:
## Summary

Closes #10856. The device status panel rendered "Measuring motion percision" and "Enhancing motion percision" during motion calibration (cases 36/37 in `src/slic3r/GUI/DeviceManager.cpp:162,164`). The intended word is `precision`. This PR fixes the source strings and updates the `msgid` keys in all 18 `bbl/i18n/*.po` files to match, so existing `msgstr` translations still resolve via gettext lookup.

## Why this matters

These two strings are visible to every user who triggers motion calibration, so the typo is high-visibility in the device status panel. Updating only `DeviceManager.cpp` would silently break the gettext lookup for every translated locale (the existing `msgid "Measuring motion percision"` entries no longer match the corrected source, so gettext falls back to English and translators' work is dropped on the floor until the next extraction cycle).

Updating the `msgid` key on each `.po` file at the same time as the source keeps the existing `msgstr` translations attached to the corrected source string. Translators don't have to redo their work, and non-English users keep seeing translated text instead of falling back to the corrected English.

## Changes

- `src/slic3r/GUI/DeviceManager.cpp:162,164` - `percision` -> `precision` in both `_L(...)` calls.
- `bbl/i18n/{cs,de,en,es,fr,hu,it,ja,ko,nl,pl,pt_BR,ru,sv,tr,uk,zh_CN,zh_TW}/BambuStudio_*.po` - `msgid` keys updated to match. 18 files, 2 lines each.

Total: 19 files, 38 added + 38 deleted lines. Mechanical search-and-replace of two exact strings.

## How to test

```
$ grep -rln "percision" --include="*.cpp" --include="*.po" --include="*.json" --include="*.h" .
(empty)
$ grep -n "Measuring motion precision\|Enhancing motion precision" src/slic3r/GUI/DeviceManager.cpp
162:        return _L("Measuring motion precision");
164:        return _L("Enhancing motion precision");
```

The 18 `.po` files each have two updated `msgid` lines; the matching `msgstr` translations are untouched.

Fixes #10856

AI-assisted.

### DIFF
--- a/bbl/i18n/cs/BambuStudio_cs.po
+++ b/bbl/i18n/cs/BambuStudio_cs.po
@@ -4412,10 +4412,10 @@ msgstr "Pozastavení (chyba první vrstvy)"
 msgid "Pause (nozzle clog)"
 msgstr "Pozastavení (ucpání trysky)"
 
-msgid "Measuring motion percision"
+msgid "Measuring motion precision"
 msgstr "Měření přesnosti pohybu"
 
-msgid "Enhancing motion percision"
+msgid "Enhancing motion precision"
 msgstr "Zvyšování přesnosti pohybu"
 
 msgid "Measure motion accuracy"
--- a/bbl/i18n/de/BambuStudio_de.po
+++ b/bbl/i18n/de/BambuStudio_de.po
@@ -4354,10 +4354,10 @@ msgstr ""
 msgid "Pause (nozzle clog)"
 msgstr ""
 
-msgid "Measuring motion percision"
+msgid "Measuring motion precision"
 msgstr ""
 
-msgid "Enhancing motion percision"
+msgid "Enhancing motion precision"
 msgstr ""
 
 msgid "Measure motion accuracy"
--- a/bbl/i18n/en/BambuStudio_en.po
+++ b/bbl/i18n/en/BambuStudio_en.po
@@ -4310,10 +4310,10 @@ msgstr ""
 msgid "Pause (nozzle clog)"
 msgstr ""
 
-msgid "Measuring motion percision"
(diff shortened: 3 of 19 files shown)


walterwongbbl (approved):

lanewei120:
Hi @mvanhorn 

no need to modify the translation files in this patch,
please just modify the cpp files here

thanks

BenJule:
Hi @mvanhorn — heads-up: #10809 fixes the same `percision`→`precision` string (#10856) and is part of a small series of source-string cleanups (#10811/#10813/#10814), now all reduced to `.cpp`-only per @lanewei120's note here. Happy to defer to whichever the maintainers prefer — just flagging the overlap so we don't both land it. Thanks for catching this too!

lanewei120:
thanks for doing this~

mvanhorn:
Really appreciate the review, @lanewei120. Small one, but fixing percision to precision in the motion config panel was overdue.

mvanhorn:
Thanks @lanewei120 - good to get percision fixed to precision in the motion calibration panel.

PR #10878, opened 2026-05-22
fix: Linux packaging — copy desktop/icons for non-AppImage builds, prevent set -e exit

BenJule:
## Summary

Two small Linux packaging fixes:

**1. Copy desktop file and icons into `build/package/` for non-AppImage builds** (`src/platform/unix/BuildLinuxImage.sh.in`)

The `install` step was only run when building an AppImage. For `.deb` / container builds that skip the AppImage step, `BambuStudio.desktop` and the icon tree were missing from `build/package/`, causing packaging scripts to fail. The install step is now always run unconditionally.

**2. Prevent `set -e` from exiting when GTK3 is not yet installed** (`linux.d/debian`)

`dpkg -l libgtk* | grep gtk-3-dev` exits with code 1 when the package is not installed, which combined with `set -e` would abort the script before the dependency-installation path is reached. Fixed by appending `|| echo ''` so the expression always succeeds.

### DIFF
--- a/linux.d/debian
+++ b/linux.d/debian
@@ -1,4 +1,4 @@
-FOUND_GTK3=$(dpkg -l libgtk* | grep gtk-3)
+FOUND_GTK3=$(dpkg -l libgtk* | grep gtk-3 || true)
 
 REQUIRED_DEV_PACKAGES=(
     autoconf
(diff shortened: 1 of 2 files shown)


BenJule:

PR #10879, opened 2026-05-22
fix: warn user when filament serial input sanitizes to empty string

BenJule:
## Summary

- Show a warning message when the user enters a filament serial number that gets sanitized (trimmed/filtered) to an empty string, preventing silent data loss

## Changes

- `src/slic3r/GUI/CreatePresetsDialog.cpp`: Add validation that warns the user when serial input becomes empty after sanitization

## Test plan

- [ ] Enter a filament serial number consisting only of whitespace or invalid characters
- [ ] Verify a warning is shown instead of silently saving an empty serial
- [ ] Verify normal serial numbers still save correctly

### DIFF
--- a/src/slic3r/GUI/CreatePresetsDialog.cpp
+++ b/src/slic3r/GUI/CreatePresetsDialog.cpp
@@ -1038,6 +1038,12 @@ wxBoxSizer *CreateFilamentPresetDialog::create_button_item()
         }
         boost::algorithm::trim(vendor_name);
         boost::algorithm::trim(serial_name);
+        if (!serial_str.IsEmpty() && serial_name.empty()) {
+            MessageDialog dlg(this, _L("The filament serial contains only invalid characters or spaces. Please correct it or leave it blank."),
+                              wxString(SLIC3R_APP_FULL_NAME) + " - " + _L("Info"), wxYES | wxYES_DEFAULT | wxCENTRE);
+            dlg.ShowModal();
+            return;
+        }
         if (vendor_name.empty() || serial_name.empty()) {
             MessageDialog dlg(this, _L("All inputs in the custom vendor or serial are spaces. Please re-enter."),
                               wxString(SLIC3R_APP_FULL_NAME) + " - " + _L("Info"), wxYES | wxYES_DEFAULT | wxCENTRE);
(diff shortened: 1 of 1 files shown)


src/slic3r/GUI/CreatePresetsDialog.cpp
```diff
@@ -1038,6 +1038,12 @@ wxBoxSizer *CreateFilamentPresetDialog::create_button_item()
         }
         boost::algorithm::trim(vendor_name);
         boost::algorithm::trim(serial_name);
+        if (!serial_str.IsEmpty() && serial_name.empty()) {
```
tonghao-bbl: * `!serial_str.IsEmpty()` is not needed here, check line 1022
* `or leave it blank.`  in the message dilaog is not OK here since we dont accpet empty string as checked above

BenJule:

tonghao-bbl (commented):
Hi @BenJule Can you have a look at these small issues in the comment?

BenJule:
Good catch, both fixed. Dropped the redundant serial_str.IsEmpty() guard since the empty case already returns earlier, and reworded the message so it no longer says you can leave it blank.

src/slic3r/GUI/CreatePresetsDialog.cpp
```diff
@@ -1038,6 +1038,12 @@ wxBoxSizer *CreateFilamentPresetDialog::create_button_item()
         }
         boost::algorithm::trim(vendor_name);
         boost::algorithm::trim(serial_name);
+        if (serial_name.empty()) {
```
tonghao-bbl, line 1041: The check is duplicated with L1042

tonghao-bbl (changes requested):

src/slic3r/GUI/CreatePresetsDialog.cpp
BenJule: Fixed, thanks. The second condition now only checks `vendor_name.empty()`, since the empty serial case already returns immediately above.

BenJule: Correction: my previous follow-up modified the earlier guard instead of the duplicated post-trim check. This is now fixed correctly: the first validation again checks both vendor and serial, while the later condition only checks `vendor_name.empty()` because the empty serial case already returns immediately above.

PR #10892, opened 2026-05-23
fix(slicer): invalidate slice result when instance printability toggles

### LINKED ISSUE #10891: Slice button stays grayed out after toggling object Printable flag
### Bambu Studio Version

2.2.1.55 (latest master, built from source — also reproducible on Bambu Lab Official 2.2.1.55 release)

Built from source by yourself or others

### OS version

macOS 26.5

### Printer

Bambu Lab H2C (multi-extruder)

### How to reproduce

1. Open a project with multiple objects on a single plate (4 copies of the same STL works fine for repro)
2. Click "Slice plate". Wait for slicing to complete.
3. Right-click an object → uncheck "Printable" (or use the printable checkbox in the object list).
4. Observe the slice button.

### Actual results

The slice button is grayed out / disabled. Clicking it has no effect.

The same happens when you re-check the previously-disabled object: the slice button stays disabled even though the object set has changed and a re-slice is needed.

The visual state of the object list updates correctly (object dims/highlights/etc.) — only the slice button state is wrong.

### Expected results

The slice button should become enabled immediately after toggling printability, because the cached slice result no longer reflects the current object set on the plate.

### Project file & Debug log

Repro is independent of project content — any project with ≥2 objects on a plate reproduces. Tested with 4 copies of an arbitrary STL.

Log evidence captured with a `BBL_INTERNAL_TESTING=1` build (so log is not encrypted) is included in the root-cause section below.

### Root cause analysis

Confirmed in source. In `Print::apply()`'s instance synchronization branch (`src/libslic3r/PrintApply.cpp` ~line 1757), the loop updates each instance's `printable` flag but only invalidates Print steps when `is_printable_filament_changed()` detects a filament-set change. A pure printability toggle that does not alter the filament assignment falls through that check.

Section 4 of `Print::apply` does fire `invalidate_all_steps()` for the toggled object, but the print steps are typically already in INVALID state from earlier background-process activity, so `invalidate_all_steps()` returns false and the overall `apply_status` caps at `APPLY_STATUS_CHANGED` (1) rather than `APPLY_STATUS_INVALIDATED` (2).

`Plater::priv::update_background_process` only calls `update_slice_result_valid_state(false)` on `APPLY_STATUS_INVALIDATED` (`src/slic3r/GUI/Plater.cpp:10210`). With `APPLY_STATUS_CHANGED`, the cached slice result stays valid, so `MainFrame::get_enable_slice_status()` returns false (button stays disabled) because `PartPlate::is_slice_result_valid()` is still true.

**Log evidence:** every printability toggle in the affected configuration produces:

```
update_print_volume_state, ... got N printable instances
... (instance synchronization runs without invalidating)
apply 2134: finished, ... apply_status 1, ...
update_slice_print_status m_slice_select 1: can_slice=1, ..., enable_slice=0, ...
```

Note `apply_status 1` (CHANGED) instead of `2` (INVALIDATED), and `enable_slice=0` (button disabled) on the line right after, even though the object set just changed.

### Regression context

Introduced by [`e22d328fa`](https://github.com/bambulab/BambuStudio/commit/e22d328fa9d1682c00617a4cad0498c67a7512f4) ("ENH: Optimizing the shape invalid condition for wipe tower of multi-extruder", May 16 2025), which:

- Added a `is_printable_filament_changed()` check in the per-instance loop, only invalidating when convex-hull-derived filament assignments change.
- Removed an unconditional `if (extruder_num > 1) invalidate_steps({psWipeTower, psSkirtBrim, psGCodeExport})` block from the `set_instances` path.

The optimization was correct for translation/rotation cases, but printability toggles do not change geometry, so they fall through both the new and old checks.

### Proposed fix

https://github.com/bambulab/BambuStudio/pull/10892

The 4-line patch invalidates `{psSkirtBrim, psWipeTower, psGCodeExport}` whenever `old_instance->printable != new_instance->printable`. This restores the invalidation contract that existed prior to `e22d328fa` for printability changes specifically, without re-introducing the perf cost of unconditional invalidation on translation/rotation. Skirt/brim is included because the set of `PrintInstance`s feeding `Print::_make_skirt` is filtered upstream by `ModelInstance::is_printable()` at `PrintApply.cpp:147` — when printability toggles, the cached skirt geometry is stale.

adele-with-a-b:
## Summary

Fix a regression where toggling an object instance's "Printable" flag leaves the slice button grayed out, requiring an unrelated change to the project before the user can re-slice.

Closes #10891

## Root cause

`Print::apply()`'s instance synchronization branch (`src/libslic3r/PrintApply.cpp` ~line 1757) updates each instance's `printable` flag but only invalidates Print steps when `is_printable_filament_changed()` detects a filament-set change. A pure printability toggle leaves the filament assignment unchanged, so this loop produces no invalidation.

Section 4 of `Print::apply` does fire `invalidate_all_steps()` for the toggled object, but in the affected workflow the print steps are already in INVALID state from earlier background-process activity, so `invalidate_all_steps()` returns false and the overall `apply_status` caps at `APPLY_STATUS_CHANGED` (1) instead of `APPLY_STATUS_INVALIDATED` (2).

`Plater::priv::update_background_process` only calls `update_slice_result_valid_state(false)` on `APPLY_STATUS_INVALIDATED` ([Plater.cpp:10210](https://github.com/bambulab/BambuStudio/blob/master/src/slic3r/GUI/Plater.cpp#L10210)). With `APPLY_STATUS_CHANGED`, the cached slice result stays valid, so `MainFrame::get_enable_slice_status()` returns false (button disabled) because `PartPlate::is_slice_result_valid()` is still true.

## Regression

Introduced by [`e22d328f`](https://github.com/bambulab/BambuStudio/commit/e22d328fa9d1682c00617a4cad0498c67a7512f4) ("ENH: Optimizing the shape invalid condition for wipe tower of multi-extruder", May 2025), which replaced an unconditional invalidation in the `set_instances` path with a narrower convex-hull-based check. The optimization was correct for translation/rotation cases, but printability toggles do not change geometry, so they fall through both the new and old checks in this loop.

## Fix

Invalidate `{psSkirtBrim, psWipeTower, psGCodeExport}` whenever `old_instance->printable != new_instance->printable`. The triplet matches the one used elsewhere in `PrintApply.cpp` (line 1852) and in `PrintObject::set_instances` ([PrintObject.cpp:116](https://github.com/bambulab/BambuStudio/blob/master/src/libslic3r/PrintObject.cpp#L116)).

Skirt/brim is included because the set of `PrintInstance`s feeding `Print::_make_skirt` is filtered upstream by `ModelInstance::is_printable()` at [`PrintApply.cpp:147`](https://github.com/bambulab/BambuStudio/blob/master/src/libslic3r/PrintApply.cpp#L147) — when printability toggles, the cached skirt geometry is stale.

The diff is 4 lines.

## Test plan

- [x] Verified with `BBL_INTERNAL_TESTING=1` build that pre-patch toggle produces `apply_status=1` (CHANGED) and `enable_slice=0`, while post-patch the same toggle produces `apply_status=2` (INVALIDATED) and `enable_slice=1`.
- [x] Toggle off, slice button enables, slice; toggle on, slice button enables, slice. Confirmed across 7 toggle events on a multi-object plate.
- [x] No regression in the non-toggling case: translation/rotation/transform changes still take the `is_printable_filament_changed` path and behave as `e22d328f` intended.
- [ ] Verified on multi-extruder H2C profile (the configuration where the bug was originally reported).

### DIFF
--- a/src/libslic3r/PrintApply.cpp
+++ b/src/libslic3r/PrintApply.cpp
@@ -1757,6 +1757,10 @@ Print::ApplyStatus Print::apply(const Model &model, DynamicPrintConfig new_full_
 	        	// Synchronize the content of instances.
 	        	auto new_instance = model_object_new.instances.begin();
 				for (auto old_instance = model_object.instances.begin(); old_instance != model_object.instances.end(); ++ old_instance, ++ new_instance) {
+                    // BBS: printable toggle changes which instances generate G-code; invalidate skirt/brim, wipe tower, and G-code.
+                    if ((*old_instance)->printable != (*new_instance)->printable) {
+                        update_apply_status(this->invalidate_steps({psSkirtBrim, psWipeTower, psGCodeExport}));
+                    }
                     if (is_printable_filament_changed(new_full_config, (*old_instance)->convex_hull_2d(), (*new_instance)->convex_hull_2d())) {
                         update_apply_status(this->invalidate_steps({psWipeTower, psGCodeExport}));
                     }


BenJule:
Heads-up: this build failure doesn't come from this PR. The `Build All (macos-*) / Build Deps` jobs fail while building `dep_Assimp`, not on your slicer change.

Assimp bundles an old zlib (`contrib/zlib`) whose K&R-style function declarations no longer parse against the macOS 15+/26 SDK, so the deps build currently breaks on every macOS PR. I've proposed a one-line fix in #11051 (use the system zlib for Assimp on macOS/Linux). Once that lands, a rebase or re-run here should go green.

XunZhangBambu:
Thanks for this fix. Your analysis is correct. LGTM

XunZhangBambu (approved):

adele-with-a-b:
@XunZhangBambu thanks again for the approval! The only thing keeping this from merging is the `Build All (macos-*) / Build Deps` job failing while building `dep_Assimp` — Assimp's bundled zlib redefines `fdopen` on the macOS 15+ SDK, which is unrelated to this small slicer change. Is there anything needed on my end, or is this good to merge whenever you get a chance?

adele-with-a-b:
@XunZhangBambu this one is approved and shows as mergeable, but it's held by a red macOS check that is no longer real.

The two failing jobs — `Build All (macos-15-intel, x86_64) / Build Deps` and `Build All (macos-15, arm64) / Build Deps` — come from a run started **2026-06-09**, back when Assimp's bundled zlib wouldn't compile against the macOS 15+ SDK and every macOS PR was failing on it. That has since been resolved upstream: as of 2026-09-06, my other open PRs build cleanly on `macos-15-intel`, `macos-15 arm64` and `macos-26 arm64`. Nothing in this PR touches the deps.

Could someone re-run CI here? I'd rather not force-push to refresh it, because that risks dismissing your approval for a reason unrelated to the change — but if you'd prefer a rebase onto current `master` I'm glad to do that instead, just say so.

PR #11051, opened 2026-06-06
fix(deps): use system zlib for Assimp on macOS/Linux

BenJule:
## Problem
The dependency build fails on macOS while building Assimp — see the failing run [`Build All (macos-15, arm64) / Build Deps`](https://github.com/bambulab/BambuStudio/actions/runs/26564990881/job/78257304925):

    [1/119] Building C object contrib/zlib/CMakeFiles/zlibstatic.dir/zutil.c.o
       22 | int ZEXPORT compress2(dest, destLen, source, sourceLen, level)
    error: expected ')'
    ninja: build stopped: subcommand failed.

Assimp bundles an old copy of zlib (`contrib/zlib`) that still uses K&R-style function declarations and an `fdopen()` macro. These no longer parse against the macOS 15+/26 SDK headers, so `ASSIMP_BUILD_ZLIB=ON` breaks the macOS dependency build.

## Fix
Build Assimp against the **system zlib** on macOS and Linux (always available there), keeping the bundled copy only on Windows. One-file change, no behavior change on Windows.

PR #11052, opened 2026-06-06
fix(macos): guard WebView async callback against use-after-free (macOS 26.5 crash/hang)

### LINKED ISSUE #10968: Very long loading time since 2.7.1.57
### Bambu Studio Version

2.7.1.57

Bambu Lab Official website

### OS version

MacOS 26.5.1 (previous 26.5)

_No response_

### Printer

Bambulab X2D

### How to reproduce

Load the bambulab software.
On the home screen, wait for au long time for loading online models button to showup and the other menues. 
Can't use the software before all the menues have been loaded. This was not the case with the previous bambulab version.

### Actual results

after waiting for loading the software works pretty well but when coming back to home screen there can be lattency

### Expected results

immediate loading as it was in the previous stable version

none

- [ ] Log file
- [ ] Project file

### LINKED ISSUE #11004: Crash on launch: PAC failure in WebView::CreateWebView (macOS 26.5.1, ARM64)
## Description

BambuStudio 2.7.1.57 crashes on launch on macOS 26.5.1 (Apple Silicon M1). The app launches but crashes within ~21 seconds, before the UI becomes usable. 

This is a more severe manifestation of the same root cause as #10968 — where most users experience a long hang during WebView initialization, some users (including myself and @spoco2 in that thread) hit an outright crash.

## Environment

- **BambuStudio**: 02.07.01.57 (official release)
- **macOS**: 26.5.1 (25F80)
- **Hardware**: MacBookAir10,1 (M1, ARM64)
- **Safari/WebKit**: 26.5 (21624.2.5.11.4)

## Crash Details

**Exception**: `EXC_BAD_ACCESS (SIGSEGV)`  
**Subtype**: `KERN_INVALID_ADDRESS at 0xa53a6f61dbbf3a83 → possible pointer authentication failure`

The crash is an Apple Silicon **Pointer Authentication Code (PAC) failure** — a lambda callback scheduled via `wxAsyncMethodCallEventFunctor` attempts to call a method on a `wxWebView*` that has already been destroyed (use-after-free).

**Thread 0 (main thread) crashing frames:**

```
0  BambuStudio  WebView::CreateWebView(...)::$_0::operator()(wxWebView*) const + 240
1  BambuStudio  wxAsyncMethodCallEventFunctor<...>::Execute() + 156
2  BambuStudio  wxEvtHandler::ProcessEventLocally(wxEvent&) + 172
3  BambuStudio  wxEvtHandler::ProcessEvent(wxEvent&) + 56
4  BambuStudio  wxEvtHandler::ProcessPendingEvents() + 236
...
```

The address `0xa53a6f61dbbf3a83` has an invalid PAC signature. Stripped to bare address `0x6f61dbbf3a83`, it is "not in any VM region" — confirming this is a dangling pointer.

## Connection to #10968

Issue #10968 describes the same `WebView::CreateWebView → RunScript → YieldFor` call path blocking the main thread during startup. In most cases this results in a long hang (30s+, memory climbing to 1GB+); in this case the WebView object is destroyed before the async callback executes, triggering the PAC failure on ARM64.

The minimize/maximize workaround from #10968 does not apply here because the app crashes before the UI window is responsive.

## Full Crash Report

Complete macOS crash report (704 lines, including all thread backtraces, register state, and VM region summary):
https://gist.github.com/E-R-Butch/d9661f29c43fc86b86ab8a0fc8d40049

BenJule:
## Problem

On macOS 26.5 / 26.5.1, BambuStudio either **hangs for a long time on startup** (#10968) or **crashes within ~20 s of launch** (#11004), with steadily growing memory use in both cases. The crash is an Apple-Silicon pointer-authentication failure:

    Exception: EXC_BAD_ACCESS (SIGSEGV)
    Subtype:   KERN_INVALID_ADDRESS ... possible pointer authentication failure

    0  BambuStudio  WebView::CreateWebView(...)::$_0::operator()(wxWebView*) const + 240
    1  BambuStudio  wxAsyncMethodCallEventFunctor<...>::Execute() + 156

## Likely cause

`WebView::CreateWebView` registers the `"wx"` script message handler from a `CallAfter()` callback. The callback captures the **raw `webView` pointer** and calls `AddScriptMessageHandler()` on it.

The stack (a `CreateWebView` lambda invoked from `wxAsyncMethodCallEventFunctor::Execute`, failing with a PAC error on a corrupted pointer) is consistent with the WebKit view being torn down before this async callback runs, so the callback would dereference a freed `wxWebView*` (use-after-free). I am **inferring** the use-after-free from the stack and the PAC failure — I have not confirmed it with a heap / ASan trace — but the captured raw pointer makes it possible and it is the most likely explanation. On Apple Silicon a corrupted pointer trips pointer authentication → hard crash; elsewhere the same path surfaces as the long startup hang.

The same hazard applies to `g_delay_webviews`, which holds raw pointers that are never cleared when a view is destroyed.

## Fix

- In the `CallAfter` lambda, check `g_webviews` (the list of live views) and bail out if `webView` has already been destroyed.
- In `~WebViewRef`, also remove the view from `g_delay_webviews`, so a later flush of the delayed list can never touch a freed view.

Small, self-contained change: no effect on Windows (the block runs synchronously there) and no functional change on the healthy path. The guards are defensive and correct regardless of whether the use-after-free is the exact mechanism.

Fixes #11004
Fixes #10968

### DIFF
--- a/src/slic3r/GUI/Widgets/WebView.cpp
+++ b/src/slic3r/GUI/Widgets/WebView.cpp
@@ -233,6 +233,12 @@ public:
         assert(iter != g_webviews.end());
         if (iter != g_webviews.end())
             g_webviews.erase(iter);
+        // Also drop it from the delayed list so a pending flush of g_delay_webviews
+        // never calls AddScriptMessageHandler() on an already-destroyed view.
+        // See bambulab/BambuStudio #11004 and #10968.
+        auto diter = std::find(g_delay_webviews.begin(), g_delay_webviews.end(), m_webView);
+        if (diter != g_delay_webviews.end())
+            g_delay_webviews.erase(diter);
     }
     wxWebView *m_webView;
 };
@@ -344,6 +350,14 @@ wxWebView* WebView::CreateWebView(wxWindow * parent, wxString const & url)
         };
 #ifndef __WIN32__
         webView->CallAfter([webView, addScriptMessageHandler] {
+            // This async callback may fire after webView has already been destroyed,
+            // which would call AddScriptMessageHandler() on a dangling pointer
+            // (use-after-free -> pointer-authentication crash on Apple Silicon, or a
+            // long hang during startup on macOS 26.5+). g_webviews lists every live
+            // view, so bail out if this one is already gone.
+            // See bambulab/BambuStudio #11004 and #10968.
+            if (std::find(g_webviews.begin(), g_webviews.end(), webView) == g_webviews.end())
+                return;
 #endif
             if (Slic3r::GUI::wxGetApp().is_adding_script_handler()) {
                 g_delay_webviews.push_back(webView);


BenJule:
Note on the failing **`Build All (macos-15, arm64) / Build Deps`** check: this failure is **not** caused by this PR.

This PR only touches `src/slic3r/GUI/Widgets/WebView.cpp` — it changes no dependencies — and the build never reaches the source step; it fails earlier while building **Assimp's bundled zlib**:

```
FAILED: contrib/zlib/CMakeFiles/zlibstatic.dir/zutil.c.o
.../MacOSX15.5.sdk/usr/include/_stdio.h:318:7: error: expected identifier or '('
```

Assimp's vendored `contrib/zlib` defines an `fdopen` macro that no longer parses against the macOS 15+ SDK headers, so `ASSIMP_BUILD_ZLIB=ON` breaks the dependency build on macOS. This is a pre-existing failure on `master` that currently affects **every** PR's macOS deps build.

It is fixed by #11051 (`fix(deps): use system zlib for Assimp on macOS/Linux`). Once that lands, this PR's macOS CI should pass on a rebase — the two changes are independent.

tonghao-bbl (commented):
LGTM @lanewei120

lanewei120:
thanks

PR #11066, opened 2026-06-07
fix(gcode): report the real OS error when G-code export fails

BenJule:
## Problem

When G-code export fails, BambuStudio always shows:

> G-code export to &lt;file&gt; failed. Is the disk full?

This is **guessed**, not measured. `GCodeOutputStream` only checks `ferror()` and `GCode::export_gcode()` discards the real `errno` from the failed `fwrite`/`fflush`. Large multi-colour models produce multi-GB G-code, and the write can fail for several different reasons — a full **temp** volume, a read-only path, an antivirus/cloud-sync lock, or a disk quota — all of which are reported identically as 'disk full'. Users (this came from a real 4-colour model export on Windows) are left guessing.

## Fix

- Capture `errno` at the first failed `fwrite`/`fflush` in `GCodeOutputStream`.
- Expose it via `get_last_error()` (`strerror`).
- Include the real cause in the export-failure message, e.g. *'No space left on device'* vs *'Permission denied'*, while keeping the original disk-full hint as a fallback when no `errno` was recorded.

No behavioural change other than a more accurate error message. Small, self-contained: `src/libslic3r/GCode.cpp` + `GCode.hpp` (+25/-3).

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -18,6 +18,8 @@
 
 #include <algorithm>
 #include <cstdlib>
+#include <cerrno>
+#include <cstring>
 #include <chrono>
 #include <math.h>
 #include <utility>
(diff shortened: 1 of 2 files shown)


Diatom-Bambu (approved):
Thanks for your submission; this certainly makes the error message more detailed.

Diatom-Bambu (approved):

XunZhangBambu (approved):

lanewei120:
thanks @BenJule

PR #11067, opened 2026-06-07
feat: fall back to user-data volume when the temp dir is low on free space

BenJule:
## Problem

The project backup/scratch directory — which holds the per-plate G-code temp file written during slicing — is created under `temporary_dir()` (the system temp dir). On Linux `/tmp` is frequently a **RAM-backed tmpfs** (commonly sized at half of RAM, e.g. ~8 GB). A large multi-colour model produces multi-GB G-code (prime tower + flushing), which **overflows the tmpfs** and makes export fail with:

> G-code export to … failed. Is the disk full?

…even though the real disk has hundreds of GB free. Reported from a real 4-colour print on a Debian host where `/tmp` is a 7.8 GB tmpfs.

## Fix

`Model::get_backup_path()` now checks the free space on the temp volume and, when it has **less than 2 GiB** free, falls back to `<data_dir>/tmp` — which lives on the user-data volume (real disk) — **if** that offers more room.

- Best-effort and fully guarded: any filesystem error leaves the default temp dir in place (slicing behaviour unchanged on error).
- The fast tmpfs is still used whenever it has enough space; the fallback only triggers under genuine pressure.
- Platform-neutral: on systems where the system temp already lives on a large disk (typical on Windows/macOS), the check simply never triggers.

Small and self-contained: `src/libslic3r/Model.cpp` (+24). Complements a separate change that makes the export error report the real errno instead of always guessing 'disk full'.

### DIFF
--- a/src/libslic3r/Model.cpp
+++ b/src/libslic3r/Model.cpp
@@ -1050,6 +1050,30 @@ std::string Model::get_backup_path()
     {
         auto pid = get_current_pid();
         boost::filesystem::path parent_path(temporary_dir());
+        // Guard against a system temp dir that is low on free space — most often a
+        // small RAM-backed tmpfs /tmp on Linux, which a multi-GB multi-colour
+        // G-code export can overflow (surfaced to the user as "Is the disk full?").
+        // When the temp volume is nearly full, fall back to a folder on the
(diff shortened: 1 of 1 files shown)


tonghao-bbl:
LGTM @lanewei120

Issue #10354, opened 2026-04-24
Material info not displayed at Material selection on objects view


tompollok:
### Bambu Studio Version

2.6.0.51

### Where is the application from?

Bambu Lab Official website

### OS version

MacOS

### Additional system information

Please add textual material information to the material selection in objects view. The space made is wide enough an in all other places you have the text visible. 

Feature: Please also add info whether it is Left or Right extruder. maybe with an L or R.

See Screenshot for info. The bug is at least 6 months old and it is pretty annoying.

<img width="1028" height="582" alt="Image" src="https://github.com/user-attachments/assets/6e4bb92a-5fd3-4cee-9ae2-0e7f438920bf" />

### Printer

All bambu printers

### How to reproduce

Use MacOS
Load object files. 
Click on objects view
click on a single object and click at material assignment for that object.
You will see there is no info text. In other places with the same UI element you show the color, number and info.

### Actual results

<img width="1028" height="582" alt="Image" src="https://github.com/user-attachments/assets/01dde966-65a3-494f-ab73-504657af86f4" />

### Expected results

<img width="1028" height="582" alt="Image" src="https://github.com/user-attachments/assets/7a2e419b-3681-4e38-96e4-416637257dbf" />

### Project file & Debug log uploads

no log

### Checklist of files to include

- [ ] Log file
- [ ] Project file

rouyuanzi-bambu:
Hello, after review, this has been identified as a feature request and has now been converted into a request ticket for further evaluation.

tompollok:
I reported as bug because given the already wide selection view with white unused area, it seemed like it was supposed to show more info but it looks like it isn’t “visible” due to a bug

tompollok:
@rouyuanzi-bambu @JohnSmithCYM as nothing happened yet I resolved the issue that it wasn't displaying the text and nozzle suffix and tested on Mac.

<img width="1724" height="1005" alt="Image" src="https://github.com/user-attachments/assets/cc5da455-87bc-4f9f-8e67-cb4587020892" />

It also fixes the issue on support material selection in the same way.

<img width="774" height="647" alt="Image" src="https://github.com/user-attachments/assets/ac2b209b-e234-407c-85a5-8e1025fc6c2a" />

When the printer is just a single nozzle printer, then it does not display the (L) or (R ) nozzle suffix.

tonghao-bbl:
resolved by #11221

PR #11221, opened 2026-06-20
Show filament name and left/right nozzle in filament selectors

tompollok:
Addresses #10354 (Material info not displayed at Material selection on objects view).

## What
Across the filament-selection UIs, show the filament **display name** and, on dual-nozzle printers (H2D/H2C), a right-aligned **(L)/(R)** indicating which nozzle the filament is loaded on:
- Object list **"Fila."** column and its dropdown editor
- **"Change Filament"** context menus
- **Support/raft base** and **Support/raft interface** selectors

## Why
On dual-nozzle machines it isn't obvious from these menus which nozzle a filament is on, and the object-view material selection showed little material info. This surfaces the filament name and nozzle side consistently, matching the sidebar Left/Right Nozzle view.

## How
- During AMS sync, write the real per-filament nozzle into `filament_map` (both the overwrite and merge paths) instead of defaulting everything to the left nozzle. The nozzle is derived from each AMS's bound extruder / current filament-switcher position, the same way the Left/Right Nozzle view computes it.
- Skip empty external-spool slots in `build_filament_ams_list` so they no longer add phantom filaments (which previously shifted the assignment).
- Add an opt-in right-aligned text column to the `DropDown` widget so the (L)/(R) labels line up vertically; other dropdowns are unaffected.
- Object-tree filament cell tooltip shows the readable color name + AMS location.

## Notes
- The (L)/(R) only shows on dual-nozzle printers once filaments are synced; single-nozzle printers are unchanged.
- Tested on macOS with a connected H2C (3× AMS + 3× AMS-HT, 2 left / 13 right).

### DIFF
--- a/src/libslic3r/PresetBundle.cpp
+++ b/src/libslic3r/PresetBundle.cpp
@@ -2560,6 +2560,11 @@ unsigned int PresetBundle::sync_ams_list(std::vector<std::pair<DynamicPrintConfi
     std::vector<std::string> ams_filament_colors;
     std::vector<std::string> ams_filament_color_types;
     std::vector<AMSMapInfo>  ams_array_maps;
+    // BBS: nozzle assignment per built filament (1 = left, 2 = right), kept in lockstep with
+    // ams_filament_presets. The nozzle is encoded in the filament_ams_list key (bit 0x10000
+    // => right/main nozzle, else left/deputy), so we capture it here instead of defaulting all
+    // filaments to the left nozzle.
+    std::vector<int>         ams_filament_nozzles;
     ams_multi_color_filment.clear();
     BOOST_LOG_TRIVIAL(info) << __FUNCTION__ << boost::format(": filament_ams_list size: %1%") % filament_ams_list.size();
     struct AmsInfo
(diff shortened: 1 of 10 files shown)


Haidiye00 (commented):
Hi, please submit the recording effect of the gif

tompollok:
@Haidiye00 images of how it looks now are at the very bottom of https://github.com/bambulab/BambuStudio/issues/10354

tonghao-bbl:
Hi @tompollok Thanks for your contribution and I checked #10354 , it looks nice.
image

Before looking into the code, I have several questions:
* some filaments only have `PLA` is displayed, I think it should be something like  `PLA Basic`, why is this?
* Only synced filaments know which nozzle it will be feed into, the use added filaments have no nozzle info, how will the `L` and `R` display?

tompollok:
@tonghao-bbl i did not change the existing PLA naming scheme. the reason why it does not say Basic in that case is because that filament is manually set as "Generic PLA" and not BambuLab Filament. Maybe in that case it should also say "Generic PLA" or "PLA Generic" so that it visuall matches the pattern.

Regarding the L and R display. if it is not synced then L or R should not be displayed.

PR #10818, opened 2026-05-20
feat(gui): auto-save the open project every 5 minutes

BenJule:
## Summary

Auto-saves the open project every 5 minutes: a timer silently calls save when the project is dirty and already has a file path, gated behind the `autosave_enabled` config key.

(Reduced from the original three-feature PR as requested, see split note below.)

> **Note:** there is no Preferences toggle for `autosave_enabled` yet, so it is off unless the key is set. Happy to add a UI toggle if you'd like.

PR #11246, opened 2026-06-22
feat(gui): fine object-move steps (Alt 0.1 mm / Alt+Shift 0.01 mm)

BenJule:
Split out from #10818 (which the review asked to break into separate PRs).

## Summary
Arrow keys move the selection in 10 mm steps (1 mm with Shift). This adds two finer steps using **Alt**:
- **Alt + arrow** → 0.1 mm
- **Alt + Shift + arrow** → 0.01 mm

Implemented by threading the Alt state through `TranslationProcessor` into the move multiplier; the keyboard-shortcuts dialog documents the new steps. 2 files.


Haidiye00:
Hello, this shortcut operation seems too niche. Can it be set directly through the UI?

Haidiye00 (changes requested):
Hello, this shortcut operation seems too niche. Can it be set directly through the UI?

BenJule:
Fair question. To clarify the scope: this isn't a brand-new shortcut, it's a modifier on the **existing** arrow-key nudge. Today arrow keys already move the selected object by a fixed step; this just lets Alt / Alt+Shift make that same nudge finer (0.1 mm / 0.01 mm) for precise placement, which matches the modifier-for-fine-step convention in most CAD/slicer tools.

That said, I'm happy to go either way:
- keep it as the modifier-based nudge (minimal, discoverable via tooltip/docs), or
- additionally expose the move step as a small numeric field in the move/arrange UI so it's also settable without the keyboard.

Which would you prefer? If you'd rather have the UI field, I'll add it; if the feature feels too niche overall, I can close this.

BenJule:
Added the UI field as suggested. There's now an **"Object move step"** input on the General preferences page (default 10 mm, range 0.01-100 mm). The arrow keys read that value as the base nudge, so the step is fully settable from the UI without any shortcut.

The Shift / Alt modifiers are still there but are now purely *relative* refinements on top of the configured step (Shift = x0.1, Alt = x0.01), so they're a convenience rather than the only way to get a fine step. With the default of 10 mm this reproduces the previous 10 / 1 / 0.1 / 0.01 mm behaviour exactly, so nothing changes for existing users.

Files: `Preferences.cpp` (the field), `AppConfig.cpp` (default), `GLCanvas3D.cpp` (handler reads the config), `KBShortcutsDialog.cpp` (help text now describes the steps relative to the configured move step).

Haidiye00:
Hi, could you merge this into one patch and then submit the related test GIFs? Is that okay? Thanks.

BenJule:
Done, squashed into a single commit.

On the GIFs: I have to be upfront that I don't have a macOS/Windows GUI build environment here, so I can't record an actual GIF of it in action without overstating things. What the change does is testable directly, though:

- Open Preferences → General → "Object move step" and set a value (default is 10 mm).
- Select an object and press the arrow keys: it moves by that step. Hold Shift for x0.1 and Alt for x0.01 of it.
- With the default 10 mm this is exactly the old 10 / 1 / 0.1 / 0.01 mm behaviour, so nothing regresses for existing users; the field just makes the base step adjustable from the UI.

If a recorded GIF is required to merge, I'd need a hand from someone with a GUI build, or I'm glad to walk through any specific case. And if you'd prefer a different default or placement for the field, just say and I'll adjust.

Haidiye00:
Hello. I strongly recommend that you set up a Windows GUI test environment. A Windows laptop or desktop will suffice. A moderately powerful machine is recommended to avoid slow compilation. This will allow us to verify your development results and carry out acceptance testing smoothly. Thank you.

BenJule:
That's fair, and thanks for the nudge. You're right that a proper GUI build is the only way to do real acceptance testing and produce the GIFs you're asking for. I'll set up a Windows build/test environment so future GUI PRs come with actual on-device verification and recordings instead of just code-traced reasoning. Appreciate the patience in the meantime.

BenJule:
Quick status so this doesn't look stale: the code side is settled. The design feedback is addressed (the move step is now a real Preferences field, default 10 mm, and the Shift/Alt modifiers are just relative refinements on top of it, so default behaviour is unchanged). I also double-checked the two small things that could look off, the Preferences string layout and the atof read, and both match the existing sibling entries on that page, so there's nothing to change there.

The one thing still open is the recorded verification you asked for. I'd rather do that properly on a Windows build than hand you a Linux clip, so it'll come once I have that environment set up rather than me guessing. Happy to hold the PR until then.

PR #11251, opened 2026-06-22
fix: restore custom bed texture after undo/redo (#11211)

### LINKED ISSUE #11211: custom bed PNG removed after klik undo on modified object
### Bambu Studio Version

2.7.1.62

Bambu Lab GitHub releases

### OS version

windows 10

_No response_

### Printer

custom 3d print

### How to reproduce

1. add stl file
2. copy paste stl file
3. add letter
4. undo
5. custom bed background removed

### Actual results

it jus happened when klik "undo" button

### Expected results

the background bed not have been effected

[log apk.zip](https://github.com/user-attachments/files/29119217/log.apk.zip)

- [x] Log file
- [ ] Project file

BenJule:
## Summary
The custom bed (logo) texture disappears after an undo (#11211).

**Root cause:** the plate logo texture filename (`PartPlate::m_logo_texture_filename`) is not part of the undo snapshot, `PartPlateList::serialize` does not include it and the per-plate `serialize` is commented out, so when the plates are rebuilt on deserialize the texture filename is lost. `update_after_undo_redo()` never re-applies it.

**Fix:** re-apply the bed texture from the config (which is *not* part of the model undo stack and therefore survives the undo) in `update_after_undo_redo()` via `set_bed_shape()`. Its internal guard only re-sets the logo texture filename when the bed shape is unchanged (the common case for a model-level undo), so this is cheap and does not rebuild the bed geometry.

> ⚠️ I can't build/run locally, so this is **untested at runtime**, the reasoning is from the code path. A quick check (load a custom bed texture, add/copy an object, undo → texture should remain) would be appreciated.

Fixes #11211

### DIFF
--- a/src/slic3r/GUI/Plater.cpp
+++ b/src/slic3r/GUI/Plater.cpp
@@ -16935,6 +16935,12 @@ void Plater::priv::update_after_undo_redo(const UndoRedo::Snapshot& snapshot, bo
 
     wxGetApp().obj_list()->update_after_undo_redo();
 
+    // #11211: the custom bed (logo) texture is not part of the undo snapshot, so it is lost
+    // when the plates are rebuilt on deserialize. Re-apply it from the (undo-independent)
+    // config. set_bed_shape() only re-sets the logo filename when the bed shape itself is
+    // unchanged, so this is cheap in the common case.
+    q->set_bed_shape();
+
     if (wxGetApp().get_mode() == comSimple && model_has_advanced_features(this->model)) {
         // If the user jumped to a snapshot that require user interface with advanced features, switch to the advanced mode without asking.
         // There is a little risk of surprising the user, as he already must have had the advanced or advanced mode active for such a snapshot to be taken.


Haidiye00:

Haidiye00 (changes requested):

BenJule:
Here's the before/after (#11211):

**Before:** set a custom bed (logo) texture, then do any edit and press Undo (or Redo). The custom bed texture **disappears** and the bed falls back to the default - the logo is gone until you reload the project.

**After:** the custom bed texture is preserved across undo/redo.

**Root cause:** the custom bed/logo texture filename isn't part of the undo snapshot (it's not serialized by `PartPlateList::serialize`), so restoring a snapshot dropped it. The fix re-applies the bed shape (which re-reads `bed_custom_texture` from config) in `update_after_undo_redo`, so the logo is restored after each undo/redo.

Repro: load a custom bed texture → move/edit an object → Undo → (before) texture gone / (after) texture stays.

Same caveat as my other GUI PRs: no build environment here, so this is traced from code rather than a recorded GIF. Can add a capture if a maintainer build is handy.

PR #11252, opened 2026-06-22
fix: settings search jumps to per-extruder/filament options (#11200)

### LINKED ISSUE #11200: setting search
### Bambu Studio Version

v02.07.01.62

Bambu Lab Official website

### OS version

windows 11

_No response_

### Printer

Bambulab H2D

### How to reproduce

Search 'filament' in the process search bar. a few suggestions come up but when i click on them, they are not there.

I've tried it with develop mode active and inactive.

### Actual results

a few suggestions come up but when i click on them, they are not there.

### Expected results

I should be able to access all settings. I purposely tried bambu studio so i could concentrate on designing and prototyping and fiddling with slicer settings. I may as well be back on orca.  

?

- [ ] Log file
- [ ] Project file

BenJule:
## Summary
Clicking a settings-search suggestion for a per-extruder/filament (vector) option does nothing, the setting is never shown (#11200).

**Root cause:** the searcher keys vector options with a `#<variant>` suffix (e.g. `filament_xxx#0`), and `Search::Option::opt_key()` keeps that suffix. But `Tab::activate_option` matches against the **bare** option key (`o.opt.opt_key == opt_key`), so the suffixed key never matches and the option/page is not activated. Scalar options (no `#` suffix) jump fine, which is why only some results "aren't there".

**Fix:** strip the `#<variant>` suffix in `Sidebar::jump_to_option` before locating/activating the option. `substr(0, find('#'))` returns the whole key when there is no `#`, so the many direct callers that pass bare keys are unaffected.

> ⚠️ Untested at runtime (I can't build locally), reasoning is from the code path. A quick check (search e.g. "filament" in the process settings search, click a result → it should now scroll to that setting) would be appreciated.

Fixes #11200

### DIFF
--- a/src/slic3r/GUI/Plater.cpp
+++ b/src/slic3r/GUI/Plater.cpp
@@ -3924,9 +3924,15 @@ void Sidebar::search()
     p->searcher.search();
 }
 
-void Sidebar::jump_to_option(const std::string& opt_key, Preset::Type type, const std::wstring& category)
+void Sidebar::jump_to_option(const std::string& opt_key_with_idx, Preset::Type type, const std::wstring& category)
 {
     //const Search::Option& opt = p->searcher.get_option(opt_key, type);
+    // The settings search keys vector options with a "#<variant>" suffix (e.g.
+    // "filament_xxx#0"), but the option groups are keyed by the bare option key, so
+    // strip the suffix before locating/activating the option, otherwise the jump
+    // silently fails for per-extruder/filament settings (#11200). substr(0, npos)
+    // returns the whole string when there is no '#', so bare keys are unaffected.
+    const std::string opt_key = opt_key_with_idx.substr(0, opt_key_with_idx.find('#'));
     if (type == Preset::TYPE_PRINT) {
         auto tab = dynamic_cast<TabPrintModel*>(wxGetApp().params_panel()->get_current_tab());
         if (tab && tab->has_key(opt_key)) {


Haidiye00:

Haidiye00 (changes requested):

BenJule:
Here's the before/after (#11200):

**Before:** open the settings search and search for a per-extruder or per-filament option (any option whose key carries a `#<index>` suffix, e.g. a filament/extruder-indexed parameter). Click the result - nothing happens, the view does **not** jump to that option. The search match is silently dropped.

**After:** clicking the same result scrolls to and highlights the correct option.

**Root cause:** `Search::Option::opt_key()` returns the key *with* its `#0` index suffix, but `Tab::activate_option` compares against the bare key (`o.opt.opt_key == opt_key`), so the per-variant options never matched. The fix strips the `#<idx>` suffix in `jump_to_option` before the comparison.

Repro: settings search → type a filament-indexed option name → click result → (before) no jump / (after) jumps to it.

Note: I don't have a GUI build environment here, so I can't attach a recorded GIF - the above is traced from the code path. Happy to add one if a maintainer build is available.

BenJule:
Sure. Before the fix, clicking a search suggestion for a per-extruder/filament option simply did nothing, the settings page never opened (that's what #11200 is about, scalar options were fine). With the fix the page opens and the option gets highlighted as expected:

demo

tonghao-bbl:
Hi @BenJule Sorry for the late reply, as discussed in #11200 , we can't reproduce this issue.
The code key with index can be handled by underline searching system withtou any issue.

So, we'll close this PR now, Thanks for your time.

Issue #11253, opened 2026-06-22
Save / Save As / import default location incorrect.


Rogueohio:
### Bambu Studio Version

2.7.1.62

### Where is the application from?

Bambu Lab Official website

### OS version

windows 11

### Additional system information

When I import a 3mf file the default location is opening as:
C:\Users\......\AppData\Local\Temp\Neutron

This is the same default location when I save and save as.

### Printer

na

### How to reproduce

1. open studio
2. File>import>import 3mf... (or other option, import configs)

### Actual results

Strange file location opens.

### Expected results

not sure

### Project file & Debug log uploads

na

### Checklist of files to include

- [ ] Log file
- [ ] Project file

PR #11256, opened 2026-06-22
fix: open file dialogs in a sensible default directory instead of the working directory

BenJule:
Several `wxFileDialog` call sites passed an **empty directory**, so they opened in the current working directory (which can be a temp folder) instead of somewhere useful.

**Relocation dialogs** (a file path is known) now open in that file's own folder, falling back to the last used directory:
- "Select a new file" (replace a volume's source file)
- "Please select a file" (locate a missing input on project load)

**Open dialogs** (no prior path) now open in the **last used directory**:
- Select a G-code file (`GUI_App`)
- Import bed shape / texture / model (`BedShapeDialog`)
- Select images (`StatusPanel`)
- Open CA certificate file (`PhysicalPrinterDialog`)

No functional changes — only the dialogs' initial directory. Adjacent to the temp-working-dir behaviour reported in #11253, though it does not by itself fix the Import-3mf / Save-As default there.

### DIFF
--- a/src/slic3r/GUI/BedShapeDialog.cpp
+++ b/src/slic3r/GUI/BedShapeDialog.cpp
@@ -547,7 +547,7 @@ void BedShapePanel::update_shape()
 // Loads an stl file, projects it to the XY plane and calculates a polygon.
 void BedShapePanel::load_stl()
 {
-    wxFileDialog dialog(this, _L("Choose an STL file to import bed shape from:"), "", "", file_wildcards(FT_STL), wxFD_OPEN | wxFD_FILE_MUST_EXIST);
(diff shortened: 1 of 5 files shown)


Haidiye00:
Hi, please merge it into one patch and then submit your test results. Thanks.

BenJule:
Done - squashed into a single commit.

On the change itself: all the affected dialogs (open/import/export/replace across BedShapeDialog, PhysicalPrinterDialog, StatusPanel, Plater and GUI_App) now start in a sensible folder - the referenced file's own directory when there is one, otherwise the last-used directory - instead of the process working directory. The behaviour to check is: open one of these dialogs, pick a file in some folder, reopen the same dialog and confirm it starts in that folder again rather than the CWD.

Full disclosure: I don't have a macOS/Windows GUI build environment here, so I've verified the logic by tracing the code paths rather than capturing a recorded run. If a maintainer build is handy, those are the flows to confirm.

tonghao-bbl:
LGTM @lanewei120

lanewei120:
thanks @BenJule

PR #11291, opened 2026-06-25
fix: wait for macOS asset availability in Homebrew workflow

BenJule:
Adds a retry loop with timeout to wait for macOS universal DMG asset to become available before attempting to download its checksums. This resolves intermittent race conditions where the Homebrew workflow runs immediately after release publication before GitHub finishes indexing the release assets.


BenJule:
Please review and approve this PR. It adds a wait-for-asset loop to the Homebrew workflow to prevent race conditions. Thanks!

MackBambu:
We installed Bambu Studio via Homebrew, and it looks fine. It can be merged.
@lanewei120

lanewei120:
thanks @BenJule

PR #11323, opened 2026-06-27
ci: build and attach native Linux packages (.deb / .rpm / Arch) to releases

### LINKED ISSUE #11321: Publish native Linux packages (.deb / .rpm / Arch) alongside the AppImage in releases
### What I'd like

Right now the official Linux releases only ship the two AppImages (ubuntu22.04 / ubuntu24.04). Would you consider also publishing native distro packages alongside them:

- **`.deb`** for Debian / Ubuntu / Mint / Pop!_OS
- **`.rpm`** for Fedora / RHEL / openSUSE
- **`.pkg.tar.zst`** for Arch / Manjaro / EndeavourOS

### Why

The AppImage works, but it has the usual downsides: no integration with the system package manager (no `apt`/`dnf`/`pacman` updates), no automatic desktop/MIME registration, sandbox/FUSE quirks on some distros, and it bundles its own libraries instead of using the system ones. A lot of Linux users would rather install a native package that updates with the rest of their system. Today they're left building from source or relying on unofficial packages (e.g. #8967).

### Offer to help

I've already got CI building all three formats on my fork, so the packaging recipes exist and produce installable artifacts (deb/rpm/arch). I'm happy to contribute the workflows upstream or open a PR so these can be attached to the release automatically, if that's something you'd be open to. Just let me know the preferred approach (attach to the GitHub release, a separate repo/PPA/COPR/AUR, etc.) and I'll put a PR together.

Thanks!

BenJule:
## Summary

Closes #11321. Publishes native Linux packages alongside the AppImage on every release, so Debian/Ubuntu, Fedora/RHEL and Arch users can install with their own package manager and get desktop/MIME integration.

This builds **proper distro-native packages**, not an AppImage wrapper: BambuStudio is compiled from source inside each distribution's own container, so the GTK/GLib/X11/etc. libraries are declared as package dependencies and used from the system rather than bundled.

When a release is published (or on manual dispatch against a tag) it produces and attaches:

- `bambustudio_<ver>_amd64.deb` (Debian/Ubuntu)
- `bambustudio-<ver>-1.x86_64.rpm` (Fedora/RHEL/openSUSE)
- `bambu-studio-bin-<ver>-1-x86_64.pkg.tar.zst` (Arch)

## How it works

Each distro builds in its own container (`debian:trixie`, `fedora:42`, `archlinux:latest`) and reuses the existing build entry point:

1. `./BuildLinux.sh -u`, install build dependencies
2. `./BuildLinux.sh -d`, build the bundled dependencies (cached between runs)
3. `./BuildLinux.sh -s`, build the slicer
4. `BuildLinuxImage.sh -i`, assemble `build/package/` (binary + libs + resources)
5. one packaging script per format turns that tree into the native package

The three packaging scripts (`scripts/build_deb.sh`, `build_rpm.sh`, `build_pkg_arch.sh`) install the compiled binary under `/usr/lib/bambustudio`, ship a `/usr/bin` wrapper, a `.desktop` entry and icons, and deliberately **exclude** GTK/GLib/X11/Wayland/DBus/fontconfig/etc. from the bundled libs so the system copies are used (declared via `Depends`/`Requires`/`depends`). These are the same scripts I've been using on my fork, where the packages have been live and installable for a while.

## Test plan / notes

- It's a heavy job (full deps + slicer compile per distro), so it's gated to release publish + manual dispatch, with `deps/build/destdir` cached per distro to keep subsequent runs reasonable.
- I can't run this repo's Actions myself, so a maintainer **`workflow_dispatch` against the latest tag** is the way to validate end to end. The packaging scripts are proven on my fork's custom build images; the main thing to confirm here is that `BuildLinux.sh -u` provisions cleanly on the stock `debian:trixie` / `fedora:42` / `archlinux:latest` base images. Happy to switch to prebuilt dep images or adjust anything to match your conventions.
- External repo hosting (signed APT repo / PPA / COPR / AUR / Homebrew) is a separate follow-up since that needs per-org secrets and hosting decisions.

### DIFF
--- a/.github/workflows/release_linux_packages.yml
+++ b/.github/workflows/release_linux_packages.yml
@@ -1,0 +1,96 @@
+name: Release Linux native packages
+
+# Builds native Linux packages by compiling BambuStudio from source in each
+# distribution's own container (so the binary links against that distro's system
+# libraries), then uploads the resulting .deb / .rpm / .pkg.tar.zst to the
+# release. This is the "proper" native package path (not an AppImage wrapper):
+# the GTK/GLib/X11/etc. libraries are declared as package dependencies and used
+# from the system rather than bundled. See issue #11321.
+#
+# Heavy job (full deps + slicer build per distro), so it only runs on release
+# publish or manual dispatch, with the dependency build cached between runs.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to attach packages to (e.g. v02.07.01.62)'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 240
+    container:
+      image: ${{ matrix.image }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Debian .deb
+            image: debian:trixie
+            bootstrap: apt-get update -qq && apt-get install -y --no-install-recommends sudo git git-lfs curl ca-certificates xz-utils
+            package_script: scripts/build_deb.sh
+            glob: build/*.deb
+          - name: Fedora .rpm
+            image: fedora:42
+            bootstrap: dnf install -y sudo git git-lfs curl rpm-build xz
+            package_script: scripts/build_rpm.sh
+            glob: build/*.rpm
+          - name: Arch .pkg.tar.zst
+            image: archlinux:latest
+            bootstrap: pacman -Sy --noconfirm --needed sudo git git-lfs curl base-devel
+            package_script: scripts/build_pkg_arch.sh
+            glob: build/*.pkg.tar.zst
+    env:
+      TAG: ${{ github.event.release.tag_name || inputs.tag }}
+    steps:
+      - name: Bootstrap container
+        run: |
+          ${{ matrix.bootstrap }}
+          git config --global --add safe.directory '*'
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          lfs: true
+          fetch-depth: 0
+
+      - name: Cache built dependencies
+        id: cache-deps
+        uses: actions/cache@v4
+        with:
+          path: deps/build/destdir
+          key: native-pkg-deps-${{ matrix.image }}-${{ hashFiles('deps/CMakeLists.txt', 'deps/**/*.cmake') }}
+
+      - name: Install build dependencies
+        run: ./BuildLinux.sh -u
+
+      - name: Build dependencies
+        if: steps.cache-deps.outputs.cache-hit != 'true'
+        run: ./BuildLinux.sh -d
+
+      - name: Build BambuStudio
+        run: |
+          export CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
+          ./BuildLinux.sh -s
+          # Assemble build/package/ (binary + libs + resources) that the
+          # packaging scripts consume.
+          (cd build && ./src/BuildLinuxImage.sh -i)
+
+      - name: Build native package
+        run: bash ${{ matrix.package_script }}
+
+      - name: Upload package to the release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.TAG }}
+          files: ${{ matrix.glob }}
+          fail_on_unmatched_files: true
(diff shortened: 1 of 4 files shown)


BenJule:
Updated: switched this from the lightweight fpm/AppImage-wrapper approach to a proper native build. It now compiles BambuStudio inside each distro's container and packages the compiled tree, so the GTK/GLib/X11/etc. libraries are real system dependencies instead of being bundled inside an AppImage. Heavier in CI, but the resulting .deb/.rpm/.pkg are genuine native packages.

YeeP79:
I wanted to try this on Arch before it lands, so I ran the workflow's steps in the same container images it uses. Everything below is from `archlinux:latest`, `fedora:42` and `debian:trixie` starting from clean images, against `master` at `12f17b06f`. Sharing what I hit, since none of it is obvious from reading the diff and CI doesn't run on fork PRs.

**Short version:** the packaging scripts themselves look sound — I got a valid `.pkg.tar.zst` out of `build_pkg_arch.sh` — but the jobs can't currently reach them, and there are two small bugs in the scripts. Nothing here is a criticism of the approach; building inside each distro's own container is clearly the right call.

### 1. All three jobs fail before packaging

These are pre-existing `BuildLinux.sh` problems, not something this PR introduced:

- **Arch** — `BuildLinux.sh -u` exits 1 with *"Your distribution does not appear to be currently supported by these build scripts"*. There's no `linux.d/arch`, and Arch reports `ID=arch` with no `ID_LIKE`, so the candidate loop matches nothing.
- **Fedora** — `-u` exits 0, but `linux.d/fedora` asks for `mesa-libOSMesa-devel`, which F42 obsoletes (`obsoletes: mesa-libOSMesa-devel < 25.1.0~rc2-1`). `dnf` prints `Complete!` and installs nothing, so the failure surfaces later as `/usr/sbin/ld: cannot find -lOSMesa`.
- **Debian** — `-u` exits 1 immediately. `BuildLinux.sh` runs `set -e` and `linux.d/debian` opens with `FOUND_GTK3=$(dpkg -l libgtk* | grep gtk-3)`; on a container without GTK3 the grep returns 1 and the script exits before installing anything.

I opened #11732 for those three, plus making the `libOSMesa` link conditional (Arch has no provider that doesn't conflict with `mesa`). With that branch, `-u`, `-d` and `-s` all complete on Arch and Debian.

### 2. `SLIC3R_VERSION_BASE` doesn't exist

`build_deb.sh`, `build_rpm.sh` and `build_pkg_arch.sh` all do:

```sh
BASE=$(grep 'set(SLIC3R_VERSION_BASE' "$ROOT/version.inc" | cut -d '"' -f2)
```

`version.inc` defines `set(SLIC3R_VERSION "02.08.01.55")` — there's no `SLIC3R_VERSION_BASE`, and `git log -S` shows there never has been one in that file. So all three scripts abort with *"Error: could not read version from version.inc"*. Changing the pattern to `set(SLIC3R_VERSION ` fixes it, and the version then comes out as `02.08.01.55.11879`.

### 3. The Arch package needs `glu`

With the above sorted, the full chain works and produces `bambustudio-02.08.01.55.11879-1-x86_64.pkg.tar.zst` (249 MB). Installing it on a clean Arch container resolves the declared dependencies fine, but the binary won't start:

```
/usr/lib/bambustudio/bambu-studio: error while loading shared libraries:
libGLU.so.1: cannot open shared object file: No such file or directory
```

`libGLU.so.1` comes from `glu` and isn't among the bundled `.so` files, so it needs to be in `depends=()`. Adding `glu` makes the installed binary run:

```
$ bambustudio --help
BambuStudio-02.08.01.55:
Usage: bambu-studio [ OPTIONS ] [ file.3mf/file.stl ... ]
```

Worth noting the ffmpeg sonames (`libavcodec.so.61`, `libavutil.so.59`, `libswscale.so.8`) look missing under a bare `ldd`, but they *are* bundled into `/usr/lib/bambustudio/` and resolve correctly through the wrapper's `LD_LIBRARY_PATH` — so `glu` is the only genuine gap I found.

Happy to send the version-grep and `glu` changes as a PR against your branch if that's easier than folding them in yourself.

BenJule:
Really appreciate you actually running these in clean containers, that's exactly the kind of testing CI can't give me on a fork PR.

Fixed #2 and #3 in the branch:

- Version grep now matches both `SLIC3R_VERSION` and `SLIC3R_VERSION_BASE` (`grep -E 'set\(SLIC3R_VERSION(_BASE)? '`) across all three scripts - my fork uses the latter so I need both to keep working here.
- Added `glu` to the Arch `depends=()`.

Leaving #1 to you / #11732 since that's pre-existing BuildLinux.sh territory, not something this PR should be bundling in. Feel free to send the PR against my branch if you'd rather do it that way for anything else you find - happy to take it either way.

PR #11324, opened 2026-06-27
feat: configurable default action on the print button (#11187)

### LINKED ISSUE #11187: Add setting to choose default action (Print / Print All / Send / Send All)
I frequently use the "Send All" option when preparing prints. At the moment I have to select it manually each time the action dialog appears, which becomes repetitive and slows down my workflow.

Add a preference setting that allows the user to choose which action is selected by default in the print/send dialog. The options should include: "Print", "Print All", "Send", and "Send All". The chosen option should be pre-selected whenever the dialog opens. The preference could be global, per-device, or an option to "remember last used action" or a checkbox in the dialog like "Always use this action".

- Remember the last-used action automatically and pre-select it.
- Add a keyboard shortcut or a one-click button for "Send All".
- Add a small checkbox in the dialog to set the current choice as the default.

**Additional context**
On my system I usually use "Send All", so having that as the default would save clicks and make the printing workflow faster and less error-prone. Thank you for considering this enhancement.

BenJule:
## Summary

Closes #11187. Adds a **Default print action** preference so users who always use the same action (e.g. "Send all") don't have to pick it from the print button's dropdown every time.

A new combobox on the General preferences page lets you choose the action that's pre-selected on the print button:

- **Print plate** (default, unchanged behaviour)
- **Print all**
- **Send**
- **Send all**

On startup the print button initialises its label and selection (`m_print_select`) from this setting. The dropdown still works exactly as before, so users can switch per session; this only changes which option is selected by default.

## Implementation

- `AppConfig`: new `default_print_action` key, default `"print_plate"` (so existing users see no change).
- `MainFrame`: reads the setting when building the print button and sets both the label and `m_print_select` accordingly (maps to `ePrintAll` / `eSendToPrinter` / `eSendToPrinterAll`).
- `Preferences`: the combobox.

## Test plan

- Pure preference + initial-selection change; no change to the print/send execution paths.
- Default value reproduces the current "Print plate" default exactly.
- I don't have a GUI build environment here, so I traced the wiring through the code rather than capturing a run, happy to adjust labels/placement to your conventions.

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -207,6 +207,8 @@ void AppConfig::set_defaults()
         set_bool("import_single_svg_and_split", true);
     if (get("user_bed_type").empty())
         set_bool("user_bed_type", true);
+    if (get("default_print_action").empty())
+        set("default_print_action", "print_plate");
     if (get("grabber_size_factor").empty())
         set("grabber_size_factor", "1.0");
     if (get("assembly_part_number_label_font_size").empty())
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -1959,12 +1959,26 @@ wxBoxSizer* MainFrame::create_side_tools()
     m_slice_select = eSlicePlate;
     m_print_select = ePrintPlate;
 
+    // #11187: the default action on the print button is configurable in Preferences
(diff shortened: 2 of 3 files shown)


tonghao-bbl:
Hi @BenJule we are currently making a refactor on the preference page, we'll check this after its done

tonghao-bbl:
@Mujin-code what do you think about this feature?

PR #11325, opened 2026-06-27
feat: zoom the live camera view with the mouse wheel (#11322)

### LINKED ISSUE #11322: Zoom Live Video
We can zoom live video in Bambu Handy but Bambu Studio doesn't support this. The **video should zoomable with mouse whell**.

BenJule:
## Summary

Closes #11322. The live camera view in the device tab couldn't be zoomed, unlike the pinch-zoom in Bambu Handy. This adds **mouse-wheel digital zoom** to the live view (1x fit up to 5x, centred on the view).

`wxMediaCtrl3` already paints each decoded frame itself (`wxPaintDC` + `DrawBitmap`), so this is a client-side change: a zoom factor is applied on top of the existing contain-fit scale, and the image is re-centred (cropped to the window when zoomed in).

- Scroll up to zoom in, down to zoom out; clamped to 1x-5x.
- At 1x the rendering is mathematically identical to the previous letterboxed, centred output, so the default view is unchanged.

## Implementation

- `wxMediaCtrl3.h`: a `double m_zoom` member and a `mouseWheelEvent` handler.
- `wxMediaCtrl3.cpp`: bind `wxEVT_MOUSEWHEEL`, adjust/clamp the zoom and `Refresh()`; the paint path multiplies the fit scale by the zoom and re-centres.

## Test plan

- Single self-contained control, no change to decoding or stream handling.
- Verified by tracing the math that `m_zoom == 1` reproduces the existing centring exactly (I don't have a GUI build here to capture a recording). A maintainer build is the way to confirm the feel; the step/limits (1.1x per notch, 1x-5x) are easy to tune if you'd prefer different values, and I can add cursor-anchored zoom instead of centre-anchored if that's preferred.

### DIFF
--- a/src/slic3r/GUI/wxMediaCtrl3.cpp
+++ b/src/slic3r/GUI/wxMediaCtrl3.cpp
@@ -34,6 +34,25 @@ wxMediaCtrl3::wxMediaCtrl3(wxWindow *parent)
     SetBackgroundColour("#000001ff");
     m_render_timer.SetOwner(this);
     Bind(wxEVT_TIMER, &wxMediaCtrl3::OnRenderTimer, this);
+    Bind(wxEVT_MOUSEWHEEL, &wxMediaCtrl3::mouseWheelEvent, this);
+}
+
+void wxMediaCtrl3::mouseWheelEvent(wxMouseEvent &evt)
+{
+    const int rotation = evt.GetWheelRotation();
+    if (rotation == 0) {
+        evt.Skip();
+        return;
+    }
+    // Digital zoom of the live view, 1x (fit) up to 5x, centred on the view.
+    const double factor = rotation > 0 ? 1.1 : 1.0 / 1.1;
+    double zoom = m_zoom * factor;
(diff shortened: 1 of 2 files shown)


Bambu-Pomfret:
Hi @BenJule，

Thanks a lot for your contribution! This feature works really well and brings great value.

After merging this PR, we plan to extend this function further, such as supporting left-right drag operations.
Could you please resolve the current merge conflicts first? We can proceed with the review once conflicts are cleared.

BenJule:
@Bambu-Pomfret thanks for the feedback. I have rebased the PR onto the latest master and resolved the merge conflict while preserving the updated idle-preview rendering path. The mouse-wheel zoom is now applied only to the live video rendering path. The PR is ready for review again.

lanewei120:
thanks for doing this @BenJule

osti7:
Thanks for you efforts; @BenJule, @Bambu-Pomfret, @lanewei120, @ggriffiniii, @reedy!

Issue #10328, opened 2026-04-21
X2D: Add digital AMS / external spool override when Filament Track Switch is installed


barbaraschiavinato:
**Is your feature request related to a problem? Please describe.**

Yes.

On X2D, when the Filament Track Switch is installed, the normal workflow for using external filament becomes unclear or inaccessible compared to the standard external spool workflow.

In practice, it becomes difficult (or impossible) to manually load external filament and/or clearly disable AMS from the usual interface flow. This creates unnecessary friction and can force users to physically remove the Filament Track Switch accessory just to regain a simple external spool workflow.

This defeats the purpose of the accessory: installing a hardware accessory intended to improve filament routing should not make a basic source selection workflow harder.

**Describe the solution you'd like**

Please add a clear digital filament source override for X2D when the Filament Track Switch is installed.

Ideally, users should be able to:

- Disable AMS per print
- Manually load external filament while the Filament Track Switch remains installed
- Explicitly choose filament source in software (AMS / External Spool / Auto)

This could be exposed in Bambu Studio, on the printer UI, or both.

The key goal is to avoid requiring physical removal of the accessory for what is fundamentally a software routing decision.

**Describe alternatives you've considered**

The current workaround appears to be physically removing the Filament Track Switch in order to restore the simpler external spool workflow.

Another possible workaround is relying on the current AMS enable/disable logic during print setup, but this is not always clear or sufficient from a UX perspective, especially when users need a quick manual external filament load workflow.

A less ideal alternative would be a hidden advanced setting, but a clear visible override would be much better.

**Additional context**

Please add a clear digital filament source override for X2D when the Filament Track Switch is installed.

Ideally, users should be able to:

- Disable AMS per print
- Manually load external filament while the Filament Track Switch remains installed
- Explicitly choose filament source in software (AMS / External Spool / Auto)

This could be exposed in Bambu Studio, on the printer UI, or both.

The key goal is to avoid requiring physical removal of the accessory for what is fundamentally a software routing decision.

PineCone227:
+1 Agreed. This needs to be implemented as soon as possible - it also creates a hidden hardware conflict on the X2D, where the Filament Track Switch and TPU Feed Assist Modules are mutually exclusive, as the latter depends on using external spools which the former forbids.

barbaraschiavinato:
[...]

PR #11336, opened 2026-06-28
Allow external spool workflow with Filament Track Switch

PollyBot13:
## Summary

This draft addresses `bambulab/BambuStudio#10328` by allowing the existing per-nozzle external spool virtual slots to be selected in Studio when a Filament Track Switch is installed.

Current behavior blocks external spool selection in the FTS / dynamic mapping dialog and tells users to uninstall the Filament Track Switch. That forces users to physically unplug/remove the accessory for cases where the filament can be routed externally:

- normal external spool through the external buffer
- TPU Feed Assist / flexible filament routed directly to a toolhead
- PTFE/Y-split bypass setups where FTS remains installed for the other nozzle

## Implemented Scope

This is intentionally a small Studio-side change:

- keep existing virtual external slots:
  - `255` / main-right external spool
  - `254` / deputy-left external spool
- allow those virtual slots to remain selectable in `LEFT_AND_RIGHT_DYNAMIC` mapping mode
- when sliced nozzle data is available, keep only the external slot matching the filament's sliced nozzle selectable in dynamic mapping mode, so a left-nozzle filament cannot be mapped to the right external slot and vice versa
- remove the click-handler guard that silently ignored virtual external slots in that mode
- allow the device-page external-slot `Load` / `Unload` actions to reach the existing assisted feed-direction flow when the Filament Track Switch is installed
- do not require Filament Track Switch setup/readiness for `Ext-L` / `Ext-R`, because those virtual external slots represent user-assisted external routing that may intentionally bypass the switch
- persist device-page filament choices for `Ext-L` / `Ext-R` by sending material settings for virtual slots and preserving the correct virtual tray id (`254` or `255`)
- replace the old "uninstall Filament Track Switch" warning with guidance that the external filament must be routed through the external buffer or directly/bypassing FTS to the selected toolhead

Studio sends the same existing external virtual-slot mapping for both physical routing cases; the operator/printer must enforce the actual route.

## Non-goals

This does not implement a full routing graph or claim that the Filament Track Switch can retract arbitrary external spools.

It also does not add a persisted `via buffer` versus `direct bypass` mode, because I have not found an existing printer/job protocol field for that distinction in Studio. The patch only re-enables the existing external virtual-slot mapping that Studio already serializes.

## Desired Product Behavior

The desired product behavior is:

- FTS may remain electrically connected and installed.
- A selected nozzle may use an external spool.
- The external spool may be routed through the existing buffer path for normal filaments, or directly to the selected toolhead for TPU/flexible-feed-assist setups.
- FTS/AMS can still be used for the other nozzle when the mapping is otherwise valid.

If printer firmware currently rejects external virtual-slot mapping while FTS is installed, this Studio patch will need a matching firmware capability/update. In that case, the same UI change should be gated behind a firmware/capability flag instead of being rejected as a hardware impossibility.

## Validation

- `git diff --check`
- local macOS arm64 build:
  - dependencies: `CMAKE_BUILD_PARALLEL_LEVEL=8 ./BuildMac.sh -d -a arm64 -x -c Release`
  - app: `ninja -C build/arm64 -j8`
  - packaged app: `build/arm64/BambuStudio/BambuStudio.app`
- local X2D LAN UI smoke test with Filament Track Switch present:
  - bound `20P9BJ620301973 (LAN)` from an isolated test profile
  - sliced bundled `Bambu_Cube.stl` with the X2D preset
  - confirmed the filament-switcher path is active in Studio
  - opened the send/mapping dialog without clicking `Send`
  - confirmed the selector shows Ext-L/Ext-R plus the new route/firmware note
  - confirmed Ext-L is selectable for the sliced nozzle side while Ext-R is shown but disabled
- source inspection of the device-page load menu confirmed the old FTS guard also blocked virtual external-slot `Load` / `Unload`; this patch removes that blanket block so both assisted operations can proceed through the existing feed-direction flow without requiring FTS readiness for `Ext-L` / `Ext-R`
- source inspection of the device-page filament setting path confirmed virtual external slots were not persisted reliably: ordinary filament presets were not sent unless the slot was already classified as third-party, both virtual slots were serialized as tray `254`, and the response parser only refreshed the main/right virtual slot. This patch sends virtual-slot material settings explicitly and keeps `Ext-L` / `Ext-R` distinct.
- Code inspection of existing external virtual-slot mapping path:
  - `src/slic3r/GUI/DeviceCore/DevDefs.h`
  - `src/slic3r/GUI/DeviceCore/DevMapping.cpp`
  - `src/slic3r/GUI/SelectMachine.cpp`
  - `src/slic3r/GUI/AmsMappingPopupUpdate.cpp`
  - `src/slic3r/GUI/AmsMappingPopup.cpp`
  - `src/slic3r/GUI/StatusPanel.cpp`

Still not validated:

- firmware/printer acceptance of `254`/`255` external virtual-slot mapping while FTS is installed

Addresses #10328.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -544,9 +544,22 @@ void AMSMaterialsSetting::on_select_reset(wxCommandEvent& event) {
     }
 
     if (obj) {
-        if(m_is_third){
+        if (m_is_third || is_virtual_tray()) {
             obj->command_ams_filament_settings(ams_id, slot_id, ams_filament_id, ams_setting_id, std::string(col_buf), m_filament_type, nozzle_temp_min_int,
                                                nozzle_temp_max_int);
+        }
+
+        if (is_virtual_tray()) {
+            int vt_slot_idx = ams_id == VIRTUAL_TRAY_MAIN_ID ? MAIN_EXTRUDER_ID : DEPUTY_EXTRUDER_ID;
+            while (obj->vt_slot.size() <= vt_slot_idx) {
+                obj->vt_slot.push_back(DevAmsTray(std::to_string(obj->vt_slot.size() == MAIN_EXTRUDER_ID ? VIRTUAL_TRAY_MAIN_ID : VIRTUAL_TRAY_DEPUTY_ID)));
+            }
+
+            auto &vt_tray = obj->vt_slot[vt_slot_idx];
+            vt_tray.reset();
+            vt_tray.id = std::to_string(ams_id);
+            vt_tray.color = std::string(col_buf);
+            vt_tray.set_hold_count();
         }
 
         // set k / n value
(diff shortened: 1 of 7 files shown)


BenJule (commented):
Read through the draft. The overall shape is reasonable: gating the external virtual slots (254/255) on nozzle compatibility in LEFT_AND_RIGHT_DYNAMIC mode, and surfacing why through the two tooltips, is a sensible way to let the FTS stay installed while still stopping a left-nozzle filament from mapping to the right external slot.

The main thing I would double check is the fallback in `is_ext_slot_compatible_with_current_filament`. When the nozzle group result is missing or the info list is empty it returns true (slot stays selectable). That is fail-open: if the nozzle data is unavailable for any reason, the compatibility guard effectively switches off and the wrong-nozzle mapping this PR is meant to block becomes possible again. It may be intentional (do not block when you cannot tell), but for a mapping-correctness feature it is worth confirming which failure mode you want, and whether those early `return true` paths can be hit in normal use rather than only in genuinely unknown states.

Since it is a draft, the structure and the i18n (new `_L` strings on the source side) look fine. I cannot verify the actual hardware behavior here: the FTS routing, the nozzle-to-external-slot mapping, or the assisted feed-direction flow. So this is a read of the Studio-side logic only, not a functional test on a printer with an FTS installed. Someone with the accessory should confirm the load/unload path end to end.

XinZhangBambu:
Thanks for the detailed and well-scoped contribution

One key point: **this can't be enabled from Studio alone**. Studio only serializes the external virtual-slot mapping (254/255) and sends it to the printer — whether an external spool can be used while the Filament Track Switch stays installed is enforced by the printer/AMS firmware.

BenJule:
Thanks for confirming.

That matches the limitation noted in the PR description. Studio can expose and serialize the external virtual-slot mapping, but the printer firmware ultimately decides whether the mapping is accepted while the Filament Track Switch remains installed.

Is there already a firmware capability or version check that Studio can use to detect support for this workflow?

If firmware support is not available yet, should this PR remain open until the corresponding firmware change exists, or would you prefer the Studio-side work to be postponed?

PollyBot13:
Closing this as blocked by printer/AMS firmware support: Studio can serialize the external-slot mapping, but cannot enable the workflow by itself. Happy to revisit once firmware exposes the required capability or a version/capability check Studio can rely on.

Issue #2075, opened 2023-07-18
Switch to next AMS spool when filament is empty


MrD-RC:
It would be a nice feature if you could switch to the next (or specified) slot, when the filament in the current spool runs out. This should be optional, and disabled by default.

For example, I'm printing a large project, which will use up the remaining filament on the spool in slot 1. It would be nice to be able to specify a slot to automatically use once the current spool empties. A simpler option would be to just use the next slot. But selecting a slot would be much more customisable. Maybe you could even present the option if Bambu Studio knows the print uses 700g of filament, for example. But the selected slot has less filament available on it.

Hologos:
This feature is already in there, it's called auto switch filament.

MadMax389:
> This feature is already in there, it's called auto switch filament.

Yes, but only if you remember to set up the AMS beforehand. It would be nice, as OP requested, to be prompted to switch to any slot with the same material when it runs out.

Hologos:
@MadMax389 Sorry, somehow missed that. +1 it would be cool

ChrisVonFrosta:
1 year later. Somthing happend? I still run in this error each role. But from remote i can't resume the print, even not with other color

morsik:
It's 2026 and issue remains.

I've had tangled spool which I untangled and put (without roll with RFID part) into Slot 1. Because it was without RFID, I had to set filament manually.

On Slot 2 I put exact same spool this time with RFID. It was detected.

But auto-switch filament didn't worked because apparently manually setting "Bambu PLA Black" is not the same as "Bambu PLA Black" from RFID (WHY?!) and printer refuses to continue even it had exact same filament put into AMS and even set in settings... :/

There should be simple button: **"Force another filament, I'm aware of possible problems"**.

ryclorak:
I ended up just cutting the RFID tags off, shoved one of the old tags into the new spool, and it's finally printing. 

At the very least, the RFID tag and manual settings should be recognized as the same, especially if/when the RFID reader malfunctions on one of the slots. However, it would still be better to have a force override option.

mmmankey:
Bought an AMS because I assumed this was already feature. I've just recently learned that it isn't. Auto-refill is nice but it has to be the exact filament which is inconvenient when I have my AMS filled with different filaments.

Issue #992, opened 2022-12-31
Multiple Filament Slots With Same Color (with AMS)


MadMax389:
**Describe the bug**
If multiple filament slots in the slicer are filled with the same color/material, Bambu Studio maps the first color regardless if the color in a higher slot is selected on the model. For example, if white PLA is selected for both slots 1 and 2, and I want the printer to use the PLA in slot 2 first (remnant roll), then switch to slot 1 (full roll) using the AMS auto switch function, BS will map the print to slot 1 instead of slot 2 when sending to the printer. It is possible to remap the slot on the send screen and the AMS switch function works properly, but it is easy to miss if not paying attention.

**3mf File for This Bug**
Not required

**To Reproduce**
1. Add a cube primitive to the plate
2. Set the color on filaments 1 and 2 to white generic PLA
3. Select the cube to be printed with filament 2 and slice. (shows proper filament in the slice overview)
4. Send print
5. Print is mapped to slot 1 

**Expected behavior**
Print should be mapped to slot 2

**Screenshots**
![image](https://user-images.githubusercontent.com/96731892/210148667-2b695f45-30f8-45e2-b846-4bf7fc6ea188.png)

![image](https://user-images.githubusercontent.com/96731892/210148688-ebb01dbd-3672-436f-8397-1f6d33da130e.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - BS v1.4.0.18

Benderrodrigu3z:
you can click on that image that you circled and tell it A2, as for defaulting to another line that might be a bug

MadMax389:
@Benderrodrigu3z Yes, remapping on the send print screen does work, and it switches properly. I missed the mapping the first time I did it and wondered why it printed from the full spool and not the remnant spool. Now that I'm aware, I can make sure it's mapped properly but it seems like given the choice between two slots with the same color it should default to the filament selected in the slicer for the object.

toddanglin:
I'm not sure I'd consider this a "bug," but perhaps a feature request?

You can arbitrarily configure filaments in Bambu Studio without any consideration for what is physically loaded in your AMS units. When it's time to send your print, Bambu Studio will make a "best effort" guess to match your Studio filaments to physically loaded filament in the AMS -- first matching material type (PLA, PETG, etc) and then color.

[...]

PR #11347, opened 2026-06-29
Prefer the matching AMS slot with the least remaining filament

ojimpo:
### Problem
When two or more AMS slots are loaded with the **same filament** (same type + same color), Bambu Studio's auto-mapping always assigns the print to the **lowest-numbered slot**, regardless of how much filament each spool has left.

This is a long-standing annoyance for users who keep a partially-used spool and a fresh spool of the same filament in the AMS: the slicer maps to the full spool first, leaving the partial spool to linger forever. The only workarounds today are to manually remap on the send dialog every time (easy to miss, see #992) or to physically arrange spools so the emptier one sits in the lowest slot.

Refs #992, #2075.

### Root cause
In `DevMappingUtil::ams_filament_mapping()` (`src/slic3r/GUI/DeviceCore/DevMapping.cpp`), when several trays match a sliced filament with an identical color/type distance, the tie is broken only by exact `filament_id` match. If that is still equal, the winner is simply whichever tray comes first in iteration order — i.e. the lowest slot index. Remaining amount is never considered.

### Fix
Add a secondary tie-break, implemented in a small shared helper `_prefer_tray_by_remain()`: when the color/type distance is equal **and** the `filament_id` matches, prefer the tray with the **smaller `remain`** value, so partially-used spools are consumed first.

The printer already reports per-tray remaining percentage (`DevAmsTray::remain`, populated from MQTT in `DeviceManager.cpp`). This value is propagated into `FilamentInfo` via `_parse_tray_info()` and used in the tie-break.

**Safety / backward compatibility:**
- The new rule only fires on an *exact distance tie* among same-`filament_id` trays. The scoring function and all existing mapping decisions are untouched.
- It only applies when **both** candidate trays report a valid `remain` (`>= 0`). Spools that don't report a remaining amount (e.g. most non-genuine spools) keep the previous lowest-slot-first behavior. In practice this means the new behavior kicks in for genuine/RFID spools where the remaining amount is actually known — exactly the case where it is useful and correct.
- An **empty tray (`remain == 0`) is never deliberately preferred**; conversely, a non-empty candidate wins over an empty picked tray, so auto-mapping never knowingly starts a print on an empty spool (review feedback).

### Changes
- `src/libslic3r/ProjectTask.hpp`: add `int remain{-1}` to `FilamentInfo`.
- `src/slic3r/GUI/DeviceCore/DevMapping.cpp`: copy `tray.remain` in `_parse_tray_info()`; add `_prefer_tray_by_remain()` and call it from the tie-break in both mapping passes.

### Testing
Manual: load two slots with the same filament (one partial, one full) on an RFID-reporting AMS, slice a single-color model, and send. The mapping now targets the partial spool. With remain unavailable, mapping is unchanged.

### Notes
This changes only the *initial* slot selection at send time. The mid-print auto-refill / backup failover order is firmware-controlled and out of scope.

### DIFF
--- a/src/libslic3r/ProjectTask.hpp
+++ b/src/libslic3r/ProjectTask.hpp
@@ -48,6 +48,7 @@ struct FilamentInfo
     float       used_g{0.f};
     int         tray_id{0}; // start with 0
     float       distance{0.f};
+    int         remain{-1};    // filament remain on the mapped tray: 0~100, -1 = unknown / not reported by the printer
     int         ctype = 0;
     std::vector<std::string> colors = std::vector<std::string>();
     int         mapping_result = 0;
--- a/src/slic3r/GUI/DeviceCore/DevMapping.cpp
+++ b/src/slic3r/GUI/DeviceCore/DevMapping.cpp
@@ -80,6 +80,7 @@ namespace Slic3r
         result.setting_id = tray.filament_setting_id.empty() ? tray.setting_id : tray.filament_setting_id;
         result.ctype = tray.ctype;
         result.colors = tray.cols;
+        result.remain = tray.remain; // 0~100, -1 when the printer does not report a remain (e.g. non-genuine spools)
 
         /*for new ams mapping*/
         result.ams_id = std::to_string(ams_id);
@@ -105,6 +106,24 @@ namespace Slic3r
                 result.id = ams_id * 4 + slot_id;
             }
         }
+    }
+
+    // Tie-break helper for ams_filament_mapping(): returns true when 'candidate' should
+    // replace 'picked' on an exact distance tie because it has less filament remaining,
+    // so partially-used spools are consumed first.
+    // Rules:
+    //  - only when the candidate tray is an exact filament_id match for the sliced filament;
+    //  - only when both trays report a valid remain (>= 0, e.g. genuine/RFID spools);
+    //    an unknown remain (-1) keeps the legacy lowest-slot-first behavior;
+    //  - a tray reporting remain == 0 is never deliberately preferred: a non-empty
+    //    candidate wins over an empty picked tray, never the other way around.
+    static bool _prefer_tray_by_remain(const FilamentInfo& filament, const FilamentInfo& candidate, const FilamentInfo& picked)
+    {
+        if (filament.filament_id != candidate.filament_id)
+            return false;
+        if (candidate.remain <= 0 || picked.remain < 0)
+            return false;
+        return picked.remain == 0 || candidate.remain < picked.remain;
     }
 
     int DevMappingUtil::ams_filament_mapping(const MachineObject* obj, const std::vector<FilamentInfo>& filaments, std::vector<FilamentInfo>& result, std::vector<bool> map_opt, std::vector<int> exclude_id, bool nozzle_has_ams_then_ignore_ext)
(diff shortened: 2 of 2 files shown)


BenJule (commented):
Nice, small and correct. The thing I would normally worry about here is the unknown-remain case, and you have already handled it: guarding on `picked.remain >= 0 && j.remain >= 0` means a tray reporting -1 (non-genuine / no RFID) keeps the old lowest-slot-first behavior instead of being treated as "least remaining" and wrongly preferred. That is exactly right, since -1 would otherwise sort as smaller than any real percentage. Good comments too.

Two small things, neither blocking:

* The tie-break condition is duplicated verbatim in the two branches. A tiny helper (`bool prefer_by_remain(...)`) would keep them from drifting apart later, but it is correct as is.
* One edge worth a thought: a spool reporting `remain == 0` (essentially empty) would now be preferred and mapped first, which starts the print on an empty tray and leans immediately on AMS runout/backup to switch. It still prints, but you might want a small floor (skip 0, or only prefer above some minimum) so auto-mapping does not deliberately pick a spool that is already out. Minor and arguably out of scope.

I did not build or test this, just read the mapping logic. The tie-break and the -1 guard are sound.

ojimpo:
Thanks for the careful read @BenJule — both points addressed in 8dfc74b.

**Helper extracted.** The tie-break now lives in a single `_prefer_tray_by_remain()` next to `_parse_tray_info()`, called from both mapping passes, so the two branches cannot drift apart.

**`remain == 0` edge.** Good catch. An empty tray is now never deliberately preferred (`candidate.remain <= 0` bails out). I also made it symmetric: when the currently-picked tray reports `0` and the candidate is non-empty, the candidate wins — so auto-mapping never knowingly starts a print on an empty spool from either direction. Unknown remain (`-1`) still keeps the legacy lowest-slot-first order on both sides of the comparison, exactly as before.

The resulting preference rules (candidate vs currently picked, exact distance tie, same `filament_id`):

| candidate remain | picked remain | candidate preferred? |
|---|---|---|
| -1 (unknown) | any | no — legacy order |
| any | -1 (unknown) | no — legacy order |
| 0 (empty) | any | no — never pick an empty spool |
| > 0 | 0 (empty) | **yes** — don't start on an empty spool |
| 10 | 50 | **yes** — least remaining first |
| 50 | 10 | no |
| equal | equal | no — stable, lower slot wins |

Happy to drop the symmetric "rescue from empty picked tray" rule if you or the maintainers would rather keep this PR strictly to the original tie-break — it is a one-line change in the helper.

XinZhangBambu:
Thanks again for the clean fix and for addressing the review points, @ojimpo! 🙏

Could you **squash the two commits into a single commit** before we take this further? The original change and the review follow-up are one logical change, so a single clean commit will make it easier to review and land. Thanks!

ojimpo:
Done! Squashed into a single commit (0155e2b) and force-pushed — the branch now carries one commit on top of master with the final state of the change (tie-break helper included). No code changes beyond the two commits you already saw; the diff is identical.

Thanks for taking a look, @XinZhangBambu! 🙏

XinZhangBambu:
Thanks!  Looks good to Me now.

XinZhangBambu:
LGTM

lanewei120:
thanks for doing this@ojimpo

Issue #10376, opened 2026-04-25
program closes when applying add part - svg - use surface


repon06:
### Bambu Studio Version

2.6.0.51

### Where is the application from?

Bambu Lab Official website

### OS version

mac os 26.3.1 apple m3

### Additional system information

CPU: apple m3
memory: 16 gb
displ card: integrated apple m3

### Printer

bambulab a1

### How to reproduce

open bambu studio
add new stl object
add part svg
check - use surface

### Actual results

program crash/closed

### Expected results

applying SVG images pattern to the entire surface of a part

### Project file & Debug log uploads

[test_svg.3mf.zip](https://github.com/user-attachments/files/27084094/test_svg.3mf.zip)

click on pattern - use surface

### Checklist of files to include

- [ ] Log file
- [x] Project file

repon06:
what to do? was it possible to reproduce?

Haidiye00:
Hello, we are very sorry but due to some bugs in third-party libraries on the Mac platform, we have not been able to effectively solve them yet. We suggest using the Windows version. Thank you

Haidiye00:
https://github.com/bambulab/BambuStudio/releases
Hi, please download the latest beta version to try it out. If there are still issues, we can discuss them again. Thanks a lot.

Issue #6403, opened 2025-04-02
SVG to surface crashes Bambu Studio


despisedatlanta:
### Bambu Studio Version

2.0.1.50

### Where is the application from?

Bambu Lab Official website

### OS version

Mac OS Sequoia 5.3.2 (24D81)

### Additional system information

Apple M1 Max Macbook Pro
RAM - 32GB
Built in Liquid Retina XDR Display
16-inch (3456 × 2234)
Apple XDR Display P3-1600nits
Promotion refresh rate
207GB available on 1TB SSD

### Printer

Bambu X1 Carbon.  

### How to reproduce

Insert curved object
Select object and the add modifier/SVG
Once file is selected it'll pop up on the object with the box to set adjustments to size/rotation etc.
If I click the checkbox "Use Surface" to have the SVG wrap the image onto the curved surface the program will crash.
Sometimes it's instantaneous, others it'll give me a couple seconds....long enough to think "oh cool, it'll work this time" and then poof.  Crash.

Video of what's happening attached.

https://github.com/user-attachments/assets/b563901a-f6de-4475-b161-071fe6c598bb

And note: I can do everything above in this file with the SVG as long as I don't click "Use Surface" and try to wrap it on the curve.  I can insert it flat and the program does not crash.

### Actual results

The program crashes.  I attached a video of the entire process above. The PDF attached is the Crash Report it sends to Apple.

[BambuCrashReport.pdf](https://github.com/user-attachments/files/19577370/BambuCrashReport.pdf)

### Expected results

What SHOULD happen?  The program should not crash.  LOL

### Project file & Debug log uploads

[CrashFiles.zip](https://github.com/user-attachments/files/19577269/CrashFiles.zip)

The project file, the last log file and the .svg file are included in the .zip file.

### Checklist of files to include

- [x] Log file
- [x] Project file

Haidiye00:
Hello, we have identified this issue and have confirmed that it is a problem with the third-party library on the Mac. We suggest that you temporarily try this feature on the Windows version. We will continue to analyze this issue~We are very sorry

despisedatlanta:
That’d be a great idea if I had a PC but as I don’t I’m going to have to
leave this one to you guys. 😂

Thanks!

Shayne Huff

678-283-8376 cell
Sent from my iPhone

PRIMATE - Atlanta, GA
www.facebook.com/primategrind

Haidiye00:
https://zhuanlan.zhihu.com/p/26009691548
[...]

PR #11361, opened 2026-06-30
fix(svg): prevent use-after-free when dragging a "Surround surface" SVG over an object

amhunt:
## Problem

BambuStudio frequently crashes while **dragging an SVG decal across the surface of an object**, but only when the SVG's placement mode is set to **"Surround surface"** (`Edit SVG` → *Use surface* / `use_surface`). Dragging the same SVG in flat mode never crashes. The crash is intermittent — it happens "often," not every time — which is the signature of a use-after-free landing on memory that has sometimes already been reused.

I do not have an ASan/heap trace for this one; the diagnosis below is from reading the code path, in the same spirit as #11052. The fix is defensive and correct regardless of whether the exact mechanism is as described.

## Root cause

A timing-dependent use-after-free of the dragged `GLVolume`:

1. In "Surround surface" mode — and only there — finishing a surface drag queues an **asynchronous** `UpdateSurfaceVolumeJob` (the CGAL surface re-cut). See `GLGizmoSVG::on_mouse_for_translate` → `process_job()`:

   ```cpp
   // End with surface dragging?
   if (was_dragging && !is_dragging) {
       if (m_volume->emboss_shape->projection.use_surface)
           process_job();        // queues UpdateSurfaceVolumeJob (worker thread)
   ```

2. When that background job finalizes, `UpdateJob::update_volume` calls `set_mesh()` + `set_new_unique_id()` on the SVG volume and then `Plater::changed_object()` → `reload_scene()`.

3. `reload_scene()` matches old `GLVolume`s to the new model state by `geometry_id`. Because the SVG volume's id and mesh just changed, its old `GLVolume` no longer matches and is **deleted** (and a fresh one created):

   ```cpp
   if (mvs == nullptr || force_full_scene_refresh) {
       ...
       m_volumes.release_volume(volume);
       delete volume;            // old SVG GLVolume freed here
   }
   ```

4. `SurfaceDrag::gl_volume` is a raw `GLVolume*` captured once at `start_dragging`. If the user begins the **next** drag before the background cut finishes (exactly what happens when repeatedly moving an SVG around a surface), the job finalizes *mid-drag*, freeing that volume. The next `SurfaceDrag::dragging()` call then dereferences the dangling pointer:

   ```cpp
   const ModelVolume *volume = get_model_volume(*surface_drag.gl_volume, ...); // reads freed GLVolume::object_idx() → UAF
   ```

A single slow, deliberate drag does not crash, because the job is queued only *after* `surface_drag` is reset — the race needs the async cut still running when the next drag starts. Flat mode never queues that job, so it is unaffected.

## Fix

Two changes, both small and defensive:

- **`SurfaceDrag::dragging()`** — re-resolve the dragged volume from the live selection via `get_selected_gl_volume()` at the top of every call. That helper returns a `GLVolume` from the *current* scene (matched by `composite_id`) or `nullptr`, so the pointer can never dangle; if the volume is genuinely gone the drag aborts safely instead of crashing. This is shared code, so it also fixes the identical race in the **text-emboss** surface drag.

- **`GLGizmoSVG::register_single_mesh_pick()`** — skip volumes whose `ori_mesh` is `nullptr` before passing them to `PickRaycaster`, whose constructor dereferences the mesh unconditionally (`MeshRaycaster(*mesh)`). Secondary hardening of the surface-drag raycaster setup.

No functional change on the healthy path.

## Relationship to other SVG-surface crashes

This is a **distinct** crash from the apply-time "Use surface" crashes — #6403, #10376, #7782, #5995 — which occur in the CGAL `cut_surface` path when the surface projection is *computed* (the deep-recursion stack overflow addressed by #10847). This PR does not fix those; it specifically addresses the use-after-free that happens while **moving** an already-placed surface-mode SVG. The two are complementary.

## Test plan

- [x] Static analysis of the drag/reload lifetime; change is type-checked against existing `get_selected_gl_volume(canvas)` call sites in `SurfaceDrag.cpp`.
- [ ] Local compile not yet run (full build is heavy); compile verification by a maintainer/CI appreciated.
- [ ] Functional repro: rapidly drag a "Surround surface" SVG decal back and forth across a curved object — previously crashed intermittently, should no longer.

### DIFF
--- a/src/slic3r/GUI/Gizmos/GLGizmoSVG.cpp
+++ b/src/slic3r/GUI/Gizmos/GLGizmoSVG.cpp
@@ -2168,6 +2168,10 @@ void GLGizmoSVG::register_single_mesh_pick()
         }
         auto world_tran = v->get_instance_transformation() * v->get_volume_transformation();
         auto mesh       = const_cast<TriangleMesh *>(v->ori_mesh);
+        // PickRaycaster dereferences mesh unconditionally (MeshRaycaster(*mesh)); a
+        // volume without an original mesh would crash the raycaster setup.
+        if (mesh == nullptr)
+            continue;
         if (m_mesh_raycaster_map.find(v) != m_mesh_raycaster_map.end()) {
             m_mesh_raycaster_map[v]->world_tran.set_from_transform(world_tran.get_matrix());
         } else {
--- a/src/slic3r/GUI/SurfaceDrag.cpp
+++ b/src/slic3r/GUI/SurfaceDrag.cpp
@@ -654,6 +654,19 @@ bool dragging(const Vec2d                 &mouse_pos,
               const RaycastManager        &raycast_manager,
               const std::optional<double> &up_limit)
 {
+    // Re-resolve the dragged volume from the current selection before using it.
+    // An asynchronous job - notably the "use surface" mesh recompute that is queued
+    // at the end of every surface drag - can reload the scene and delete/recreate
+    // this GLVolume while a subsequent drag is in progress. The pointer cached in
+    // surface_drag would then dangle and dereferencing it (below) is a use-after-free.
+    // get_selected_gl_volume() always returns a live volume from the current scene
+    // (matched by composite_id) or nullptr, so this keeps the drag pointing at valid
+    // memory instead of crashing.
+    if (GLVolume *gl_volume = get_selected_gl_volume(canvas))
+        surface_drag.gl_volume = gl_volume;
+    else
+        return false;
+
     Vec2d offseted_mouse = mouse_pos + surface_drag.mouse_offset_without_sla_shift;
     std::optional<RaycastManager::Hit> hit = ray_from_camera(
         raycast_manager, offseted_mouse, camera, &surface_drag.condition);


BenJule (commented):
Had a look at this since it is a concrete UAF and those are worth checking carefully. The fix looks correct, and importantly it is complete rather than a partial patch. From tracing the code:

The cached raw pointer `surface_drag.gl_volume` is only ever dereferenced inside `dragging()` (lines 668 and 679-680), and both are below the new re-resolution block, so after this change every use of the cached pointer runs against a freshly resolved, live volume. I checked the rest of SurfaceDrag.cpp too: every other gl_volume access (start_dragging, the world matrix helpers, etc.) already pulls the volume from `get_selected_gl_volume()` / `selection.get_first_volume()` on each call rather than from the cached field, so no other site is left dereferencing the stale pointer. So this really does close the hole, not just the line that showed up in the crash.

The diagnosis is plausible: the "use surface" path is the only one that queues the async UpdateSurfaceVolumeJob, that job's finalize goes through reload_scene() which frees and recreates the GLVolume by geometry id, and SurfaceDrag::gl_volume is a raw pointer captured at start_dragging, so starting the next drag before the cut finishes leaves it dangling. Re-resolving from the current selection (matched by composite_id) is the right way back to the live volume, and returning false when there is no selected volume fails safe.

A couple of notes, none blocking:

* The null branch returns false to abort the frame. I did not fully trace what every caller does with false, worth a quick confirm that it leaves drag state consistent, but it looks fine.
* The ori_mesh null check in `register_single_mesh_pick` is a reasonable separate guard (MeshRaycaster(*mesh) would deref null), but it is a different crash vector from the UAF. Not a problem, just noting it is effectively two fixes in one PR, could be split for a cleaner history.
* Longer term, the root cause is that SurfaceDrag::gl_volume is a cached raw pointer at all, when the rest of the file already re-resolves per call. Dropping the cached field, or documenting it as re-validate-before-use, would stop this class of bug returning. Not needed for this PR.

Same honesty caveat as the author: no ASan trace here either, and I did not build or reproduce. But the code path and the completeness check hold up, and the fix is low risk and defensively correct.

amhunt:
@Haidiye00 @guanyun-gudujian any idea on timeline for reviewing this? 

let me know if you have any questions or requests 🙏

Haidiye00:
Hi, is it possible to record a GIF on a Mac, thanks?

amhunt:
Hi @Haidiye00, thanks for taking a look! I can certainly record GIFs on a Mac — my hesitation is that for this particular fix, an ordinary screen recording wouldn't really prove anything, and I'd rather not offer misleading evidence.

The crash is a race condition (use-after-free). `SurfaceDrag::gl_volume` only dangles when the asynchronous "use surface" re-cut job happens to finish while the *next* drag is already in progress — and even then, the stale dereference only visibly crashes if the freed memory has already been reused. That's why the crash was intermittent to begin with: before the patch, most drags already completed fine. A GIF of me dragging a surface-mode SVG without a crash would therefore look identical with or without the patch — it can't distinguish the fix from luck.

If a recording would still be helpful, there are two I could make that would actually be meaningful:

1. **No-regression demo** — the patched build dragging a "Surround surface" SVG normally, showing the drag and the surface re-cut still behave correctly.
2. **Deterministic A/B reproduction** — temporarily add a ~2 s sleep at the top of `UpdateSurfaceVolumeJob::process()` (this only widens the existing race window; it doesn't change the mechanism) and run with `MallocScribble=1` so the stale read faults reliably. On master, this crashes on the second quick drag; with this patch, the identical gesture survives. I would record both sides and clearly note the temporary instrumentation.

That said, the fix may honestly be easiest to evaluate from the code: `reload_scene()` deletes any `GLVolume` whose `geometry_id` no longer matches (which the surface job's `set_new_unique_id()` guarantees), while `SurfaceDrag` holds a raw pointer captured once at drag start. The patch simply re-resolves that pointer from the live selection on each `dragging()` call, so it can never point at a freed volume.

Please let me know which of these would be most useful — happy to provide whichever you prefer. Thank you!

Haidiye00 (commented):
Hi, is it possible to record a GIF on a Mac, thanks?

amhunt:
> Hi, is it possible to record a GIF on a Mac, thanks?

Is this a bot reply?  I see you've now commented twice with the exact same comment...

@Haidiye00 @BenJule @guanyun-gudujian

Haidiye00:
Hello, this is not a robot response. We suggest submitting the code. In addition to the necessary explanations, providing a GIF recording video for testing would be more persuasive. Thank you

amhunt:
> We suggest submitting the code.

@Haidiye00 Thanks for confirming - sorry for the bot question. Just to check I've understood: the code is already pushed on this branch, so if "submitting the code" means something further on my side, please let me know. 

I'm building locally now and will add a GIF of the drag after the fix.

amhunt:
Here's the drag after the fix, @Haidiye00 — 10 mm SVG on a 25.6 mm cube with Surround surface on. Eight quick successive drags across the face, ~220 ms between release and the next grab; no crash, and the surface re-cut lands correctly after each release. Ran ~50 drags total off-camera, same result.

Surround-surface SVG drag on the patched build

macOS arm64, built from `9b010a57b`. cc @guanyun-gudujian

guanyun-gudujian (approved):

amhunt:
@lanewei120 mind slotting this one in when you get a chance? It's approved and the GIF that was requested is in the comment above — think it just slipped past the last round of merges. 🙏

lanewei120:
@amhunt 
thanks for doing this

Issue #11395, opened 2026-07-03
Support AUX fan filtration override for modified H2 recirculation paths


Rickdiculousme:
**Is your feature request related to a problem? Please describe.**

Some users modify the H2C, H2D, or H2S AUX fan path so it recirculates chamber air through a HEPA/charcoal filter instead of acting only as a stock part-cooling fan.

One example of this type of physical modification is:

https://makerworld.com/en/models/1416135-h2d-h2s-aux-filter-xl#profileId-1470685

In this setup, P2 is being used as a chamber recirculation/filter fan. However, Bambu Studio still treats P2 as a normal auxiliary part-cooling fan, so normal fan logic can turn it off during first-layer suppression, layer transitions, filament/tool changes, and shutdown.

The current workaround is to create a custom printer preset and override complete machine-start, machine-end, layer-change, and filament-change G-code blocks to force P2 back on. That works, but it freezes copied stock G-code templates and prevents the custom preset from inheriting future Bambu template updates.

**Describe the solution you'd like**

Add an explicit opt-in Advanced process setting for supported H2 printers that keeps the AUX fan at a configured minimum speed when used as a modified recirculation/filter fan.

Suggested behavior:

- Disabled by default.
- Visible only for H2C, H2D, and H2S.
- Hidden or inactive for H2D Pro, X2D, and other printers.
- Includes a warning that this is only for user-modified AUX recirculation/filter paths, not stock AUX fan behavior.
- Keeps P2 at a configured minimum speed during relevant print phases.
- Preserves higher normal AUX cooling requests.
- Restores P2 after filament/toolchange fan resets.
- Keeps P2 active through shutdown/purification.
- Optionally runs a final post-print dwell before turning P2 off.

**Describe alternatives you've considered**

The main workaround is using custom printer presets with copied machine G-code blocks.

The downside is that copied machine-start, machine-end, layer-change, and filament-change G-code no longer inherit future stock-template updates. That makes the workaround fragile, especially on newer H2 profiles where stock G-code can change over time.

A filament-profile setting was also considered, but this behavior is tied to a physical printer modification rather than a filament property.

**Additional context**

This feature is not intended to change stock printer behavior. It is for users who have intentionally modified the AUX fan path and want to avoid maintaining copied machine G-code templates.

I have a draft implementation that adds an Advanced process setting under:

`Process → Others → AUX filtration override`

Screenshot of the proposed UI:

[...]

PR #11396, opened 2026-07-03
Add H2 AUX filtration override

Rickdiculousme:
Related issue: #11395

## Summary

Add an opt-in, Advanced process setting that keeps the AUX fan running as a chamber recirculation filter on user-modified Bambu Lab H2C, H2D, and H2S printers.

This feature is specifically for printers whose AUX fan path has been physically modified to recirculate chamber air through a filter. It is not intended for an unmodified stock AUX part-cooling path.

Physical modification that motivated this work:

- [H2D/H2S AUX Filter XL on MakerWorld](https://makerworld.com/en/models/1416135-h2d-h2s-aux-filter-xl#profileId-1470685)

## Problem

On a modified H2 printer, the AUX fan can be repurposed to pull chamber air through a HEPA/charcoal filter.

Bambu Studio still treats P2 as an ordinary auxiliary part-cooling fan. Normal cooling behavior can therefore turn it off during:

- Initial-layer fan suppression
- Layer transitions
- Filament or tool changes
- Print shutdown

Users can force P2 on by overriding complete machine-start, machine-end, layer-change, and filament-change G-code fields in a custom printer preset. However, those copied fields stop inheriting future stock-template updates.

## Solution

Add three process-level settings under:

`Process → Others → AUX filtration override`

Settings:

- `enable_auxiliary_fan_filtration`
  - Boolean
  - Default: disabled
- `auxiliary_fan_filtration_speed`
  - Minimum P2 speed
  - Range: 1–100%
  - Default: 70%
  - 70% emits `M106 P2 S178`
- `auxiliary_fan_filtration_post_time`
  - Guaranteed final filtration dwell
  - Range: 0–180 seconds
  - Default: 60 seconds
  - Zero skips the dwell but still turns P2 off

All settings are Advanced-only. The UI warns that the feature requires a user-modified AUX recirculation/filter path.

When disabled, normal stock AUX behavior is retained.

## Supported printers

The feature is enabled only when all of the following are true:

- `auxiliary_fan` is enabled
- `support_cooling_filter` is enabled
- `printer_model` is exactly one of:
  - `Bambu Lab H2C`
  - `Bambu Lab H2D`
  - `Bambu Lab H2S`

Explicitly excluded from V1:

- Bambu Lab H2D Pro
- Bambu Lab X2D
- Every other AUX-fan printer

The UI and G-code generation use the same shared capability gate.

### Startup

After the final stock P2-off operation and immediately before the nozzle-load/purge line:

```gcode
{if auxiliary_fan_filtration_active}
M106 P2 S{auxiliary_fan_filtration_speed_num} ; AUX filtration override
{endif}

;===== nozzle load line ===============================
```

Earlier calibration, chamber-control, homing, and leveling commands remain unchanged.

### Generated layers

The effective AUX speed is:

```text
max(normal AUX request, configured filtration speed)
```

The floor is applied after first-layer suppression and fan ramping.

This means:

- First-layer cooling suppression cannot turn filtration off.
- Higher normal AUX cooling requests are preserved.
- P2 is reasserted at generated layer-change markers.
- No layer-change template override is required.

### Filament and tool changes

The stock P2 reset remains in place and filtration is immediately restored before material-switch operations:

```gcode
M106 P2 S0
{if auxiliary_fan_filtration_active}
M106 P2 S{auxiliary_fan_filtration_speed_num} ; AUX filtration override
{endif}
```

There is no wait or material-switch command between the stock P2-off and recovery.

AMS, flushing, synchronization, tool selection, temperature, and detection commands are unchanged.

### Machine end

The early stock P2 shutdown becomes conditional:

```gcode
{if auxiliary_fan_filtration_active}
M106 P2 S{auxiliary_fan_filtration_speed_num} ; AUX filtration override
{else}
M106 P2 S0 ; turn off remote part cooling fan
{endif}
```

P2 remains active through normal unloading, heater shutdown, safe movement, state reset, and stock purification.

Existing P3/P6/M145/M145.2 purification behavior is unchanged.

After stock purification and before the finish sound and motor disable:

```gcode
{if auxiliary_fan_filtration_active}
M106 P2 S{auxiliary_fan_filtration_speed_num} ; final AUX filtration dwell
{if auxiliary_fan_filtration_post_time > 0}
M400 S{auxiliary_fan_filtration_post_time}
{endif}
M106 P2 S0 ; turn off AUX filtration
{endif}
```

The post-print setting is a guaranteed final dwell, not total post-print P2 runtime. P2 may already have been running during the preceding shutdown and purification sequence.

No undocumented asynchronous timer or M106 parameter is introduced.

## Validation

Completed locally on macOS:

- Core production code compiled.
- New AUX filtration test objects compiled.
- `libslic3r_gui` compiled.
- Full BambuStudio application built and launched.
- The new UI appeared correctly.
- Speed and post-time fields were disabled when the override was off.
- The group was hidden on unsupported printers.
- Exported G-code was validated for:
  - H2C enabled and disabled
  - H2D enabled and disabled
  - H2S enabled and disabled
  - X2D enabled and disabled control samples
- H2C/H2D/H2S exports contained the expected:
  - Startup P2 reassertion
  - Layer-change reassertion
  - Toolchange recovery
  - End-sequence retention
  - `M400 S60` final dwell
  - Final P2-off
- Higher normal AUX requests were not reduced.
- No sustained in-print P2-off command defeated the override.
- X2D emitted no filtration hooks or final dwell when the setting was enabled.
- Disabled exports emitted no filtration hooks or final dwell.
- `git diff --check` passed.
- All nine modified H2 template JSON files parsed successfully.

Focused tests cover:

- Setting defaults and bounds
- Exact H2C/H2D/H2S capability gate
- Explicit H2D Pro rejection
- Filtration as a minimum speed
- Preservation of higher normal AUX speed
- Startup and layer reassertion
- Enabled machine-end boundary behavior
- Disabled stock shutdown behavior
- Post-time zero behavior
- Final dwell ordering
- Unsupported X2D generated output

### Full test executable limitation

The complete local test executables currently fail to compile before reaching the new AUX tests because of unrelated, pre-existing tests that are stale relative to current production APIs:

- `tests/libslic3r/test_3mf.cpp`
- `tests/fff_print/test_gcodewriter.cpp`

The new AUX test translation units compile successfully. The GUI library and full application also build successfully.

### Stock-template dependency

Final dwell and final P2 shutdown are emitted through the supported stock H2 machine-end templates. Custom H2 machine-end G-code that replaces the stock template would also need to include the new hook.

The intended usage is a process setting layered onto current stock H2 printer profiles.

### Transient toolchange P2-off

Filament-change templates intentionally retain the stock:

```gcode
M106 P2 S0
```

This is followed immediately by the configured P2 recovery. No wait or material-switch operation occurs between the two commands.

Keeping the stock reset avoids changing undocumented toolchange assumptions while still restoring filtration before flushing and switching operations.

### Internal template synchronization

The H2 template resources appear to be synchronized from Bambu’s internal custom-G-code source.

Equivalent changes may need to be applied to that source so a future automated template synchronization does not overwrite these hooks.

## Reviewer notes

- The template JSON files store complete escaped G-code blocks on a single JSON line. Their raw diffs therefore appear much larger than the semantic changes. Decoded snippets are included above for readability.
- No BBL profile version bump is included. Vendor profile versioning appears to be managed separately during release preparation.
- No `.pot`, `.po`, or `.mo` changes are included in this draft. Translation resources appear to be regenerated by the localization pipeline.
- V1 is intentionally limited to an explicit opt-in process setting; it does not add Auto mode, nozzle-temperature thresholds, or filament-profile behavior.
- This feature is disabled by default and does not target unmodified stock AUX fan behavior.
- AUX filtration support is now profile-driven through `support_auxiliary_fan_filtration`; future supported printers can opt in through profile data once their P2 AUX semantics and template hooks are verified.
```


Rickdiculousme:
This is now ready for maintainer review. GitHub Actions is awaiting maintainer approval before running the workflow.

Local validation completed:
- Core production code compiled.
- New AUX test translation units compiled.
- `libslic3r_gui` compiled.
- Full `BambuStudio` app built and launched locally.
- UI appeared as expected.

I have exported validation G-code samples for H2C, H2D, H2S, and X2D with the override enabled/disabled. I have not committed them because they are generated artifacts, but I can attach them if useful.

Summary of observed output:
- H2C/H2D/H2S enabled exports emit `M106 P2 S178` for 70%.
- Startup reassertion appears immediately before the nozzle-load line.
- Layer/toolchange reassertions are present.
- Final `M400 S60` dwell appears after stock purification.
- P2 turns off after the final dwell.
- X2D emits no filtration hooks even when the setting is enabled.
- Disabled exports emit no filtration hooks or final dwell.

BenJule (commented):
Took a read through this (the approach and the gcode side, not a local build). It is cleanly done, a few notes in case they help.

What I like: the capability check lives in one place (`supports_auxiliary_fan_filtration`) and both the UI and gcode generation go through it, so they cannot drift apart. Default off and gated behind `auxiliary_fan` + `support_cooling_filter` means zero impact on anyone not opting in, and the layer path using `max(normal aux, filtration speed)` avoids fighting the normal cooling logic. Good that there are tests too, including the negative case that X2D does not pass the gate. Targeting the 0.4 nozzle template is also correct here, since the 0.6/0.8/0.2 machines inherit from it, so no nozzle variants are missed.

One thing worth raising: the gate hardcodes the model names (`printer_model == "Bambu Lab H2C" || ...`). It works today but it is brittle: a new H2 variant, a rename, or any future printer with the same mod needs a C++ change, and it splits "which printers can do this" between code and profiles. A machine-profile capability flag that this function reads instead would keep it data driven and let X2D / H2D Pro opt in later without touching code. Since X2D is excluded from V1 and the same MakerWorld filter mod exists for it, that may come up quickly.

Two practical points: CI has not run on the branch yet (no checks reported), so it would help to get a maintainer to trigger the workflows and confirm build/tests are green before review. And the real call here is probably policy rather than code, since this officially supports a user modified AUX path. The PR handles that about as responsibly as it can (opt in, off by default, clearly scoped, warned in the UI), so calling that out explicitly for whoever reviews might speed the decision.

I could not verify runtime behavior (no slicing on my end), this is a read of the diff and the gcode injection points.

Rickdiculousme:
I updated the PR to address the maintainability concern about hard-coded printer model names.

The AUX filtration support gate is now profile-driven via `support_auxiliary_fan_filtration` instead of checking `printer_model` strings in C++. The common machine profile defaults this capability off, and only H2C/H2D/H2S opt in for V1.

For future printers, support can be added in profile data rather than C++ by:
1. confirming the printer has `auxiliary_fan` and `support_cooling_filter`,
2. confirming P2 is the correct AUX fan target for that printer,
3. confirming the stock start/change-filament/end templates include the required AUX filtration hooks or inherit from templates that do,
4. setting `support_auxiliary_fan_filtration` to `1` in the supported printer profile.

This keeps UI visibility and G-code generation using the same shared capability check while avoiding a growing model-name allowlist in source code.

Rickdiculousme:
@FengMi-BamBu @XunZhangBambu 
Should I update this to be in line with the current version? or wait until the reviewers have time to commit to reviewing? I'm not sure how this works and if Bambu doesn't plan on integrating it, then I don't want to keep updating it. Sorry, it's my first time doing a PR like this. Cheers

PR #11409, opened 2026-07-05
deadlock in server connected callback calls set_user_selected_machine synchronously

vjanelle:
On macOS, the 'server connected' callback fires while the closing networking plugin holds internal locks. The callback in init_networking_callbacks() calls set_user_selected_machine() directly (not via CallAfter), which tries to acquire those same locks on the UI thread, causing a deadlock that freezes the device page UI.


vjanelle:
I can't quite prove this (I think) because I can't test with the binary networking plugin because of unsigned code, but it appears to be stable for me so far.

harry-bbl:
Thanks for the PR! Just a quick heads-up: this lambda itself is used inside CallAfter, meaning it won't actually run while holding the lock. Appreciate the effort though!

PR #11412, opened 2026-07-05
feat: add Preferences toggle to hide the daily tips panel (#11402)

### LINKED ISSUE #11402: Disable Daily Tips on MacOS
I am on version 2.7.1.62 on MacOS.

PLEASE allow us to completely disable the Daily Tips in Settings. There is no check box in Settings currently.

BenJule:
### What
Closes #11402. Adds a Preferences checkbox to turn off the daily tips panel shown in the slicing progress notification. There was no way to disable it before.

### Why it was missing
The `show_daily_tips` preference already exists (it defaults to true in AppConfig) and a checkbox for it was written but left commented out in Preferences, with nothing ever reading the preference. The panel only reads `show_hints`, which controls its expand/collapse state, not whether it appears at all. So there really was no Settings control, which is what the reporter hit.

### Change
* Preferences: uncomment the checkbox, relabelled to "Show daily tips" with a clearer tooltip.
* SlicingProgressNotification: gate the panel on `show_daily_tips`. When it is off, the notification height drops the panel contribution and the separator line plus the panel render are skipped.

The preference defaults to true, so the default path is the same as before (the height formula reduces to the original expression when the flag is true). Only the new opt-out branch is added, so it cannot regress the current behaviour.

### Note
I do not have a GUI build here, so I have not visually confirmed the notification layout with the panel hidden. CI covers compilation, and since default behaviour is unchanged the risk is limited to the disabled state. Happy to adjust the label or wording.

### DIFF
--- a/src/slic3r/GUI/Preferences.cpp
+++ b/src/slic3r/GUI/Preferences.cpp
@@ -1610,11 +1610,11 @@ void PreferencesDialog::create_gui_page()
 
     auto title_index_and_tip = create_item_title(_L("Home page and daily tips"), page, _L("Home page and daily tips"));
     auto item_home_page      = create_item_checkbox(_L("Show home page on startup"), page, _L("Show home page on startup"), 50, "show_home_page");
-    //auto item_daily_tip      = create_item_checkbox(_L("Show daily tip on startup"), page, _L("Show daily tip on startup"), 50, "show_daily_tips");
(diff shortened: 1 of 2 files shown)


tonghao-bbl:
Hi @BenJule We have made a refactor on the preference page and will be available on github days later, can you refactor this next version?

BenJule:
@tonghao-bbl thanks for the update!

Yes, I can refactor this once the new Preferences page refactor is available on GitHub. Please ping me or point me to the relevant branch/commit when it lands.

I will keep the change focused on wiring the existing `show_daily_tips` preference into the new Preferences UI and keeping the slicing progress notification gated by that setting.

PR #11413, opened 2026-07-05
feat: add "Clear recent projects" menu entry (#11401)

### LINKED ISSUE #11401: Add "Clear Recent Projects" option to Bambu Studio
## Is your feature request related to a problem? Please describe.

The "Recent Projects" list in Bambu Studio keeps growing over time.

Unlike the printer print history, there is currently no simple way to clear the local Recent Projects list from within the application.

The only workaround is manually deleting configuration files, which is not practical for normal users.

## Describe the solution you'd like

Please add an option to manage the Recent Projects list.

Examples:
- File → Clear Recent Projects
- Right-click → Clear Recent Projects
- Remove individual entries
- Option to limit the maximum number of recent projects

## Describe alternatives you've considered

The only current solution is deleting configuration files inside the Bambu Studio settings folder, which should not be necessary.

## Additional context

Most Windows applications provide an easy way to clear the Recent Files list.

Bambu Studio already allows clearing the printer print history, so providing the same functionality for the local Recent Projects list would make the user experience more consistent.

BenJule:
### What
Closes #11401. Adds a "Clear recent projects" entry to the File menu so the recent projects list can be cleared from the UI. Until now the only way was to delete config files by hand.

### Change
The new item sits next to the existing "Recent projects" submenu. When clicked it removes every entry from the in-memory FileHistory, clears the persisted recent_projects config and saves. It is disabled when the list is already empty, using the same condition as the submenu.

### Note
No GUI build here, so I have not clicked it in a running build. CI covers compilation. It reuses existing, already-used APIs (FileHistory::RemoveFileFromHistory, AppConfig::set_recent_projects and save), so the risk is low. Happy to move it inside the submenu instead if you prefer that placement.

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -2909,6 +2909,15 @@ void MainFrame::init_menubar_as_editor()
 
         Bind(wxEVT_UPDATE_UI, [this](wxUpdateUIEvent& evt) { evt.Enable(can_open_project() && (m_recent_projects.GetCount() > 0)); }, recent_projects_submenu->GetId());
 
+        // #11401: let users clear the recent projects list from the UI
+        append_menu_item(fileMenu, wxID_ANY, _L("Clear recent projects"), _L("Clear the recent projects list"),
+            [this](wxCommandEvent&) {
+                while (m_recent_projects.GetCount() > 0)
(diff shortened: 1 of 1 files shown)


Haidiye00 (approved):
approve

BenJule:
Closing in favor of #11687 and #11494, which cover the same recent-projects clearing and also keep the WebView list in sync (this one doesn't).

Issue #11392, opened 2026-07-03
support_top_z_distance


krknopp:
### Type of issue

Other (functional bug, wrong result, UI glitch, etc.)

### Bambu Studio Version

2.x (tested all the way to latest beta)

### Where is the application from?

Bambu Lab GitHub releases

### OS version

MacOS 26.5.2

### Additional system information

Apple M4 Max
128GB RAM

### Printer

X1C

### How to reproduce

Any model with tree supports enabled
Layer height: 0.3mm, nozzle: 0.6mm
Set support_top_z_distance to 0.6mm (or any value other than 0.3mm)
Slice and export gcode
In the gcode header, confirm support_top_z_distance = 0.6 is recorded correctly
Search for ; FEATURE: Support interface and note the Z_HEIGHT of the last interface layer
Find the next Z_HEIGHT after the interface ends — this is the first model layer
Subtract: the gap is always 0.3mm regardless of setting

### Actual results

The gap is always exactly one layer height (0.3mm) regardless of what support_top_z_distance is set to. Setting the value to 1.2mm produces the same 0.3mm gap as setting it to 0.2mm. The setting is written correctly to the gcode header but has no effect on toolpath generation.

### Expected results

The Z gap between the last support interface layer and the first model layer should match support_top_z_distance. At 0.6mm setting and 0.3mm layer height, the gap should be two layers (0.6mm).

### File uploads (project / log / crash dump / performance profile)

[support-test.zip](https://github.com/user-attachments/files/29647361/support-test.zip)

### Checklist of files to include

- [x] Project file (always)
- [x] Log file (always)
- [ ] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)
- [ ] Performance profile — required for performance / lag issues

BenJule:
I had a dig through the code for this one. It looks like the tree support top gap is effectively hardcoded to one layer.

In src/libslic3r/Support/TreeSupport.cpp, where the contact nodes are created, there is:

```cpp
const int z_distance_top_layers = round_up_divide(scale_(z_distance_top), scale_(layer_height)) + 1;
int gap_layers = z_distance_top == 0 ? 0 : 1;
```

[...]

PR #11426, opened 2026-07-05
fix: snap support_top_z_distance to layer multiple in non-independent…

krknopp:
this is a fix for #11392 

In non-independent support layer height mode (e.g. when the prime tower is enabled), z_distance_top was passed raw to the gap-node accumulation loop, which always rounded UP to the nearest layer height. Any configured value between 0 and layer_height produced a gap of exactly one layer regardless of the setting.

Restore the nearest-multiple rounding that was commented out in 39ae64fc5 so the gap snaps to the physically achievable multiple closest to the configured value. When the value rounds to zero (e.g. 0.1 mm with 0.3 mm layers), clamp to one layer minimum so the gap node is still generated. Independent mode is unaffected and continues to produce the exact configured distance.

### DIFF
--- a/src/libslic3r/Support/TreeSupport.cpp
+++ b/src/libslic3r/Support/TreeSupport.cpp
@@ -3948,14 +3948,12 @@ void TreeSupport::generate_contact_points()
 
     const coordf_t layer_height = config.layer_height.value;
     coordf_t       z_distance_top = this->top_z_distance;
-  //  if (!m_support_params.independent_layer_height) {
-  //      z_distance_top = round(z_distance_top / layer_height) * layer_height;
-  //  // BBS: add extra distance if thick bridge is enabled
-  //  // Note: normal support uses print_z, but tree support uses integer layers, so we need to subtract layer_height
-  //  if (!m_slicing_params.soluble_interface && m_object_config->thick_bridges) {
-  //      z_distance_top += m_object->layers()[0]->regions()[0]->region().bridging_height_avg(m_object->print()->config()) - layer_height;
-		//}
-  //  }
+    if (!m_support_params.independent_layer_height) {
+        z_distance_top = std::round(z_distance_top / layer_height) * layer_height;
+        // If the configured gap rounds to zero, use one layer minimum so the gap node is still created.
+        if (z_distance_top < EPSILON && this->top_z_distance > EPSILON)
+            z_distance_top = layer_height;
+    }
     const int z_distance_top_layers = round_up_divide(scale_(z_distance_top), scale_(layer_height)) + 1; //Support must always be 1 layer below overhang.
     int gap_layers = z_distance_top == 0 ? 0 : 1;
 


BenJule (commented):
Dug into this one properly since I had looked at #11392 earlier. Short version: the diagnosis is right and the fix looks correct and well scoped.

Confirming the regression: the rounding this restores was commented out in 39ae64fc5 ("ENH: accurate top z distance for tree support"). `git log -S` on the `z_distance_top = round(...)` line points at exactly that commit, so the "accurate top z distance" change is ironically what introduced the inaccuracy in non-independent mode. Good catch pinning it down.

On the mechanism: z_distance_top is passed into create_node as the node height, and the comment right above it literally says "height=z_distance_top: it's height is exactly the gap distance", so snapping it to a layer multiple does flow through to the real gap on the non-independent path. The `gap_layers = z_distance_top == 0 ? 0 : 1` on the next line is only the virtual contact node marker ("distance_to_top=-1: it's virtual"), not the gap magnitude, so leaving it untouched is correct.

Scope looks right too. independent_layer_height comes from independent_support_layer_height, which is forced off when the prime tower is enabled, which matches the reporter's setup. It also lines up with an independent mode gcode I measured for #11392: layer 0.3 / top distance 0.6 gave a roughly two layer (0.6 mm) gap, i.e. the independent path was already honoring the setting. So restricting the change to `!independent_layer_height` and leaving independent mode alone is the right call.

The zero clamp is sensible: it checks `this->top_z_distance` (the original value) so an explicit 0 (soluble interface) still yields 0, while a sub layer value like 0.1 mm on 0.3 mm layers floors to one layer so the gap node is still created instead of fusing support to the model.

Two suggestions, neither blocking:

* A regression test would be worth it: force non-independent mode with a sub layer top distance and assert the resulting gap, so this does not get silently reverted again (which is essentially what 39ae64fc5 did). At minimum, a before/after sliced gcode with the prime tower on for say 0.6 and 1.2 mm would make the effect concrete.
* Consider a one line comment on why the rounding is required here (tree support walks integer object layers in non-independent mode), so a future cleanup does not comment it out a second time.

Caveat on my end: I verified the code path, the git history and the logic, but did not build or slice, so I have not confirmed the exact runtime gap. A before/after gcode would close that in the review.

ArthurBambulab (changes requested):
As mentioned in https://github.com/bambulab/BambuStudio/issues/11392, it's not a bug.

exc2k:
> As mentioned in #11392, it's not a bug.

I don't get why this isn't a bug... I am having the same issues:

Having independent support structure enabled, everything works fine with my settings of z-distance: 0,05mm at 0,1mm layer height.

But when printing multicolor with **prime tower** the independent layer height for supports get deactivated in the background without any notice and none of my models sticks to the interface layers anymore because of the gap produced.

so I think this is a bug as for me the z-distzance has much higher priority than an throw away prime tower.. so why is it not possible to use towers with independent support layer height?

PR #11105, opened 2026-06-11
Preserve per-triangle painting through merge, split, merge-to-multipart, and repair

Cstm3DBldr:
## What
Bambu Studio drops per-triangle painting (color/MMU, support, seam, fuzzy-skin) on several mesh-editing operations. Upstream already fixed **cut** in #10613 (@solidblu). This PR continues that work, extending the **same source-face-index / annotation-propagation idea** from plane cuts to the remaining operations:
- **merge** (to a single object), **split**, and **merge-to-multipart** — exact carry-through of the facet annotations.
- **repair** — best-effort transfer across the re-projected mesh, with a cancelable **warning dialog** before a repair that may alter painted data.

## Why
Losing paint on these routine edits forces users to re-paint after every operation. Cut already preserves it (#10613); these operations should behave consistently.

## Relationship to #10613
This is the natural next step after @solidblu's merged cut-preservation PR (#10613), which was itself based on the corresponding PrusaSlicer fix. Same approach, applied to merge/split/multipart/repair.

## Changes (6 files, +178/-4)
- `libslic3r/Model.{cpp,hpp}` — carry annotations through merge/split/multipart and repair
- `slic3r/GUI/GUI_ObjectList.cpp`, `GUI_App.{cpp,hpp}` — warning dialog and operation hooks
- `slic3r/Utils/FixModelByWin10.cpp` — preserve through the Win10 repair path

## Testing
Built locally on Windows (VS2022 / MSVC v143) from current master; validated interactively:
- Merge -> single part, Split, Merge-to-multipart — paint retained (including painted supports)
- Repair keeps paint; warning dialog appears and cancels correctly

## Scope
Cut is already upstream (#10613, untouched here). Exact preservation for merge/split/multipart; best-effort for repair. Simplify/smooth/boolean preservation are intentionally not in this PR (separate work, still WIP).

Edit::::
Related to #750 — the report of paint being lost on Cut. That case is fixed upstream in 2.7.1 (#10613); this PR extends the same preservation to the operations that still drop paint.

PR #11444, opened 2026-07-08
Preserve per-triangle painting through Boolean, merge and Simplify

Cstm3DBldr:
Follow-up to #11105, which brought painting through merge, split, merge-to-multipart and repair. This covers the remaining mesh-rebuilding operations: **Boolean** (union / intersection / difference, both the gizmo and the right-click menu), **merge into single part**, and **Simplify**.

Rebased onto current master.

### Approach

`PaintReproject.hpp` already described this as the intended extension point:

> *"Boolean is not wired up yet, but `reproject_paint_geometric` is the intended extension point: pass the boolean result as the destination mesh and each participating operand mesh as a source (nearest-distance source selection or a merged source selector can be layered on later)."*

`reproject_paint_from_volumes()` implements the nearest-distance source selection described there. Each operand is reprojected **on its own** through `reproject_paint_geometric` - a single source per call, so operands cannot overwrite one another where they overlap - and each destination face then takes the paint of the nearest operand that **covers** it. An operand only counts as covering a face when the closest point lies inside one of its own faces, which stops paint spreading past a part's outline where two parts share a plane; a tie there goes to the smaller part. A face no operand reached, such as surface newly cut by a boolean, takes the plain filament of its nearest operand. The header comment is updated accordingly.

### Solid part color

A boolean or merge result is a single volume with a single extruder, so any area left unpainted silently takes that one color. `TriangleSelector::fill_unpainted()` bakes each operand's own filament into its unpainted areas while leaving painted detail untouched.

This also fixes **color loss in merge-into-single-part**: the per-face copy left the area *around* each brush stroke unpainted, and those faces fell back to the merged volume's extruder, so a part merged with a differently-colored one lost its color.

Simplify keeps its paint via `ModelVolume::set_mesh_keep_paint()` instead of `set_mesh()`. The `clear_before_change_mesh()` call that preceded it reset the paint on every volume of the object, which defeated the reprojection.

### Second commit: self-intersecting parts

`check_boolean_possible()` runs CGAL, whose `does_self_intersect()` rejects a part outright, while the boolean that follows runs mcut, which copes with far messier meshes. Because a CGAL verdict skipped the attempt entirely, a model with even a couple of overlapping triangles never reached mcut - the operation fell through to the branch that merges the parts' meshes and reports a failure, so the user asked for a boolean and silently got the parts concatenated, overlap included.

Allowing `SelfIntersect` through lets mcut try. The call is already wrapped in try/catch and its result is only accepted when non-empty, so a mesh mcut cannot handle still lands in the same merge fallback and warning as before. This only turns former hard failures into successes. Note `combine_mesh_fff()` is shared with STL export, so a self-intersecting model now exports the boolean result rather than the merged fallback. Kept as a separate commit so it can be dropped independently.

### Testing

Verified by decoding the per-triangle annotations out of saved 3MFs rather than by eye. Two parts on different filaments, each brush-painted, then each operation run and saved:

| Operation | Faces | Unpainted | Result |
|---|---|---|---|
| Union | 48 | 0 | correct per-part color, 16 faces retain brush detail |
| Intersection | 22 | 0 | 0 mismatches against per-plane source ownership |
| Difference | 36 | 0 | cut walls take the kept part's filament |
| Merge into single part | 24 | 0 | exactly the source's 12 brush-painted faces |
| Merge -> Simplify | 24 | 0 | detail survives decimation |
| Merge -> Split to object | 24 | 0 | |
| Merge -> Split -> Merge multipart | 24 | 0 | |
| Union (part mode) + slice | 48 | 0 | colors reach the slicer |

Cut and repair were re-checked for regressions and are unchanged; the surface a cut newly creates stays unpainted, as before.

Also exercised on a real 45,680-triangle sculpted model: unpainted faces dropped from 68.2% to 0.1%, brush detail rose from 704 to 1,063 faces as the boolean subdivided along existing paint boundaries, and the operation succeeded on a mesh that self-intersects.

### Demos

Boolean Union (toolbar)
Boolean Union Toolbar

Boolean Intersection (toolbar)
Boolean Intersection Toolbar

Boolean Subtraction (toolbar)
Boolean Subtraction Toolbar

Right-click Boolean
Right Click Boolean

Simplify, high to low resolution
Simplify high-low res

Merge into single part - the operation this PR fixes for parts on different filaments (clip from #11105, showing the original merge preservation)
merge into single part paint preservation

Repair (unchanged by this PR, shown for continuity with #11105)
Repair

### DIFF
--- a/src/libslic3r/Model.cpp
+++ b/src/libslic3r/Model.cpp
@@ -2840,26 +2840,27 @@ ModelObjectPtrs ModelObject::merge_volumes(std::vector<int>& vol_indeces)
 
 #if 1
     TriangleMesh mesh;
-    // BBS: preserve painting across the merge. its_merge() appends faces in
-    // order (only vertex indices are offset), so face f of the merged mesh maps
-    // exactly to the captured per-part triangles below. Capture before
-    // reset_mesh() empties the source volumes.
-    std::vector<std::string> merged_supported, merged_seam, merged_mmu, merged_fuzzy;
+    // Preserve painting across the merge. Each part is collected in object frame with
+    // the filament it prints in, so a part that is unpainted but assigned a different
+    // filament keeps its color once the parts share a single volume. The copy itself
+    // stays an exact per-face one, the merged mesh being the parts concatenated.
+    // Source meshes are released only after the copy has read them (see below).
+    std::vector<PaintSourceVolume> paint_sources;
     for (int i : vol_indeces) {
         auto volume = volumes[i];
         if (!volume->mesh().empty()) {
-            const size_t nf = volume->mesh().its.indices.size();
-            for (size_t f = 0; f < nf; ++f) {
-                merged_supported.emplace_back(volume->supported_facets.get_triangle_as_string((int)f));
-                merged_seam.emplace_back(volume->seam_facets.get_triangle_as_string((int)f));
-                merged_mmu.emplace_back(volume->mmu_segmentation_facets.get_triangle_as_string((int)f));
-                merged_fuzzy.emplace_back(volume->fuzzy_skin_facets.get_triangle_as_string((int)f));
-            }
-            const auto volume_matrix = volume->get_matrix();
-            TriangleMesh mesh_(volume->mesh());
-            mesh_.transform(volume_matrix, true);
-            volume->reset_mesh();
+            PaintSourceVolume s;
+            s.mesh                  = &volume->mesh();
+            s.supported             = &volume->supported_facets;
+            s.seam                  = &volume->seam_facets;
+            s.mmu                   = &volume->mmu_segmentation_facets;
+            s.fuzzy                 = &volume->fuzzy_skin_facets;
+            s.source_to_destination = volume->get_matrix();
+            s.base_filament         = volume->extruder_id(); // bake solid part color into paint
+            paint_sources.push_back(s);
 
+            TriangleMesh mesh_(volume->mesh());
+            mesh_.transform(volume->get_matrix(), true);
             mesh.merge(mesh_);
         }
     }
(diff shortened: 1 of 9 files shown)


Haidiye00 (approved):
approve

Cstm3DBldr:
How often do public releases get pushed and what is the deciding factor when to push?

Cstm3DBldr:
@Haidiye00 still waiting on the merge is there something i need to do to get this pushed in?

lanewei120:
Hello @Cstm3DBldr 

could you help to rebase to the newest code base?

sorry previously we missed this PR

Cstm3DBldr:
@lanewei120 
Sure i started to once i saw the new release. and got busy should be freed up in a day or so and ill get updated code uploaded.

I have a few tools that would be nice to have in the works as well mostly related to painting models. Ill Keep you posted when those are ready to merge and fully working.

Thanks for reaching out.

Cstm3DBldr:
@lanewei120 Rebased onto the latest master — no conflicts, and it shows as mergeable now.

One thing I want to flag rather than let you find it in review: this is not the same patch
that was approved back in July. Rebasing onto 2.8.2 meant rewriting it against the new
PaintReproject module, and while I was testing the rebase I found the merge and Boolean
paths were still losing color in cases the original patch never covered. So the
implementation changed substantially and I think it deserves a fresh look rather than
carrying the old approval forward.

What it does now:

- Boolean (union / subtract / intersect) keeps per-triangle paint. This covers both the
  synchronous and the asynchronous result paths — the async one is the default and was
  producing completely unpainted results.
- Merge into single part copies the paint per face instead of reprojecting it, so the
  merged mesh is an exact copy of what the parts had.
- Simplify no longer wipes paint. It was clearing all annotations before reprojecting.
- A self-intersecting model is no longer blocked from Boolean outright; the tolerant path
  is allowed to attempt it, which real-world models need.

I checked the results by reading the per-triangle annotations back out of saved 3MFs rather
than judging by eye: union 48 faces / 0 unpainted, subtract 36 / 0, intersect 22 / 0, merge
24 / 0 with exactly the 12 painted faces the source had.

Separately, I opened #12088 for the SVG work I mentioned — reading CSS styling and clip
paths that the parser currently ignores, plus importing outline art as colored parts. It
touches different files and does not depend on this one, so the two can go in any order.

Cstm3DBldr:
@lanewei120, @Haidiye00  — no urgency, just flagging that it is
ready whenever you have time.

All fourteen build jobs are green against current master. The only loose end
is the one I mentioned above: the July approval predates the rebase, and the
implementation changed substantially in it. I would rather it get a fresh look
than went in on an approval for different code.

If it would make review easier, I am happy to split the Boolean, merge and
Simplify changes into separate PRs — they do not depend on each other.

lanewei120:
Hi @Cstm3DBldr 

I have asked @StoneLiBambu to do the final check as he also did some work with boolean internally

liyan0827:
Hi @Cstm3DBldr ,
First of all, thank you very much for your code contribution. However, I encountered a few issues while working on your PR:

Applying the patch and rebasing the code consistently resulted in conflicts. Our internal codebase also contains some adaptations for new printer models, including a replacement of the backend library used for Boolean operations.
The current behavior of the color-preserving Boolean operations and mesh simplification does not fully match our expected interaction and processing requirements. For example, we would like to add a checkbox to control whether color information should be preserved. Also, for an A - B Boolean operation, the color information from object B should not be preserved. In addition, these changes still need to go through our internal testing process.

After discussion, we decided to integrate the changes into our internal project while preserving your authorship information. Once the implementation and testing are completed internally, we will synchronize the changes back to the BambuStudio project on GitHub.
Thank you again for your contribution and for the work you put into this PR.

Stone Li

PR #11455, opened 2026-07-09
fix(text): resolve weight-qualified and localized font names on macOS

newrise0410:
## Problem

On macOS the Text tool validates every face name with `wxFontEnumerator::IsValidFacename()`, which only knows font *family* names. This breaks several real cases:

1. **Weight-qualified names** stored in projects (e.g. `Pretendard Black` — on Windows GDI lists these as families, so cross-platform projects carry such names). Loading silently falls back to Arial, and for CJK text the backup-font substitution then makes it look like "the font never changes".
2. **Localized family names** (e.g. `나눔고딕`) fail the same check although CoreText resolves them.
3. Re-editing a text volume could **overwrite its font** with an app-config style matched by style name during style recovery.
4. Weights packed in a **TrueType collection** (`AppleSDGothicNeo.ttc` has 18 faces) were unreachable: only index 0 of a .ttc was ever rendered.

## Fix

- `GLGizmoText::select_facename()`: on macOS skip the `IsValidFacename` gate and let font-file resolution decide; unresolvable names still fail as before.
- `WxFontUtils`: when the family-attribute descriptor yields no file URL, fall back to `CTFontCreateWithName` (matches full / PostScript / localized names) and verify the resolved font actually carries the requested name, so unknown names are not silently substituted by the system default.
- `GLGizmoText::load_init_text()`: re-apply the volume's recorded face after style recovery.
- Enumerate weight-qualified face full names into the font list on macOS (parity with Windows GDI); relax the same enumerator gate in the font preview job.
- For .ttc collections, find the selected face's index by matching its PostScript name (name table id 6, read via stb_truetype) against the fonts in the file, and store it in `FontProp::collection_number`.

## Testing (macOS 15, arm64)

- Project saved with `font_name="Pretendard Black"` now reloads and re-embosses with the real Black weight (previously Arial → backup-font fallback).
- `Apple SD Gothic Neo Heavy` / `UltraLight` render distinct weights from the same .ttc (previously always Regular).
- Localized names (`나눔고딕`) select correctly.
- Nonexistent names still fail cleanly (verified `CTFontDescriptorCopyAttribute(kCTFontURLAttribute)` returns NULL → no silent substitution).
- Windows/Linux code paths unchanged (`IsValidFacename` gate kept there).

### DIFF
--- a/src/slic3r/GUI/Gizmos/GLGizmoText.cpp
+++ b/src/slic3r/GUI/Gizmos/GLGizmoText.cpp
@@ -28,6 +28,11 @@
 #include <wx/hashmap.h>
 #include <wx/utils.h>
 
+#ifdef __APPLE__
+#include <CoreText/CoreText.h>
+#include <wx/osx/core/cfstring.h> // wxCFStringRef
+#endif
+
 #include <numeric>
 #include <codecvt>
 #include <boost/log/trivial.hpp>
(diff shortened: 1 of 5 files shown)


Haidiye00:
Hello, please use some third-party software to record the GIF and provide the pre modification and post modification effects. Thank you

PR #11456, opened 2026-07-09
feat(text): filter font list to faces supporting the current text (macOS)

newrise0410:
> Stacked on #11455 (includes its commit) — the first commit here is that PR; this one adds the filter feature on top.

## Motivation

The Text tool font list shows every installed face (several hundred entries), but for non-Latin text most of them cannot render the input at all: the emboss code silently substitutes glyphs from the bundled backup fonts, so many users perceive it as "changing the font does nothing" (for Korean, everything renders as Nanum Gothic).

## Feature

A checkbox under the font combo — **"Only fonts for current text"**:

- hides faces that cannot render the characters of the current text; with Korean text this cuts the list from ~470 faces to the actual Korean-capable ones,
- composes with the existing name-search filter,
- empty text shows the full list,
- state persists in app config (`text_only_supported_fonts`).

Coverage is read from each font's character set via `CTFontCopyCharacterSet` — the same cmap table the emboss engine reads glyph outlines from — so no font file is opened. All installed faces are checked in ~15 ms, and the result is recomputed only when the set of distinct characters in the text changes (sampled, max 16 distinct chars).

macOS only for now; the checkbox is not shown on other platforms.

## Testing (macOS 15, arm64)

- Korean input + filter on → list reduces to Korean-capable faces (AppleGothic, Apple SD Gothic Neo weights, Nanum family, Pretendard, Noto Sans KR, …); Latin-only and script-specific fonts (Arial, Noto Sans Gothic, …) are hidden.
- ASCII input → filter has no visible effect (all faces pass).
- Verified check correctness against glyph lookup with stb_truetype on 11 known fonts (0 mismatches).

### DIFF
--- a/docs/superpowers/specs/2026-07-09-font-list-filter-design.md
+++ b/docs/superpowers/specs/2026-07-09-font-list-filter-design.md
@@ -1,0 +1,46 @@
+# Text tool: "only fonts supporting current text" filter
+
+Date: 2026-07-09 · Status: approved by user (option 1)
+
+## Problem
+
+The text-emboss font combo lists every installed face (~470 after the
+macOS weight-name additions). Most cannot render 한글; picking one silently
+falls back to the bundled backup font (Nanum Gothic), which reads as
+"the font does not change".
+
+## Design
+
+A checkbox under the font combo: **입력 문구 지원 폰트만** (only fonts
+supporting the current text). When enabled, the combo hides faces that
+cannot render the characters of the current text. It composes with the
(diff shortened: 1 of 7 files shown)


Haidiye00:
Hi, could you use a Mac GIF recording software or screen capture software to show the comparison between before and after your changes? Thanks.

Haidiye00 (commented):
Hello, please review the comments above and merge the patches into one. Thank you

PR #11459, opened 2026-07-09
ENH: Update Preferences UI to split settings to multiple page.

mpaperno:
This is a full re-work of the Preferences dialog.

Properties-General-windows-dark

There is also a lot of code cleanup and refactoring involved, primarily to eliminate a lot of duplication and verbosity.  In fact there are negative lines of code in the `Preferences` class changes, even though I added a lot of line breaks and such for easier reading.  

Please see the commit notes for more technical details.

**Pages and property sorting**

How to organize & sort all the options and what to name the pages and section titles is of course very subjective.  I did put thought and time into this, with quite a few iterations.  But I'm completely open to suggestions and changes here!  What I think is logical may seem chaotic to others.  Of course this will evolve over time anyway, but the more sensible it is to start with, the better.

The good news is that it is very easy to re-organize anything as desired.  Each individual setting is contained in one block of code (both creation and insertion into the layout). Each "page" can also easily be embedded into another page, w/out needing to move code around.

**Navigation style choice**

I decided to utilize the existing `Tabbook` control for the paging and navigation, though this required making it more flexible for styling (it is otherwise used as the pager/nav on Device and Calibration pages). So the first commit in this branch does that (the appearance on Device/Calibration pages is unchanged). This could be re-used again in the shortcuts dialog, for example.  The styling is basically copied from that shortcuts dialog, in fact.

At first I did try to use `wxListBook` and `wxTreeBook` but the styling options for the navigation buttons were severely limited... I was spending more time fighting the default styles than anything.  They just look and act like data lists, not menus. No way to add padding, etc.

I chose the sidebar design vs. horizontal tabs across the top because this will be a lot more flexible.  Tabs are quickly limiting horizontally, and scrolling or stacking tabs is just awkward. The navigation sidebar will already scroll vertically if needed, so we're not limited in that regard either.

**Modified strings**

I did end up changing some label texts (grammar, capitalization, inconsistent line endings)... I'm not sure if this was really a great idea because I realize it invalidates translations. I don't know what BL considers "best practices" in these cases. I'm going to assemble a list of modified strings, and post it here for review.  I could then revert those changes in the code and perhaps make them in the English translation file instead.  There are also a lot of new strings to translate (page names & tips, new section titles, etc).  **Please advise on best way forward here.**

**ADDED:** Here is the list of modified strings.
This list only includes changes to strings that were and are still being used in the Preferences UI. Not all the additions and removals (though I have that as well, if it helps). Fuzzy matching would probably cover most of these (if that's being utilized).

<details>

```
- Auto plate type
+ Remember plate type

- Auto-fill previously logged-in accounts.
+ Auto-fill previously logged-in accounts

- Enable gamma correction for the imported obj file
+ Enable gamma correction for imported OBJ files

- Filament Manager
+ Enable Filament Manager

- Improve rendering performance by lod
+ Improve rendering performance by LOD

- Keep liveview when printing.
+ Keep liveview when printing

- Multi-device Management(Take effect after restarting Studio).
+ Multi-device Management

- No warnings when loading 3MF with modified G-codes
+ Warn when loading 3MF with modified G-codes

- Open full screen camera view on active monitor only.
+ Open full screen camera view on active monitor only

- Perform gamma correction on color after importing the obj model.
+ Perform gamma correction on color after importing the OBJ model.

- Remove the restriction on mixed printing of high and low temperature filaments.
+ Remove the restriction on mixed printing of high and low temperature filaments

- Set grabber size for move,rotate,scale tool.
+ Set grabber size for move, rotate, and scale tools.

- Show history on the home page
+ Show MakerWorld history

- Show online staff-picked models on the home page
+ Show online staff-picked models

- Show the step mesh parameter setting dialog.
+ Show the STEP mesh parameter setting dialog

- Support beta version update.
+ Support beta version update

- Take effect after restarting Studio
+ Takes effect after restarting Studio.
```

</details>

**Compatibility**

Tested on all 3 major OSs -- Win 10 & 11, MacOS 26, Linux... only on 1 Debian-based distro (Manjaro) with Ubuntu-24 build from GitHub CI -- Flatpack not tested.  I was pleasantly surprised about how consistent it looks across all those systems though.

Although there's a lot of refactoring, and the `Preferences.cpp` diff certainly reflects that (sorry!), there really should be no functional changes (except as mentioned in the commit notes). Most of the refactor was just removing literally duplicated code (like creating the same UI element types over and over).  Not saying a regression issue is impossible here, but this was all done very carefully and deliberately.

<details>

<summary>More screenshots here</summary>

**Windows:**
Properties-General-windows-light
Properties-online-windows-dark
Properties-Projects-windows-dark
Properties-Filess-windows-dark
Properties-3Dview-windows-dark

This is from an "internal" build so it's showing the extra options (see below from a release build):
Properties-Advanced-internal-windows-dark

Showing button hover highlight with tooltip:

Properties-General-tooltip-windows-light

**MacOS:**

Properties-General-macos-dark
Properties-Projects-macos-dark
Properties-Advanced-macos-dark

**Linux** (Manjaro Cinnamon dark theme, running Ubuntu-24.04 build from GitHub CI):

Properties-General-linux-dark
Properties-Online-linux-dark
Properties-3Dview-linux-dark

</details>

**Example of adding a new page**

This change, while overdue, was especially inspired by the conversation in #10234 and how that adds a lot of new options to the already-long list.  

<details>
<summary>So here's what it looks like now in my other development version:</summary>

Properties-Camera-windows-dark

The "Camera" section currently with one option in the "3D View" page moves here entirely.
</details>

Thanks for your consideration!
-Max

### DIFF
--- a/src/slic3r/GUI/Preferences.cpp
+++ b/src/slic3r/GUI/Preferences.cpp
@@ -1,75 +1,231 @@
 #include "Preferences.hpp"
-#include "OptionsGroup.hpp"
 #include "GUI_App.hpp"
 #include "MainFrame.hpp"
+#include "OpenGLManager.hpp"
 #include "Plater.hpp"
 #include "MsgDialog.hpp"
 #include "I18N.hpp"
 #include "UxProgramTermsDialog.hpp"
 #include "libslic3r/AppConfig.hpp"
-#include <wx/notebook.h>
-#include "Notebook.hpp"
 #include "ReleaseNote.hpp"
-#include "OG_CustomCtrl.hpp"
+#include "Tabbook.hpp"
 #include "fila_manager/wgtFilaManagerFeature.h"
-#include "wx/graphics.h"
-
-#include <wx/listimpl.cpp>
-#include <map>
 #include "Gizmos/GLGizmoBase.hpp"
-#include "OpenGLManager.hpp"
+#include "Widgets/Button.hpp"
+#include "Widgets/ComboBox.hpp"
+#include "Widgets/CheckBox.hpp"
+#include "Widgets/LinkLabel.hpp"
+#include "Widgets/RadioBox.hpp"
+#include "Widgets/StaticLine.hpp"
+#include "Widgets/TextInput.hpp"
+#include <wx/listimpl.cpp>
+
 #ifdef __WINDOWS__
 #ifdef _MSW_DARK_MODE
 #include "dark_mode.hpp"
 #endif // _MSW_DARK_MODE
 #endif //__WINDOWS__
 
+#define DESIGN_GRAY900_COLOR      wxColour(38, 46, 48)
+#define DESIGN_GRAY800_COLOR      wxColour(50, 58, 61)
+#define DESIGN_GRAY600_COLOR      wxColour(144, 144, 144)
+#define DESIGN_GRAY400_COLOR      wxColour(166, 169, 170)
+#define DESIGN_CONTENT_BG_COLOR   *wxWHITE
+#define DESIGN_SIDEBAR_BG_COLOR   wxColour(241, 241, 241)
+#define DESIGN_HIGHLIGHT_COLOR    wxColour(0, 174, 66)  // BBL green
+
+#define DESIGN_PAGE_PADDING          FromDIP(12)  // padding around page content
+#define DESIGN_ITEM_SPACING          FromDIP(3)   // vertical spacing between items
+#define DESIGN_SECTION_SPACING       DESIGN_PAGE_PADDING  // vertical spacing between titled sections on a page
+#define DESIGN_ITEM_INDENT           FromDIP(23)  // left side offset for all editable properties, indent relative to section titles
+#define DESIGN_INPUT_PAD             FromDIP(8)   // horizontal space between label(s) and input widget
+
+#define DESIGN_LABEL_W               100          // common width of labels on left side of input (forms an even column)
+#define DESIGN_COMBO_W               140
+#define DESIGN_COMBO_LRG_W           160
+#define DESIGN_INPUT_W               100          // default text box width
+#define DESIGN_BUTTON_RADIUS         12
+#define DESIGN_BUTTON_SIZE           wxSize(58, 22)
+
 namespace Slic3r { namespace GUI {
 
 WX_DEFINE_LIST(RadioSelectorList);
-wxDEFINE_EVENT(EVT_PREFERENCES_SELECT_TAB, wxCommandEvent);
 
-#define PreferenceBtnSize wxSize(FromDIP(58), FromDIP(22))
-class MyscrolledWindow : public wxScrolledWindow {
+// Scrollable panel container for property pages.
+class ScrolledPanel : public wxScrolledWindow
+{
+    bool m_has_top_title { false };
+    wxBoxSizer *m_content_sizer;
 public:
-    MyscrolledWindow(wxWindow* parent,
-        wxWindowID id = wxID_ANY,
-        const wxPoint& pos = wxDefaultPosition,
-        const wxSize& size = wxDefaultSize,
-        long style = wxVSCROLL) : wxScrolledWindow(parent, id, pos, size, style) {}
+    ScrolledPanel(wxWindow* parent, bool top_level = true) :
+        m_content_sizer{ new wxBoxSizer(wxVERTICAL) },
+        wxScrolledWindow(parent, wxID_ANY, wxDefaultPosition, wxDefaultSize, wxNO_BORDER | wxTAB_TRAVERSAL)
+    {
+        if (top_level)
+            SetScrollRate(FromDIP(15), FromDIP(15));
+        SetBackgroundColour(DESIGN_CONTENT_BG_COLOR);
+        SetSizer(new wxBoxSizer(wxVERTICAL));
+        GetSizer()->Add(m_content_sizer, 0, wxEXPAND | wxALL, top_level ? DESIGN_PAGE_PADDING : 0);
+    }
 
-    bool ShouldScrollToChildOnFocus(wxWindow* child) override { return false; }
+    wxSizerItem* Add(wxSizerItem *item)
+    {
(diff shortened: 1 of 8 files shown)


tonghao-bbl:
Hi @mpaperno Thank you again for this great PR. we wanted this for a long time and we just got some time to work on this.
So we have planed this since last sprint, and finally did it just days ago. The QA team is testing on this and everything goes well. I hope you will like the new UI of the preference page. Please let us know if you have any suggestions on the new UI when its shipped.

as said we can not merge this PR. we should have a roadmap for this project.

mpaperno:
Hi @tonghao-bbl ,

Well, thanks for letting me know, but... Bummer.  I hope you cleaned it all up as well.  It was a mess of redundancy.  I hope your team can at least get some ideas from my changes. And I just pushed a minor update too.

> I hope you will like the new UI of the preference page. Please let us know if you have any suggestions on the new UI when its shipped.

Seems a bit late at that point.  Like this PR.  :)

> we should have a roadmap for this project.

Indeed.  Quite frustrating, especially since I had specifically asked if anyone was working on it (in the camera settings PR, which is being ignored after getting encouraged).  This is the 2nd time that has happened (that I've asked before doing anything).

-Max

mpaperno:
> I hope you will like the new UI of the preference page. Please let us know if you have any suggestions on the new UI when its shipped.

Hi @tonghao-bbl ,

I hope you meant that. I'm sorry, but I have almost nothing positive to say about the new design or implementation.  It is a temporary workaround at best, in my opinion.  A step backwards in some ways.

My comments aren't directed at you or anyone personally.  These are my professional opinions and I'm not trying to be especially polite about them.  No offense is intended.

**Tabs**

As I mentioned in the PR, the choice of using tabs is very limiting. Adding new tabs increases horizontal space, but this space isn't used for the content, or the content becomes too wide.  This fact makes only a few tabs practical, which eventually leads back to the same problem with scrolling over-filled pages.

Compare that with the sidebar layout where the number of pages could be increased to at least a dozen before the navigation scrolls, leading to less-cluttered content which also doesn't need scrollbars.

Case in point - I added camera controls to the 3D tab (in #10234), and now the page scrolls again. There's no reason that couldn't be an entirely separate page, as presented in this PR.  Much nicer to use.

There's a reason no application of any real complexity uses tabbed preferences panels anymore.  Also a nice convention is to remember the last used page next time the user opens the dialog, which my version did.  If a user needs to switch tabs _and_ scroll to see a setting several times in a row, that's very annoying.

**Checkbox Alignment**

Having the checkboxes to the right of the label is poor UX, at least the way it is implemented.  When the labels are shorter, there is a large visual gap between the label and corresponding checkbox.  With several of these close together, it becomes awkward to trace which box is which.

This alignment works better in the other UI elements (like all profiles) because the width of the labels is limited, with longer ones wrapping, so there is almost never that large gap to the right of the label.

Also the periods on the ends of some of the labels make even less sense now.  That was one of the things I normalized in my version.  And the labels should show tooltips, like the other labels do.

**Dialog Reminders**

Punting all "reset my choice" for dialogs into a single button is going backwards.  Compare to how I got rid of those buttons by actually letting the user choose what they want to do for each situation in the first place.  Instead of having to "reset" some dialog so that next time it pops up so _then_ one can choose what to do.  Why not just let the user choose the preference in the first place?  Makes no sense.  And "one reset to rule them all" makes less sense.  And why is it on every tab?  It looks as if it may only reset warnings relevant to the selected tab, and adds clutter.

**ADDED**: Also the prompt on "reset my choice" is wrong -- the changes do take effect immediately, they're just not reflected in the Preferences panel itself until that it re-opened.

**Fixed Size Dialog**

Why?  Compare to how my version auto-adjusts to fit contents (with a max. size based on display the dialog is actually on), and is user-resizable in case scrollbars do appear unnecessarily. 

Fixed size makes it necessary to "reserve" space for the "Developer" tab which only appears on internal builds, which is just awkward for several reasons.

**Technical**

Tabbing focus order is still messed up, partly because of the awkward two-step item creation being confusing.

Redundant code still all over the place with almost nothing cleaned up. Even though _some_ effort was put into centralizing a few design features (widths/padding). For example the padding around each line of settings, moved to a single `flags` declaration, which is nice, but why does every page need to define those separately?  (I moved all that to the scrollable pane helper class itself.) Finally some central theme colors though, so that's a positive!

Yet another tab book control, just for this?  I don't think people realize how many custom book controls are already in the code.  More bloat, the last thing this code base needs.

---

What I have presented in this PR and what was published in the latest beta are barely comparable.  If someone really decided the current version is better, I'm shocked.  I realize the timing was awkward, but it wasn't too late.  I guess that no one even read this PR or commit notes.

Regards,
-Max

tonghao-bbl:
Hi @mpaperno I really appreciate the feedback personally and I always think you are very professional.

**Tabs design**
This is from the design team and I think they have some good reason to do this. However, they might made a bad design, especially in the cases you mentioned. I think we are always open to fixing any mistakes made before, so lets see how things will go and the feedback from other people, hope you can understand this.

I've noticed #10234 but its other team's work, so I can't say anything about this. If the preference works bad for the new options we can talk about this then.

**Checkbox Alignment**

This is an evidence of ` However, they might made a bad design`  I said before. I did have some words on this and the design team ignored it. However, many people have noticed this and we have decided to make a hover background on the row. So some optimizations on the preference page will be available in next version, including the long label issue you mentioned.

**Dialog Reminders**

I don't like the new design either, and I have talked to the team too.  However, it seems many people don't like the reminders and want close them once for all. So again, we have to wait and see how people thinks about this design.

**Fixed Size Dialog**

I don't like this design from the design team but I think you can understand every team has its own responsibility. When I don't like something I will talk with them, but I have to respect their work. Maybe I'm wrong, right? The design team decides this until users says we don't like this, we need something else.

**Technical**

I feel bad about this, not because you pointed this out, but because I know this. I planed to do the clean up but got some issues in other place,so the time is limited. PR is welcome here.

---

As for the issues that can be optimized a PR is always welcome and I will try best to review it. You are an professional engineer and we always want the contributions from people like you. Thank you

Issue #11349, opened 2026-06-29
Feature Request: Improve AMS Filament Selection with Favorites, Filters and Search


MrDrCreates:
# Feature Request: Improve AMS Filament Selection with Favorites, Filters and Search

## Related Issues

This request is **complementary** to #11278 and #10445.

While those issues focus on **profile management** and **filament assignment workflows**, this proposal specifically addresses the **usability of the AMS material selection dialog** as the number of available filament profiles continues to grow.

---

# Summary

The current AMS filament selection dialog becomes increasingly difficult to navigate as new materials, manufacturers and preset profiles are added.

Although Bambu Studio provides an excellent collection of predefined filament profiles, most users regularly work with only a small subset of them. Every spool assignment therefore requires scrolling through a long list of materials that are rarely or never used.

This request proposes several optional usability improvements that would greatly simplify everyday workflow **without changing the existing behavior for users who prefer the current complete list.**

---

# Current Situation

When assigning a spool to an AMS slot, users must browse through a long list containing:

* numerous PLA variants
* numerous PETG variants
* specialty materials
* manufacturer-specific profiles
* Generic profiles

As Bambu Studio continues to add new filament profiles, this list keeps growing.

For many users who primarily print with only a few materials (for example PLA, PETG HF and TPU), most entries are never selected.

This leads to:

* unnecessary scrolling
* slower workflow
* visual clutter
* increased likelihood of selecting the wrong profile

---

# Proposed Improvements

### ⭐ Favorites

Allow users to mark filament profiles as favorites.

Favorites should always appear at the top of the list.

---

### 🔍 Search

Add a search field allowing users to quickly filter by keywords such as:

* PETG
* TPU
* eSUN
* HF
* Transparent

Searching is significantly faster than scrolling through dozens of entries.

---

### 👤 Show only my profiles

Add an optional filter to display only:

* user-created profiles
* duplicated profiles
* custom manufacturer profiles

This would allow users to work exclusively with their own profiles while keeping system presets available when needed.

---

### 👁 Hide unused preset profiles

Add an optional checkbox such as:

> Hide unused preset profiles

This would simply filter the visible list without deleting any built-in profiles.

---

### 🕒 Recently Used

Display the last 5–10 selected filament profiles above the complete list.

Many users repeatedly assign the same few materials during normal operation.

---

### 📂 Simplified View

Example:

```
⭐ Favorites
🕒 Recently Used
👤 My Profiles

-------------------------

Show all profiles ▼
[...]

PR #11470, opened 2026-07-10
Add search filter to AMS filament selection

BenJule:
## Summary

This PR adds a search field to the AMS filament selection dialog.

## Why

When many filament presets are available, especially with user/custom presets, the AMS filament dropdown can become difficult to navigate.

This change allows users to quickly filter the existing filament list by typing part of the filament name, material, or vendor as shown in the dropdown.

Related to #11349

## Implementation

- Add a search field above the existing AMS filament combo box.
- Keep the existing filament collection, sorting, selection, and assignment flow unchanged.
- Filter only the displayed combo box items.
- Preserve the current selection when it is still visible after filtering.

## Test plan

- Build Bambu Studio.
- Open the Device tab.
- Edit a third-party AMS slot.
- Confirm that the filament search field is visible.
- Type part of a filament name, vendor, or material.
- Confirm that the AMS filament list is filtered.
- Select a filtered filament and confirm the slot assignment still works.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -160,7 +160,17 @@ void AMSMaterialsSetting::create_panel_normal(wxWindow* parent)
     m_comboBox_filament = new ::ComboBox(parent, wxID_ANY, wxEmptyString, wxDefaultPosition, AMS_MATERIALS_SETTING_COMBOX_WIDTH, 0, nullptr, wxCB_READONLY);
 #endif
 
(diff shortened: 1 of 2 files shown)


hemai249:
Hi！@BenJule Thank you so much for your support! We’ve come up with a new UI for selecting consumables that makes it really easy to find what you need without needing filters. We’ll be rolling this out in a future version.

BenJule:
Thank you for the update. That sounds good.

Since the new consumable selection UI will cover the use case addressed by this PR, I will close it to avoid duplicating the upcoming implementation.

Thanks for reviewing it and for letting me know.

PR #11471, opened 2026-07-10
Prioritize user filament presets in AMS selection

BenJule:
## Summary

This PR prioritizes user/custom filament presets in the AMS filament selection list.

## Why

When many system and user filament presets are available, custom calibrated filaments can be harder to find in the AMS filament dropdown.

This change keeps the existing displayed filament names unchanged, but sorts user/custom presets before system presets.

Related to #11349

## Implementation

- Keep existing filament collection and assignment logic unchanged.
- Track whether each filament item comes from a user preset.
- Sort user/custom presets before system presets.
- Preserve the existing vendor/type/name sorting inside each group.

## Test plan

- Build Bambu Studio.
- Create or import at least one custom filament preset.
- Open the Device tab.
- Edit an AMS slot.
- Confirm that user/custom filament presets appear before system presets.
- Confirm selecting a filament still updates Brand / Material / Color and PA calibration fields as before.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -951,7 +951,8 @@ void AMSMaterialsSetting::get_filaments_info(const MachineObject*               
                                              wxArrayString&                           filament_items,
                                              std::map<std::string, FilamentInfos>&    map_filament_items,
                                              std::unordered_map<wxString, wxString>&  query_filament_vendors,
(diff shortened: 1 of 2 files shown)


hemai249:
Hi! @BenJule  We’ve designed a new UI for consumable selection that makes it much easier for users to find what they need. As a result, the sorting feature is no longer necessary. This new interface will be rolled out in a future version.

BenJule:
Closing this PR for now. The implementation may be revisited in a new PR if needed.

Issue #10643, opened 2026-05-09
K-Factor/Filament Profile doesnt show after start


Solilord:
### Bambu Studio Version

every, up to 2.6.1 beta

### Where is the application from?

Bambu Lab GitHub releases

### OS version

Win 11 25H2

### Printer

Bambu Lab P1S with AMS 2 Pro

### How to reproduce

After start: go to "Device" and check the K-Factor / Filament Profile.

### Actual results

only Standard (K: 0.020) is available
BUT: If i go to the calibration page and click on: "Manage Results" (and just let the window open)
close it directly after and go to "Devices" again -> Filament Profile is theren, and set, with the correct K-Factor.
After restart of Bambu Studio only "Standard" Filament profile is shown again...

### Expected results

Filament Profile should always be visible and selected.

### Project file & Debug log uploads

[studio_Sun_May_03_19_04_04_55964_enc.log.0.zip](https://github.com/user-attachments/files/27547077/studio_Sun_May_03_19_04_04_55964_enc.log.0.zip)

### Checklist of files to include

- [x] Log file
- [ ] Project file

PR #11473, opened 2026-07-10
Request PA history for main AMS extruder

BenJule:
## Summary

This PR ensures PA calibration history is requested for AMS slots bound to the main extruder.

## Why

The AMS filament settings dialog requests PA calibration history when the calibration table has not been loaded yet. However, the request was only sent when the resolved extruder id was greater than zero.

For AMS slots bound to the main extruder, the resolved extruder id can be `0`, which means the PA history request could be skipped and the PA profile dropdown could remain limited to the default entry until another UI path loads the history.

Related to #10643

## Implementation

- Remove the `ext_id > 0` gate before requesting PA history.
- Reuse the same fallback style already used in the dialog when resolving the AMS extruder nozzle diameter.
- Fall back to extruder `0` if the resolved extruder id is not available.
- Keep the existing pending refresh flow via `m_pa_data_pending` and `TryRefreshPAProfiles()` unchanged.

## Test plan

- Build Bambu Studio.
- Open the Device tab.
- Edit an AMS slot bound to the main extruder.
- Confirm that PA calibration history is requested when it is not loaded yet.
- Confirm that the PA profile dropdown is refreshed after history becomes ready.
- Confirm that K/N values still update when selecting a PA profile.


BenJule:
Closing this as a duplicate of #10850. The same PA history request issue for main extruder id 0 was already fixed there.

Issue #10515, opened 2026-05-02
Filament Manager is too disorganized


trinity1561:
**Is your feature request related to a problem? Please describe.**
It's very had to see what filament I have with the Filament Manager.  There's no clear order for how the filaments are listed.  The button "Group" is confusing - I thought it would allow you to group filaments by brand or type but that seems to group all of the same brand/type/color which I find useless.  I rather have the option to group filaments by type or by brand, maybe even color.  It would also be nice to be able to sort the filaments!

**Describe the solution you'd like**
1)  Have a "Group" button that allows you to group filaments by Brand or Type or Color.  That way it's easier to view.  Maybe it's just easier to view it as a table like a spreadsheet.

2) Have an option to sort by filament name, type, color, date added, and filament remaining.

**Describe alternatives you've considered**
Move the hex color code and add the SN for the rolls on the same line as filament diameter.  This will help if you have multiple rolls of the same brand/type/color. (see attached image)

**Additional context**

<img width="535" height="269" alt="Image" src="https://github.com/user-attachments/assets/12178d6e-8736-44ca-90c8-81bef22dbadf" />

BenJule:
I opened a small PR that improves the default Filament Manager organization: https://github.com/bambulab/BambuStudio/pull/11475

It does not add a full sort/group UI yet, but it changes the spool list display order from internal spool-id order to a predictable Brand → Material Type → Series → Color order.

PR #11475, opened 2026-07-10
Sort filament manager spools by brand and material

BenJule:
## Summary

This PR makes the Filament Manager spool list use a more predictable display order.

## Why

The Filament Manager currently returns spools in the internal map order, which is based on spool ids. This can make the list feel disorganized when users have many filaments.

Related to #10515

## Implementation

- Keep the stored spool data unchanged.
- Sort the JSON spool representation before returning it.
- Sort spools by:
  - Brand
  - Material type
  - Series
  - Color name
  - Color code
  - Created date
  - Spool id
- Use case-insensitive string comparison for the visible text fields.
- Do not change any cloud sync fields, spool ids, or saved data schema.

## Test plan

- Build Bambu Studio.
- Open the Filament Manager.
- Add or load multiple spools with different brands and material types.
- Confirm the list is grouped in a predictable brand/material order.
- Confirm adding, editing, removing, syncing, and saving spools still works as before.

### DIFF
--- a/src/slic3r/GUI/fila_manager/wgtFilaManagerStore.cpp
+++ b/src/slic3r/GUI/fila_manager/wgtFilaManagerStore.cpp
@@ -1,4 +1,7 @@
 #include "wgtFilaManagerStore.h"
+#include <tuple>
+#include <cctype>
+#include <algorithm>
 #include "libslic3r/Utils.hpp"
 
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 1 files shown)


JingyuSunBambu:
Thank you very much for your valuable feedback on Bambu Studio. As mentioned in our previous reply, we have optimized the sorting function for filament names. For users with a large collection of filaments, making full use of the grouping, sorting, and filtering features in the Filament Manager will help keep everything well-organized.

BenJule:
Thanks for the clarification. Since filament-name sorting has already been optimized and the Filament Manager provides grouping, sorting, and filtering controls, this additional backend sorting is no longer necessary.

I’ll close the PR. Thanks for reviewing it.

Issue #6265, opened 2025-03-28
[Feature Request] Tooltip Showing Filament Colour Name in AMS View


TheCableGuy99:
**Is your feature request related to a problem? Please describe.**
As there’s an ever-increasing number of filaments and colours, it can be difficult to identify the exact colour of the filament currently loaded, as it’s only printed on the original box—which usually gets thrown away once opened. If you have multiple AMS units, like I do, you have to constantly keep a manual record of exactly which filament and colour is in each slot. The slicer only identifies the filament type and gives a visual reference of the colour, which can easily be confused with other similar shades.

**Describe the solution you'd like**
Where Bambu Studio identifies the filament type and shows a visual reference of the colour, a simple balloon tip popup when hovering over the colour that gives the filament's name would be great. No additional redesign is needed, as it would just be a tooltip.

**Describe alternatives you've considered**
I currently keep manual records, which is a pain. I’ve also tried using labels, but it’s a lot of work and hassle to create and attach them to each spool—and to keep doing it as things change.

**Additional context**
The RFID tag already contains all the filament information, so this should be easy to implement. I’d be really grateful if this could be added.

BenJule:
I opened a small PR for this: https://github.com/bambulab/BambuStudio/pull/11477

The AMS slot tooltip already showed the material name and calibration info. This extends it with the filament color hex value, including all available color values for multicolor / gradient filaments. The implementation keeps AMS tray data and layout unchanged.

PR #11477, opened 2026-07-10
Show AMS filament color in slot tooltip

BenJule:
## Summary

This PR adds filament color information to the AMS slot tooltip.

## Why

AMS slots already show a visual color swatch, but similar filament colors can be hard to distinguish by sight alone.

Related to #6265

## Implementation

- Extend the existing AMS slot tooltip.
- Keep the current material and K-value tooltip information.
- Add the slot color as a hex value.
- Show all color hex values for multicolor / gradient filaments.
- Do not change AMS tray data, RFID parsing, or layout.
- Avoid showing color-only tooltips for slots without a material name.

## Test plan

- Build Bambu Studio.
- Open the Device tab with an AMS connected.
- Hover over AMS slots with normal single-color filaments.
- Confirm the tooltip includes the material and color hex value.
- Hover over multicolor / gradient filaments.
- Confirm the tooltip includes all available color hex values.
- Confirm empty or unknown slots do not show a misleading color-only tooltip.
- Confirm the existing K-value tooltip line still appears when applicable.

### DIFF
--- a/src/slic3r/GUI/Widgets/AMSItem.cpp
+++ b/src/slic3r/GUI/Widgets/AMSItem.cpp
@@ -1090,6 +1090,32 @@ void AMSLib::render(wxDC &dc)
     }
 }
 
+wxString AMSLib::get_filament_color_tooltip() const
+{
+    wxString color_text;
+    // External / virtual trays with no colour fall back to the white default
+    // (AMS_TRAY_DEFAULT_COL); report that as "no colour" instead of white.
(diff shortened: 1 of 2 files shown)


XinZhangBambu:
Thanks for the contribution! Could you please also support AMS Lite and external spools, and verify transparent, multi-color, and missing-color cases? Currently, transparent colors may be omitted and missing colors may appear as white.

Issue #10754, opened 2026-05-16
Feature Request - Filament Manager function


CraigLampe:
Problem: We want to manage ALL filament:
- on printer
  - in AMS
  - outside AMS
- off printer (on shelf)

- BambuLab Brand
- Other brand / generic
  - allow adding of other brands "Creality" etc !!

In order to accomplish:
- printer does track useDown of bambulab brand filament
- add track ALL useDown of ALL generic/other filaments
  - may require asking when filament added to AMS (is this X or Y?  How much g on spool (estimated)? assign serial numbers to generic? "new spool" button/checkbox? -defaults to 1000g)
  - AutoRemove at SPOOL END !! (or move down to next 1000g)
  - allow user recovery if data loss
  - "perform inventory" (scheduled?)

Allow set "par level" for a certain type / color
- send notification "time to buy COLOR / TYPE"
- add "green/orange/red indicators" as a color "getting low / critical low"

Allow search for color in UI
- "black" / "red" / "blue" in notes not searchable
- easy first step: enable search "notes" text
- AutoAdd color "NAME" to bambulab filaments to allow SEARCH !!!  -- use actual website name for colors!
- AutoAdd color name from color picker -- if red picked add "red" to searchable
- search by color picker?

Allow choose group by "color family"
- all "reds"
- all "blue" etc

OFF TO A GOOD START!!!!

BenJule:
I opened a small PR for the first search step mentioned here: https://github.com/bambulab/BambuStudio/pull/11484

It extends the existing Filament Manager client-side search to include spool notes and spool IDs, without changing the UI layout, adding columns, or changing the spool schema.

CraigLampe:
Also if ordered on BambuLab -> add automatically to inventory!!!

PR #11484, opened 2026-07-10
Search Filament Manager notes

BenJule:
## Summary

This PR extends the Filament Manager search to include spool notes and spool IDs.

## Why

The current search only checks brand, material type, series, and color name. This makes it difficult to find spools when users store identifying information in the note field.

Related to #10754

## Implementation

- Extend the existing Filament Manager search haystack.
- Include `note`, `notes`, and `spool_id` in the client-side search.
- Do not change the UI layout.
- Do not add new columns.
- Do not change the spool data schema.

## Test plan

- Build Bambu Studio.
- Open the Filament Manager.
- Add or edit a spool with a unique note.
- Search for text that only appears in the note.
- Confirm the matching spool is shown.
- Search for existing brand, material, series, and color values and confirm those still work.

### DIFF
--- a/resources/web/fila_manager/index.js
+++ b/resources/web/fila_manager/index.js
@@ -237,7 +237,15 @@ function getFilteredSpools() {
         if (g_tab === "favorite" && !s.favorite) return false;
         if (g_tab === "ams" && s.entry_method !== "ams_sync") return false;
         if (keyword) {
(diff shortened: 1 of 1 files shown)


BenJule:
Closing in favor of #11505, which contains the current clean implementation.

Issue #1729, opened 2023-05-01
Missing project files remain in the MRU list when not found


julie777:
**Describe the bug**

Bambu Studio tries to open all of the recent projects during startup. 
1. this is not a normal part of startup for an application.
2. I am guessing it is to get the icons to show on the home page. If that is the reason then the icons should be cached along with other such information.

I found this by looking at the log trying to determine the cause of another problem. In addition to the above, when Studio fails to find a recent project it does not remove it from the list. This means that it will try to open the same files and fail every time it is started until every one of those projects has been replaced in the MRU list by another one.

I would expect that you will at least remove files from the MRU list if they cannot be found.

**3mf File for This Bug**
If it is related to slicing, please append the 3mf file. It could be extremely helpful to solve the issue.

**To Reproduce**
4. Click on '....'
5. Scroll down to '....'
6. See error

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS, Windows]

bambu123:
Yes, you can remove them manually. But they are kept in case of reattaching removable storage.

PR #11494, opened 2026-07-10
Fix removing missing recent projects from history

BenJule:
## Summary

This PR fixes removal of missing recent projects from the recent projects history.

## Why

When a recent project file no longer exists, Bambu Studio already shows an error dialog and has code intended to remove the missing entry from the recent projects list.

However, the dialog was created with `wxOK`, while the code checked for `wxID_YES`. As a result, the removal branch could not be reached from that dialog, so missing files stayed in the MRU/recent projects list.

Related to #1729

Complementary to #11413, which adds a separate "Clear recent projects" action. This PR does not clear the full list. It only fixes the existing missing-file removal flow for a selected recent project.

## Implementation

- Change the missing recent project dialog from `wxOK` to `wxYES_NO`.
- Ask the user whether the missing project should be removed from recent projects.
- Reuse the existing `remove_recent_project(...)` helper instead of duplicating persistence/update logic.

## Scope

This PR does not:

- add a new menu item
- clear all recent projects
- change how recent projects are stored
- change thumbnail loading
- change project loading behavior for existing files

## Test plan

- Add or keep a project in the recent projects list.
- Delete or move the project file outside Bambu Studio.
- Click the missing entry from Recent projects.
- Confirm the dialog asks whether to remove it.
- Click Yes.
- Confirm the missing entry is removed from the recent projects list.
- Confirm existing recent project files still open normally.

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -4390,20 +4390,12 @@ void MainFrame::open_recent_project(size_t file_id, wxString const & filename)
     }
     else
     {
-        MessageDialog msg(this, _L("The project is no longer available."), _L("Error"), wxOK | wxYES_DEFAULT);
+        MessageDialog msg(this,
+            _L("The project is no longer available. Do you want to remove it from the recent projects list?"),
+            _L("Error"),
+            wxYES_NO | wxYES_DEFAULT | wxICON_WARNING);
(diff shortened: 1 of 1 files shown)


Haidiye00 (commented):
Hi, please record the effect after the changes, thanks.

Haidiye00 (commented):
Hi, please record the effect after the changes, thanks.

Issue #11495, opened 2026-07-10
Add File menu item to show current project in folder


BenJule:
## Feature request

Bambu Studio should provide a quick way to reveal the currently saved project file in the system file manager.

## Current behavior

Users can save and load projects, but there is no direct menu action to show the folder containing the currently opened `.3mf` project file.

This can be inconvenient when users want to:

- locate the saved project file
- copy it elsewhere
- attach it to a bug report
- verify where a project was saved after using Save As
- manage project files outside Bambu Studio

## Proposed behavior

Add a File menu item:

`File > Show Project in Folder`

The action should be enabled only when the current project has a valid saved `.3mf` path and that file still exists.

When triggered, it should open the current project file location in the system file manager.

## Scope

This should only add a shortcut to reveal the current project file location.

It should not:

- change project saving
- change project loading
- change recent projects
- change exported G-code behavior
- create a new project file
- prompt the user to save unsaved projects

## Why this is useful

This is a small quality-of-life improvement that helps users quickly find the active project on disk without manually searching through folders.

BenJule:
I opened a PR for this: https://github.com/bambulab/BambuStudio/pull/11496

It adds `File > Show Project in Folder`, enabled only when the current project has an existing saved `.3mf` path, and reuses the existing cross-platform folder-opening helper.

PR #11496, opened 2026-07-10
Add menu item to show current project in folder

BenJule:
## Summary

This PR adds a File menu item to reveal the currently saved project in the system file manager.

New menu item:

`File > Show Project in Folder`

## Why

After opening or saving a project, users may want to quickly locate the active `.3mf` file on disk, for example to copy it, attach it to a bug report, or verify where it was saved.

Bambu Studio already has a cross-platform helper for opening folders, so this PR reuses that instead of adding platform-specific logic.

Related to #11495

## Implementation

- Add `File > Show Project in Folder` after `Save Project as`.
- Use `m_plater->get_project_filename(".3mf")` to resolve the current project path.
- Enable the menu item only when the project path is not empty and the file exists.
- Reuse `Slic3r::GUI::desktop_open_any_folder(...)` to reveal/open the containing folder.

## Scope

This PR does not:

- change project saving
- change project loading
- change recent projects
- change exported G-code behavior
- prompt users to save unsaved projects

## Test plan

- Open or save a `.3mf` project.
- Open the File menu.
- Confirm `Show Project in Folder` is enabled.
- Click it.
- Confirm the system file manager opens the project location.
- Start a new unsaved project.
- Confirm `Show Project in Folder` is disabled.
- Delete or move the saved project file externally.
- Confirm `Show Project in Folder` becomes disabled.

### DIFF
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -2931,6 +2931,21 @@ void MainFrame::init_menubar_as_editor()
             [this](){return m_plater != nullptr && can_save_as(); }, this);
 #endif
 
+        append_menu_item(fileMenu, wxID_ANY, _L("Show Project in Folder"), _L("Show current project in folder"),
+            [this](wxCommandEvent&) {
+                if (m_plater) {
+                    const wxString filename = m_plater->get_project_filename(".3mf");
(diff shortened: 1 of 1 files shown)


Haidiye00 (approved):
approve

Haidiye00 (approved):
approve

Issue #11497, opened 2026-07-10
Add copy button to generic error dialog


BenJule:
## Feature request

Bambu Studio error dialogs should provide a simple way to copy the displayed error message.

## Current behavior

When Bambu Studio shows an application error dialog, users often need to report the exact message in an issue or support request.

Currently, users may need to manually select text, retype the message, or attach a screenshot.

## Proposed behavior

Add a `Copy` button to the generic error dialog.

The button should:

- copy the displayed error text to the clipboard
- keep the dialog open
- not change the behavior of the existing `OK` button

## Scope

This request is limited to the generic `ErrorDialog`.

It should not:

- add copy buttons to all message dialogs
- change warning, confirmation, or question dialogs
- change error handling behavior
- change dialog text formatting

## Why this is useful

Exact error messages are important for bug reports and support. A dedicated copy button makes it easier for users to provide accurate diagnostics without screenshots or manual transcription.

BenJule:
I opened a PR for this: https://github.com/bambulab/BambuStudio/pull/11498

It adds a non-closing `Copy` button only to the generic `ErrorDialog`, leaving normal message, warning, confirmation and question dialogs unchanged.

PR #11498, opened 2026-07-10
Add copy button to generic error dialog

BenJule:
## Summary

This PR adds a non-closing `Copy` button to the generic `ErrorDialog`.

## Why

Error dialogs often contain details that are useful for bug reports or support requests. A dedicated copy button makes it easier for users to copy the exact error text instead of taking screenshots or manually retyping it.

Related to #11497

## Implementation

- Extend `MsgDialog::add_button(...)` with an optional `end_modal` argument.
- Keep the default behavior unchanged for existing callers.
- Add a `Copy` button to `ErrorDialog`.
- The copy button writes the stored error message to the clipboard and keeps the dialog open.

## Scope

This PR only affects the generic `ErrorDialog`.

It does not:

- add copy buttons to all message dialogs
- change warning, confirmation, or question dialogs
- change error handling behavior
- change the existing `OK` button behavior

## Test plan

- Trigger or instantiate an `ErrorDialog`.
- Confirm the dialog shows both `OK` and `Copy`.
- Click `Copy`.
- Confirm the error text is copied to the clipboard.
- Confirm the dialog stays open after clicking `Copy`.
- Click `OK`.
- Confirm the dialog closes normally.

### DIFF
--- a/src/slic3r/GUI/MsgDialog.cpp
+++ b/src/slic3r/GUI/MsgDialog.cpp
@@ -134,7 +134,7 @@ void MsgDialog::SetButtonLabel(wxWindowID btn_id, const wxString& label, bool se
     }
 }
 
-Button* MsgDialog::add_button(wxWindowID btn_id, bool set_focus /*= false*/, const wxString& label/* = wxString()*/)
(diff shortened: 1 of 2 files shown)


Haidiye00 (commented):
Hi, please record the effect after the changes, thanks.

Haidiye00 (commented):
Hi, please submit the GIF recording, and merge the patches into one. Thanks.

Issue #11499, opened 2026-07-10
Add copy details button to device error dialog


BenJule:
## Feature request

Device error dialogs should provide a simple way to copy the displayed error details.

## Current behavior

When a printer or device error is shown, users may need to report the exact error message and error code in an issue or support request.

The dialog displays the error level, error code and message, but there is no dedicated button to copy those details.

## Proposed behavior

Add a `Copy Error Details` button to the device error dialog.

The button should copy:

- the displayed error level/title
- the displayed error code
- the displayed error message

The button should keep the dialog open and should not trigger any printer/device action.

## Scope

This request is limited to `DeviceErrorDialog`.

It should not:

- change printer error handling
- change existing action buttons
- close the dialog after copying
- change the HMS/error lookup logic
- affect the generic `ErrorDialog`

## Why this is useful

Exact printer/device error details are useful for bug reports, support requests and troubleshooting. A dedicated copy button makes it easier to share accurate diagnostics without screenshots or manual transcription.

BenJule:
I opened a PR for this: https://github.com/bambulab/BambuStudio/pull/11500

It adds a non-closing `Copy Error Details` button to `DeviceErrorDialog`, copying the displayed error title, code and message without triggering any printer/device action.

PR #11500, opened 2026-07-10
Add copy details button to device error dialog

BenJule:
## Summary

This PR adds a `Copy Error Details` button to `DeviceErrorDialog`.

## Why

Printer/device error dialogs can contain useful troubleshooting details such as the error level, error code and message. Users often need to share those details in bug reports or support requests.

This PR makes that easier by adding a dedicated copy action for the currently displayed device error details.

Related to #11499

## Implementation

- Add a `COPY_ERROR_DETAILS` action button.
- Store the currently displayed error title, error code text and error message.
- Add `Copy Error Details` to the device error dialog button area.
- Copy the stored details to the clipboard.
- Return immediately after copying so the dialog stays open and no printer/device action is triggered.

## Scope

This PR only affects `DeviceErrorDialog`.

It does not:

- change printer error handling
- change HMS/error lookup logic
- change existing printer/device action buttons
- affect the generic `ErrorDialog`
- close the dialog after copying

## Note

This is separate from the generic `ErrorDialog` copy button. `DeviceErrorDialog` is a custom printer/device error dialog and does not use the generic error dialog implementation.

## Test plan

- Trigger or instantiate a `DeviceErrorDialog`.
- Confirm the dialog shows `Copy Error Details`.
- Click `Copy Error Details`.
- Confirm the visible error title/code/message are copied to the clipboard.
- Confirm the dialog stays open after copying.
- Confirm existing action buttons still work as before.

### DIFF
--- a/src/slic3r/GUI/DeviceErrorDialog.cpp
+++ b/src/slic3r/GUI/DeviceErrorDialog.cpp
@@ -6,6 +6,9 @@
 #include "MainFrame.hpp"
 #include "ReleaseNote.hpp"
 
+#include <wx/clipbrd.h>
+#include <wx/dataobj.h>
+
 namespace Slic3r {
 namespace GUI
 {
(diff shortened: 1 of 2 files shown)


Bambu-Pomfret:
Hi @BenJule,

Could we avoid adding this new action to handle it? The action ID has to be kept consistent with firmware and Handy. If Studio reserves this ID now, it will limit the available IDs when firmware and Handy teams add new functionalities later on.

PR #11502, opened 2026-07-10
Preserve text volume center when updating text

### LINKED ISSUE #11423: Text Justification and/or Text Object Center Preservation
When editing a text object, it's position changes if the length of the text changes.

Add a justification setting (left, center, right) to the text controls, defaulting to center. When set to center, the object position of the text object remains the same. In left or right justification, the left or right edge position stays the same. For completeness, and to also support languages with vertical layouts, also add vertical justification, left, right, and center.

Currently, for centered text, one needs to preserve the X and Y positions of the text object in a separate text document, edit the text, then restore those positions. This is clumsy and is better done with software.

**Additional context**
A similar feature might be useful for changing an SVG object's source file. However I have not done this much myself. "Justification" might better be replaced with a 3x3 grid of radio button-like options for SVGs, one for each corner, edge, and center of a square. This would indicate which point to keep constant if the size and shape of the SVG changes.

But even without _any_ interface changes, preserving a text object's center point when the text changes (as SVGs preserve _their_ center point) would solve the majority of people's use cases, and should probably be the default anyway.

BenJule:
## Summary

This PR preserves the transformed center of an existing text volume when its text mesh is regenerated after editing.

Fixes #11423.

## Why

When editing an existing text object, changing the text length can change the local mesh bounding box center. The previous transformation is kept, so the visible text object may shift position after the mesh is replaced.

## Implementation

- Preserve the transformed center for normal text volume updates.
- Capture the old transformed mesh center before replacing the mesh.
- Capture the new transformed mesh center after regeneration.
- Apply a translation correction so the text volume stays centered at the same position.
- Skip this correction for surface text updates, where the update path provides an explicit surface-based transformation.

## Scope

This does not add new justification UI. It implements the minimal behavior requested in the issue: keep the text object's center stable when the text changes.

## Test plan

- Create a text object.
- Edit the text and change its length.
- Confirm the text volume keeps its center position instead of shifting.
- Change font/style values that regenerate the text mesh and confirm the center is preserved.
- Confirm surface text updates still use their explicit surface transformation path.

### DIFF
--- a/src/slic3r/GUI/Jobs/EmbossJob.cpp
+++ b/src/slic3r/GUI/Jobs/EmbossJob.cpp
@@ -147,7 +147,7 @@ void _update_volume(TriangleMesh &&mesh, const DataUpdate &data, const Transform
         if (emboss_shape.has_value() && emboss_shape->fix_3mf_tr.has_value()) volume->set_transformation(volume->get_matrix() * emboss_shape->fix_3mf_tr->inverse());
     }
 
-    UpdateJob::update_volume(volume, std::move(mesh), *data.base);
+    UpdateJob::update_volume(volume, std::move(mesh), *data.base, tr == nullptr);
 }
 
 std::vector<BoundingBoxes> create_line_bounds(const ExPolygonsWithIds &shapes, size_t count_lines = 0)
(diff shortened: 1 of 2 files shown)


Haidiye00 (commented):

PR #11503, opened 2026-07-10
Add Markdown export for preset differences

BenJule:
## Summary

This PR adds a **Copy differences** button to the Compare presets dialog.

The button copies the currently displayed preset differences as a Markdown table, making it easier to share profile differences in GitHub issues, support tickets, forums, and release/debug notes.

## Why

The Compare presets dialog already shows differences between two presets, but those differences are hard to share without screenshots or manual transcription.

Copying the diff as Markdown makes preset comparisons portable and much easier to review.

## Implementation

- Adds a Copy differences button to `DiffPresetDialog`.
- Builds a Markdown table while the existing diff tree is generated.
- Includes preset type, compared preset names, setting label, left value, and right value.
- Escapes Markdown table pipes and converts multiline values to `<br>`.
- Enables the button only when differences are available.

## Test plan

- Open Compare presets.
- Select two different presets.
- Confirm the Copy differences button becomes enabled.
- Click Copy differences.
- Paste into a Markdown editor or GitHub comment.
- Confirm the pasted output is a readable Markdown table.
- Select identical presets and confirm the button is disabled.

### DIFF
--- a/src/slic3r/GUI/UnsavedChangesDialog.cpp
+++ b/src/slic3r/GUI/UnsavedChangesDialog.cpp
@@ -8,6 +8,8 @@
 #include <boost/nowide/convert.hpp>
 
 #include <wx/tokenzr.h>
+#include <wx/clipbrd.h>
+#include <wx/dataobj.h>
 
 #include "libslic3r/PrintConfig.hpp"
 #include "libslic3r/PresetBundle.hpp"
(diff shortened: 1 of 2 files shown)


tonghao-bbl:
@BenJule This PR looks great. 

But you may have noticed in #11311 this dialog can only show one different value at a time. Many presets have more variants, X2D may have up to 6 variants. We are working on this internally, but it's still on UX stage, we hope it will be shipped at the end of the month.

If we merge this PR now you may need to create a new PR for this feature. Also, will the table layout work for the cases with up to 6 variants?

BenJule:
@tonghao-bbl  Thanks for pointing this out.

I updated the Markdown export so it no longer relies on fixed `Left value` and `Right value` columns. It now uses a normalized format with one row per preset or variant:

| Preset type | Setting | Preset / variant | Value |
| --- | --- | --- | --- |

The export helper accepts a variable-length list of labeled values. The current comparison passes the two selected presets, while a future comparison UI can pass up to six variants, or more, without requiring another table-format change.

The preset names are currently used as the row labels.

Validation performed:

- `git diff --check`
- targeted compilation of `UnsavedChangesDialog.cpp.o`
- repeated Ninja invocation returned `no work to do`

A full application build and interactive UI test have not been performed.

PR #11504, opened 2026-07-10
Add modified settings dashboard

BenJule:
## Summary

This PR adds a **modified settings dashboard** to preset tabs.

The dashboard appears when the current tab has modified settings or system overrides and gives users a quick overview of what changed.

## Why

Bambu Studio already marks modified settings in the UI, but there is no clear high-level summary at the top of the settings panel.

This makes it easy to miss that a profile, object, or printer setting has active overrides. A compact dashboard makes the current modification state visible immediately.

## Implementation

- Adds a dashboard panel below the preset toolbar in `Tab`.
- Shows the number of modified settings.
- Shows the number of system overrides.
- Shows the current settings scope.
- Adds a Review button using the existing Compare presets action where available.
- Keeps the dashboard hidden when there are no modifications.
- Deduplicates modified option keys before counting.

## Test plan

- Open a preset tab with no changes and confirm the dashboard is hidden.
- Modify one or more settings and confirm the dashboard appears.
- Confirm the modified settings count updates.
- Confirm system override count is shown.
- Click Review on a normal preset tab and confirm Compare presets opens.
- Enter and exit search mode and confirm the dashboard remains stable.

### DIFF
--- a/src/slic3r/GUI/Tab.cpp
+++ b/src/slic3r/GUI/Tab.cpp
@@ -369,6 +369,36 @@ void Tab::create_preset_tab()
     search_sizer->Fit(m_search_item);
 
     m_search_item->Hide();
+
+    m_override_dashboard = new wxPanel(this, wxID_ANY, wxDefaultPosition, wxDefaultSize);
+    m_override_dashboard->SetBackgroundColour(wxColour(248, 248, 248));
+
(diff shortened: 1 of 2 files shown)


tonghao-bbl:
@Mujin-code what do you think about this?

PR #11506, opened 2026-07-10
Remember recent AMS filament presets per slot

BenJule:
## Summary

This PR remembers recently used filament presets per AMS slot and prioritizes them in the AMS filament selection dropdown.

## Why

Users often assign the same third-party filament presets to the same AMS slots repeatedly. Today, they have to scroll through the full filament list every time.

Keeping a small per-slot recent list makes the common path faster without adding new UI or changing printer/cloud behavior.

## Implementation

- Stores the last 5 confirmed filament preset selections per AMS slot in app config.
- Uses `ams_id` and `slot_id` to keep the history slot-specific.
- Prioritizes recently used presets at the top of the AMS filament dropdown.
- Keeps the existing vendor/type/name sorting for all other entries.
- Only records confirmed selections from editable third-party AMS slots.

## Test plan

- Open the AMS material settings dialog for a third-party AMS slot.
- Select a filament preset and confirm.
- Reopen the same AMS slot and confirm the selected preset is prioritized near the top.
- Select several different presets and confirm the recent order updates.
- Open a different AMS slot and confirm it has its own recent list.
- Confirm normal Bambu/RFID readonly slots still behave as before.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -737,6 +737,13 @@ sCheckFilamentInfo(PresetBundle* preset_bundle,
     return result;
 }
 
+namespace {
+
+void remember_ams_recent_filament_preset(
+    const wxString& selected);
+
+} // namespace
+
 void AMSMaterialsSetting::on_select_ok(wxCommandEvent& event)
 {
     if (!obj) {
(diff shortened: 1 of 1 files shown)


XinZhangBambu:
Thanks for the contribution, @BenJule — a useful improvement🙏

One suggestion: I'd **drop scoping** (`ams_id` + `slot_id`). Using the same filament across different slots is very common.

Would you be open to reworking it that way? Thanks again!

BenJule:
Thanks for the feedback. Yes, that makes sense.

I’ll rework the implementation to use a shared recent-filament list instead of scoping it by `ams_id` and `slot_id`, so recently used presets are prioritized across all AMS slots.

I’ll update the PR accordingly. Thanks!

XinZhangBambu:
Looks Good To Me.

lanewei120:
thanks @BenJule

PR #11507, opened 2026-07-10
Show AMS filament compatibility hints

BenJule:
## Summary

This PR shows filament compatibility warnings directly in the AMS material settings dialog when a filament preset is selected.

## Why

Compatibility checks currently happen only after the user confirms the selection. This means users may complete the entire selection flow before learning that a filament is unsupported or requires special handling.

Showing the existing compatibility result immediately makes the selection process clearer while preserving the existing confirmation dialogs.

## Implementation

- Reuses the existing `sCheckFilamentInfo()` blacklist and compatibility check.
- Shows prohibition messages in red.
- Shows warning messages in orange.
- Hides the hint when no compatibility rule applies.
- Keeps the existing confirmation-time warnings and blocking behavior unchanged.
- Uses a separate label from the existing nozzle-temperature validation warning.
- Does not modify printer, cloud, or firmware behavior.

## Test plan

- Open the AMS material settings dialog for an editable third-party spool.
- Select a filament with a prohibition rule and confirm a red message appears.
- Select a filament with a warning rule and confirm an orange message appears.
- Select a filament without a matching compatibility rule and confirm no hint is shown.
- Clear the filament selection and confirm the hint disappears.
- Confirm existing warning and prohibition dialogs still appear after pressing Confirm.
- Confirm readonly Bambu/RFID slots do not show the new hint.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -174,6 +174,20 @@ void AMSMaterialsSetting::create_panel_normal(wxWindow* parent)
     m_sizer_filament->Add(m_readonly_filament, 1, wxALIGN_CENTER, 0);
     m_readonly_filament->Hide();
 
+    m_filament_compatibility_hint = new wxStaticText(
+        parent,
+        wxID_ANY,
+        wxEmptyString,
+        wxDefaultPosition,
+        wxDefaultSize,
+        0);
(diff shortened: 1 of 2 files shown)


XinZhangBambu:
Thanks @BenJule for the contribution — we appreciate it and will consider this change.

That said, the AMS material settings UI is currently being adjusted, so we are unable to merge this PR for now. We will revisit it once the new UI work is ready.

Thanks again!

PR #11508, opened 2026-07-10
Add user profile filter to AMS filament selection

BenJule:
## Summary

This PR adds a "Show only my profiles" filter to the AMS filament selection dialog.

## Why

Users with many installed system filament presets may need to scroll through a long list to find their own custom or duplicated profiles.

The new filter lets users temporarily display only their own compatible filament profiles.

## Implementation

- Marks collected filament entries as system or user presets.
- Keeps the complete sorted filament list in the dialog.
- Adds a "Show only my profiles" checkbox for editable AMS slots.
- Filters the existing dropdown locally without reloading printer or cloud data.
- Shows the checkbox only when user presets are supported and available.
- Clears an incompatible current selection when it is hidden by the filter.
- Leaves existing sorting and filament-selection behavior unchanged.

## Test plan

- Open an editable third-party AMS slot with system and user filament profiles.
- Confirm the checkbox is visible.
- Enable the checkbox and confirm only user-created profiles remain.
- Disable it and confirm the complete sorted list returns.
- Select a user profile and confirm the selection remains when filtering.
- Select a system profile, enable the filter, and confirm the hidden selection is cleared.
- Open a readonly Bambu/RFID slot and confirm the checkbox is hidden.
- Test a printer without user-preset support and confirm the checkbox is hidden.

### DIFF
--- a/src/slic3r/GUI/AMSMaterialsSetting.cpp
+++ b/src/slic3r/GUI/AMSMaterialsSetting.cpp
@@ -174,6 +174,19 @@ void AMSMaterialsSetting::create_panel_normal(wxWindow* parent)
     m_sizer_filament->Add(m_readonly_filament, 1, wxALIGN_CENTER, 0);
     m_readonly_filament->Hide();
 
+    m_show_only_user_presets = new wxCheckBox(
+        parent,
+        wxID_ANY,
+        _L("Show only my profiles"));
+    m_show_only_user_presets->SetFont(::Label::Body_13);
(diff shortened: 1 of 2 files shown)


hemai249:
@BenJule   Thank you very much for your contribution. We have designed a new consumable selection UI that makes selecting consumables very convenient, so this feature is no longer needed.

BenJule:
Understood, thank you for the update and for taking the time to review the contribution. I’ll close this PR accordingly.

PR #11509, opened 2026-07-10
Add Standard 3MF color import preference

### LINKED ISSUE #9734: Option to disable import color dialog
The new dialog that comes up when importing a mesh is annoying. I don't ever use it, so it just adds another step in my import process, which I do frequently.

Change it to an option that you can disable via preferences. I would ideally like to be able to set the default for how items are imported rather than have to first click "import geometry only" and now additionally have to close the window for the "standard 3mf import color" dialog. It's two steps I wish I could avoid. Sometimes I import multiple items from Fusion 360 and have to dozens of times just to ignore these dialogs, which is time consuming and provides zero benefit.

There's no alternative, I have to click two windows for every single item I import. If it is a multi part component, I have to click yet another dialog for every single import. Super annoying.

**Additional context**
None.

BenJule:
Closes #9734

## Summary

This PR adds a preference that controls how color information from standard 3MF files is handled during import.

## Available modes

- **Ask every time**: preserves the existing behavior and opens the color mapping dialog.
- **Apply automatic color mapping**: applies the dialog's existing default mapping without requiring confirmation.
- **Import geometry only**: skips color processing and imports only the model geometry.

## Why

Users who frequently import standard 3MF files currently need to dismiss or confirm the color mapping dialog for every import, even when they always choose the same action.

The new preference removes those repetitive steps while preserving the existing interactive workflow as the default.

## Implementation

- Adds `standard_3mf_color_import_mode` to the application configuration.
- Defaults the setting to `ask`, preserving existing behavior.
- Adds a new option under Preferences → 3D Settings.
- Reuses the existing default mapping logic for automatic imports.
- Falls back to the interactive dialog if automatic mapping is incomplete.
- Avoids constructing the color dialog in geometry-only mode so the model is not modified by the dialog initialization.
- Leaves OBJ color import behavior unchanged.

### Ask every time

- Set the preference to **Ask every time**.
- Import a standard 3MF file containing color information.
- Confirm the existing color mapping dialog appears.
- Confirm both OK and Cancel retain their previous behavior.

### Automatic mapping

- Set the preference to **Apply automatic color mapping**.
- Import a standard 3MF with a valid automatic mapping.
- Confirm the model imports without showing the dialog.
- Confirm mapped colors and newly required filament colors are applied.
- Import a file with an incomplete mapping and confirm the interactive dialog appears as a fallback.

### Geometry only

- Set the preference to **Import geometry only**.
- Import a colored standard 3MF.
- Confirm no color mapping dialog appears.
- Confirm the geometry imports without painted segmentation data.

### Regression checks

- Import an OBJ file with color information and confirm its dialog behavior is unchanged.
- Import a Bambu Studio project 3MF and confirm the new option does not affect project loading.
- Restart Bambu Studio and confirm the selected preference persists.

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -111,6 +111,9 @@ void AppConfig::set_defaults()
         if (get("drop_project_action").empty())
             set_bool("drop_project_action", true);
 
+        if (get("standard_3mf_color_import_mode").empty())
+            set("standard_3mf_color_import_mode", "ask");
+
 #ifdef _WIN32
         if (get("associate_3mf").empty())
             set_bool("associate_3mf", false);
--- a/src/slic3r/GUI/ObjColorDialog.cpp
+++ b/src/slic3r/GUI/ObjColorDialog.cpp
@@ -168,6 +168,16 @@ bool ObjColorDialog::Show(bool show) {
     }
 };
 
+bool ObjColorDialog::apply_default_mapping()
+{
+    if (!m_panel_ObjColor || !m_panel_ObjColor->is_ok())
+        return false;
+
+    m_panel_ObjColor->clear_instance_and_revert_offset();
+    m_panel_ObjColor->send_new_filament_to_ui();
+    return true;
+}
+
 ObjColorDialog::ObjColorDialog(wxWindow *parent, Slic3r::ObjDialogInOut &in_out, const std::vector<std::string> &extruder_colours)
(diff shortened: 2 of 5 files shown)


Haidiye00 (commented):

BenJule:
Sure, here's the new preference in action:

demo

Haidiye00:
Hi, the code for the latest preferences window has been updated. Please resubmit the code and submit the gif. Thanks.

Issue #496, opened 2022-10-22
[Feature Request] Set Number of Instances/Fill bed with instances


harriettubgirl:
Hello I just wanted to put in my request for re-adding the ability to add instances or set number of instances of a specific model from prusaslicer/superslicer as it's implemented over there. 

I find it's far easier to work with than using clone.

Thanks!

YifanWuBambu:
@harriettubgirl 
Bambu Studio optimized the clone logic.
All cloned objects shared the same geometry data and only one object is really sliced.

rncfiles:
I agree instances is important functionality.  The key benefit of instances is that they inherit the object print settings.  Clones, once created must be managed as separate objects.  What is interesting is that BambuStudio appears to currently support instance functionality.  If a 3mf file that was created in PrusaSlicer has model instances, they will carry over when imported to BambuStudio.  They show as "instance" and behave as expected.  It just seems like in BambuStudio, the + and - options were just omitted.  

Please add this functionality.

jeffamelang:
Cloning and instances are fundamentally different. I don't think that the end user cares that "only one object is really sliced", we would like to have a primary copy of the geometry and a bunch of instances, so that if you make changes to the primary copy, that change affects all of the instances. 

Please add this functionality.

harriettubgirl:
Thank you for adding your feedback on this! I agree, it has a lot of important functionality and I really hope it's added!

ftheriachab:
Definitely would like this feature implemented.  When batching out parts, it's so nice to have the slicer determine the maximum number of objects in can print.

kozakm00:
Agree with OP, need to add instancing to the Bambu studio. Prusa slicer has it

pennystockss:
This is huge for me. Printing 3-20 versions of the same object routinely. Kind of a pain having to delete every one but a single one. Make a change, then clone it back.

khibchen:
Linked instances that all share the same parameters and modifiers is a critical feature for anyone that produces multiple identical parts on a build plate. The way it’s implemented in Prusa Slicer is very powerful and simple to use. The clone function in Bamboo Studio basically just makes ‘dumb’ copies of a part. Any changes to one part are not automatically reflected in the other parts.

[...]

PR #11514, opened 2026-07-10
feat: add fill bed with object instances

BenJule:
## Summary

Adds a separate **Fill bed with instances** action alongside the existing **Fill bed with copies** action.

The existing action continues to create independent object copies. The new action fills the remaining plate area with linked `ModelInstance` entries belonging to the selected object.

## Changes

- Add separate menu entries for:
  - Fill bed with copies
  - Fill bed with instances
- Add a dedicated persistent `FillBedJob` for instance creation
- Reuse the selected instance as the transformation template
- Add generated instances to the existing `ModelObject`
- Update the object tree correctly when transitioning from one instance to multiple instances
- Preserve the existing copy-based behavior unchanged

## Implementation notes

Bambu Studio already contained the `fill_bed_with_instances()` entry point, but it invoked the same copy-based fill-bed job.

This change gives the fill-bed job an explicit copies/instances mode and registers separate persistent jobs for both operations.

## Testing

- Targeted compilation of `FillBedJob.cpp`
- Targeted compilation of `GUI_Factories.cpp`
- Targeted compilation of `Plater.cpp`
- `git diff --check`

Not tested:

- Full application build
- Runtime UI behavior

Related to #496

### DIFF
--- a/src/slic3r/GUI/GUI_Factories.cpp
+++ b/src/slic3r/GUI/GUI_Factories.cpp
@@ -2461,8 +2461,26 @@ void MenuFactory::append_menu_item_locked(wxMenu* menu)
 void MenuFactory::append_menu_item_fill_bed(wxMenu *menu)
 {
     append_menu_item(
-        menu, wxID_ANY, _L("Fill bed with copies"), _L("Fill the remaining area of bed with copies of the selected object"),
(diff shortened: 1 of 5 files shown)


Haidiye00 (commented):
Hi, please submit the recorded effect of the gif. Also, BanbuStudio might not support multiple instances, so consider extra slicing.

PR #11516, opened 2026-07-11
feat: link identical object copies as instances

BenJule:
## Summary

Add a context-menu action that converts selected identical object copies into linked instances of a single object.

This provides the inverse workflow of **Set as individual objects** and is useful when independent copies created through cloning or copy-based bed filling should share the same object geometry and settings again.

## Changes

- Add **Link selected copies as instances** to the multiple-object context menu.
- Keep the selected object with the lowest model index as the master object.
- Copy the instance transformation of every other selected object to the master.
- Remove the converted source objects in descending index order.
- Rebuild the instance hierarchy in the object list.
- Preserve plate assignments and printable states.
- Refresh the 3D scene, assembly view, PartPlate data, slicing state, and sidebar controls.
- Store the complete conversion as a single undo operation.

## Compatibility checks

The action is only available when all selected entries are complete object nodes and can be converted without losing object-specific information.

Each selected object must:

- contain exactly one instance
- use the same number and order of volumes
- share the same underlying mesh allocations
- use identical object and volume configurations
- use identical volume types, materials, and transformations
- use identical layer ranges and layer-height profiles
- use identical support, seam, fuzzy-skin, MMU, and exterior facet annotations
- use compatible assembly metadata

The action is disabled for:

- SLA objects
- cut objects and cut connectors
- editable text or SVG volumes
- objects with individual brim points
- objects with SLA support points or drain holes
- separately imported meshes that merely have similar geometry

Mesh pointer equality is intentional. It identifies actual Bambu Studio object copies and prevents unrelated imported objects from being linked accidentally.

## Transformation preservation

The conversion preserves each source instance's:

- position
- rotation
- scaling
- mirroring
- printable state
- plate placement

The operation does not call `ensure_on_bed()`, preventing the converted instances from being shifted vertically.

## Validation

- `git diff --check`
- targeted compilation of:
  - `GUI_ObjectList.cpp.o`
  - `GUI_Factories.cpp.o`
- repeated Ninja build returned `no work to do`
- no compilation errors were reported

A full application build and interactive UI test have not been performed.

### DIFF
--- a/src/slic3r/GUI/GUI_Factories.cpp
+++ b/src/slic3r/GUI/GUI_Factories.cpp
@@ -1144,6 +1144,24 @@ void MenuFactory::append_menu_items_convert_unit(wxMenu* menu)
     }
 }
 
+void MenuFactory::append_menu_item_link_copies_as_instances(wxMenu *menu)
+{
+    append_menu_item(
+        menu,
+        wxID_ANY,
+        _L("Link selected copies as instances"),
+        _L("Convert the selected identical object copies into linked instances"),
+        [](wxCommandEvent &) {
+            obj_list()->link_selected_copies_as_instances();
+        },
+        "",
+        menu,
+        []() {
+            return obj_list()->can_link_selected_copies_as_instances();
+        },
+        m_parent);
+}
+
 void MenuFactory::append_menu_item_merge_to_multipart_object(wxMenu* menu)
 {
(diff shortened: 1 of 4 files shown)


Haidiye00 (commented):

BenJule:
Here you go:

demo

Haidiye00 (commented):
Hello, we are very sorry, but bambuStudio is currently designed to not support multiple instances. Multiple instances of objects may cause new problems.

Issue #37, opened 2022-07-21
Feature Request: Purge to STL


kuljaden:
Hello everyone, I am writing this a feature request for slicer. It would neat to see a feature where we could purge to an STL file, a random or selected file, to purge all color changes through. For example, a mutli color model and instead of using the purge in the back, then wiping to a purge tower, simple use a model on the side to purge into and maybe have it auto size the model as it would the purge tower.

Thanks!

PR #11533, opened 2026-07-12
fix: show Bambu splash screen on Linux instead of black window

### LINKED ISSUE #11532: Linux: black window shown at startup instead of the Bambu splash screen
On Linux, Bambu Studio shows a plain black window during startup instead of the splash screen. The splash only becomes visible (if at all) for a moment right before the main window appears. This affects the official AppImage as well as distro builds, on X11 and Wayland alike (reproduced on Debian 13 with KDE Plasma, both 2.7.x and 2.8.0 beta).

The reason is in `GUI_App::on_init_inner()` in `src/slic3r/GUI/GUI_App.cpp`:

```cpp
scrn = new BBLSplashScreen(bmp, wxSPLASH_CENTRE_ON_SCREEN, 0, splashscreen_pos);
#ifndef __linux__
    wxYield();
#endif
```

wxYield() is deliberately skipped on Linux, but on wxGTK it is exactly what flushes the pending paint events. Without it GTK never gets a chance to draw the splash before the long blocking startup work (preset loading, network agent init, ...) begins, so the window stays black the whole time.

How it looks today:

black window instead of splash

And with wxYield() enabled on Linux plus a Refresh()/Update() after the first SetText() call:

splash screen rendering correctly

I've been running this fix in my fork since May without any regressions on Windows or Linux. PR follows.

BenJule:
Fixes #11532

On Linux the startup splash is currently a plain black window, because wxYield() is explicitly skipped on this platform right after the BBLSplashScreen is created. On wxGTK that call is what flushes the pending paint events — without it GTK never draws the splash before the long blocking startup work begins.

This change enables wxYield() on Linux too (it is harmless on the other platforms, which already ran it), raises the window and adds a Refresh()/Update() pair after the first SetText() so the status line actually becomes visible.

Before and after on Debian 13 / KDE Plasma:

before: black window

after: splash renders

I have been running this in my fork since May 2026 (Debian and Ubuntu builds, X11 and Wayland) with no regressions on Windows either.

### DIFF
--- a/src/slic3r/GUI/GUI_App.cpp
+++ b/src/slic3r/GUI/GUI_App.cpp
@@ -3286,10 +3286,18 @@ bool GUI_App::on_init_inner()
         BOOST_LOG_TRIVIAL(info) << "begin to show the splash screen...";
         //BBS use BBL splashScreen
         scrn = new BBLSplashScreen(bmp, wxSPLASH_CENTRE_ON_SCREEN, 0, splashscreen_pos);
-#ifndef __linux__
+        // Process pending paint events so the splash is drawn immediately on all
+        // platforms. Without this, GTK never paints the window before the heavy
+        // loading work begins, leaving a black window until the app is ready.
         wxYield();
-#endif
+        scrn->Raise();
+        scrn->Update();
         scrn->SetText(_L("Loading configuration")+ dots);
+        // BBLSplashScreen::SetText() does not force a repaint on non-macOS.
+        // Refresh() + Update() ensure the first status line is visible before
+        // the heavy startup work begins.
+        scrn->Refresh();
+        scrn->Update();
     }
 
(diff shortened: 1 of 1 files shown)


BenJule:
Additional verification after rebasing onto current master:

- Jenkins Linux PR build #37: **SUCCESS**
- Built commit: `c3d8244db46a3cdf58c1f083f5bf104c7b583e05`
- Debian Trixie build artifact was successfully generated and verified
- I downloaded and launched the exact Jenkins #37 artifact locally
- Manual startup test: **PASS**
- The Bambu splash screen is rendered immediately again instead of showing the black startup window

So the fix is now both compile-tested and manually verified with the CI-built artifact.

tonghao-bbl:
@lanewei120 LGTM

PR #11577, opened 2026-07-16
Fix macOS 26+ dock icon losing system theming (#8816)

### LINKED ISSUE #8816: macOS Tahoe icon correction
When Bambu Studio is launched on macOS, it changes the icon in the dock to pull from "/Applications/BambuStudio.app/Contents/Resources/images/gcode.icns", which is a scaled up version, this overrides Tahoe's Liquid Glass effects (also dark mode/color tint) and the icon is sized wrong so it becomes bigger than all other app icons. 

Can you please update the app to either not pull gcode.icns, or have an appropriately sized icon? Thanks. 

(Closed, Dark mode applied by Tahoe)
Image

(Open, icon overrides from gcode.icns)
Image

d3ns0n:
### Problem

On macOS 26 (Tahoe) the Bambu Studio dock icon does not respect the system's
new icon theming (Liquid Glass / Tinted / Clear / Dark). When the app launches
it loses the themed appearance and shows a plain raster icon instead.

Root cause: on startup `MainFrame` sets a custom dock icon at runtime via a
`wxTBI_DOCK` taskbar icon (`BambuStudio-mac_256px.ico`). This calls
`-[NSApp setApplicationIconImage:]` under the hood, which overrides the system
theming that macOS applies to the app bundle icon (`Icon.icns`).

Fixes #8816

### Change

Skip the runtime dock-icon override on macOS 26 and later, so the system-themed
bundle icon is used instead. Behaviour on macOS < 26 is unchanged.

- Add `Slic3r::is_mac_os_at_least(int major)` in `MacUtils.mm`, implemented with
  `-[NSProcessInfo isOperatingSystemAtLeastVersion:]`. This is a lower-bound
  check, so future releases (27+) are covered automatically, and it works
  regardless of the SDK the project is built with.
- Guard the `#ifdef __APPLE__` dock-icon initialization in `MainFrame` with
  `if (!is_mac_os_at_least(26))`.

### Testing

- Built and ran locally on macOS 26.5 (Apple Silicon, arm64).
- With the fix, the dock icon now follows the system Icon & Widget Style
  (verified with the "Tinted" style — the running app's dock icon stays tinted
  instead of reverting to the raster icon).
- Change is fully contained in `#ifdef __APPLE__`, so non-Apple builds are
  unaffected.

### DIFF
--- a/src/libslic3r/MacUtils.hpp
+++ b/src/libslic3r/MacUtils.hpp
@@ -5,6 +5,9 @@ namespace Slic3r {
 
 bool is_macos_support_boost_add_file_log();
 int  is_mac_version_15();
+
+// Returns true when running on macOS `major` or any later version.
+bool is_mac_os_at_least(int major);
 }
 
 #endif
--- a/src/libslic3r/MacUtils.mm
+++ b/src/libslic3r/MacUtils.mm
@@ -20,4 +20,12 @@ int is_mac_version_15()
         return false;
     }
 }
+
+// Runtime check that works regardless of the SDK used to build, and is
+// inherently true for the given major version and all later ones.
+bool is_mac_os_at_least(int major)
+{
+    NSOperatingSystemVersion version = { major, 0, 0 };
+    return [[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:version];
+}
 }; // namespace Slic3r
--- a/src/slic3r/GUI/MainFrame.cpp
+++ b/src/slic3r/GUI/MainFrame.cpp
@@ -47,6 +47,9 @@
 #include "Widgets/ProgressDialog.hpp"
 #include "BindDialog.hpp"
 #include "../Utils/MacDarkMode.hpp"
+#ifdef __APPLE__
+#include "libslic3r/MacUtils.hpp"
+#endif
 
 #include <fstream>
 #include <string_view>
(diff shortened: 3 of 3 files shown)


d3ns0n:
### Screenshots (app running on macOS 26.5, Apple Silicon)

**Before → After** — Icon &amp; Widget Style: *Tinted – Light*

| Before (current behaviour) | After (this PR) |
| :---: | :---: |
| before | after-tinted-light |

With the fix the dock icon follows the system Icon &amp; Widget Style instead of the raster override:

| Dark | Tinted – Dark | Tinted – Light |
| :---: | :---: | :---: |
| after-dark | after-tinted-dark | after-tinted-light |

XinZhangBambu:
Thanks a lot for your suggestion and for taking the time to raise this.

We’ve discussed it with our interaction design team, and we’re aligned that this change is not necessary at the moment, so we won’t move forward with it in this PR.
Really appreciate the input — we’ll keep it in mind for future iterations.

d3ns0n:
Thank you for your response. I am not quite sure why this change is not considered necessary, and I would appreciate it if you could explain how you came to this conclusion. Are any other steps planned to address the icon color mismatch on macOS?

Looking forward to hearing from you.

XinZhangBambu:
**from interaction design team:**

@d3ns0n Thank you for following up and for giving us the opportunity to clarify.

Our decision is based on brand consistency rather than technical feasibility. Bambu Lab’s visual identity, including the approved colors and shape of its brand symbols, is a core brand asset and should not be altered across different platforms or system themes.

We really appreciate you bringing this to our attention and helping us improve the product experience.

d3ns0n:
@XinZhangBambu Thank you for the answer. Then I'd suggest applying the brand colors to the closed-state icon as well, so both states stay consistent. Would you be interested in a PR for that?

PR #11599, opened 2026-07-19
Add Windows ARM64 and ARM64EC build support

RyanEwen:
## What this does

Adds native Windows on ARM support to BambuStudio in two variants:

- **arm64**: a fully native ARM64 build. Everything compiles and links for
  ARM64, including OCCT/OpenCASCADE, so STEP import, SVG to 3D and text emboss
  all work. The one limitation is the network plugin (see below).
- **arm64ec**: native ARM64 code built with the x64-compatible ABI. An ARM64EC
  process runs the application at native speed and can still load x64-only DLLs
  in process, which is what allows the closed-source Bambu network plugin
  (cloud login, device management, camera) to work. Bambu currently ships that
  plugin as x64 only, so this is the variant most users on ARM would want.

Both are folded into the existing Windows build matrix on the `windows-11-arm`
runner. All ARM-specific logic is gated, so the x64, x86, macOS and Linux
paths are unchanged.

## Why

Windows on ARM machines (Snapdragon X and similar) currently run BambuStudio
under x64 emulation. A native build removes the emulation overhead from
slicing, and the ARM64EC variant keeps full networking/device support while
doing so.

## Notes for reviewers

- The `windows-11-arm` GitHub runner ships Visual Studio 2022, not 2026, so the
  ARM jobs use the `Visual Studio 17 2022` generator and resolve the Windows
  SDK at runtime rather than pinning it.
- GMP and MPFR have no upstream win-arm64 prebuilt binaries; the ARM64 blobs
  were already present in the tree. MPFR 4.x ships as `libmpfr-6.lib`, so an
  aliased `libmpfr-4.lib` is installed for CGAL's `FindMPFR`.
- FFMPEG uses BtbN's winarm64 shared build for the ARM targets because the
  bundled prebuilt package is x64 only. The tag and hash are pinned.
- For ARM64EC, oneTBB is patched at configure time (drop the hard-coded `/GL`
  that blocks EC links, and stub the Intel TSX intrinsics, which are unreachable
  under emulation but must resolve at link). The patch is a no-op on non-MSVC
  toolchains.
- OpenSSL for ARM64EC is built as its x64 flavor under the x64 cross
  environment, since ARM64EC links the x64 static libraries.

## Testing

- CI builds, links and uploads native arm64 and arm64ec portable artifacts on
  `windows-11-arm`; the x64, macOS and Linux jobs remain green.
- The arm64ec build was validated on a Snapdragon X machine: it launches,
  slices, imports STEP, and loads the x64 network plugin in process (verified
  by module list and by signing in to a Bambu account and driving a printer).

## Not included

Signed installer and release-pipeline wiring for the ARM artifacts are left as
a maintainer follow-up. This is a build-and-CI change.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -47,6 +47,10 @@ jobs:
           - os: ubuntu-24.04
           - os: ubuntu-26.04
           - os: windows-latest
+          - os: windows-11-arm
+            arch: arm64
+          - os: windows-11-arm
+            arch: arm64ec
           - os: macos-15-intel
             arch: x86_64
           - os: macos-15
--- a/.github/workflows/build_bambu.yml
+++ b/.github/workflows/build_bambu.yml
@@ -37,7 +37,7 @@ jobs:
           fail-on-cache-miss: true
 
       - name: Get the version and date on Ubuntu and macOS
-        if: inputs.os != 'windows-latest'
+        if: ${{ !startsWith(inputs.os, 'windows') }}
         run: |
           ver_pure=$(grep 'set(SLIC3R_VERSION' version.inc | cut -d '"' -f2)
           if [[ "${{ github.event_name }}" == "pull_request" ]]; then
@@ -51,7 +51,7 @@ jobs:
         shell: bash
 
       - name: Get the version and date on Windows
-        if: inputs.os == 'windows-latest'
+        if: startsWith(inputs.os, 'windows')
         run: |
           $date = Get-Date -Format 'yyyyMMdd'
           $ref = "${{ github.ref }}"
@@ -114,11 +114,11 @@ jobs:
 
   # Windows
       - name: setup MSVC
-        if: inputs.os == 'windows-latest'
+        if: startsWith(inputs.os, 'windows')
         uses: microsoft/setup-msbuild@v2
 
       - name: Install nsis and pkgconfig
-        if: inputs.os == 'windows-latest'
+        if: startsWith(inputs.os, 'windows')
         run: |
           dir "C:/Program Files (x86)/Windows Kits/10/Include"
           choco install nsis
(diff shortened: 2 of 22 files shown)


RyanEwen:
@MackBambu any estimate as to when you may get to look into this? Orca already supports this - hoping to get it into Bambu Studio asap.

lexcyn:
This is awesome. Let's get this added ASAP @MackBambu as those of us with ARM hardware are struggling with performance issues under emulation.

blauver:
Is there any ability to get this reviewed and committed for us ARM64 users?

RyanEwen:
@blauver and @lexcyn I got tired of waiting for this and decided to create a web-based UI for Bambu Studio as part of [PrintStream](https://printstream.app) - yes this is a shameless plug :). 

Additionally, and this isn't advertised much yet, there's a public (no sign-up/login) version of the editor/slicer feature by itself, at https://printstream.app/3mf-editor - but YMMV as it's not thoroughly tested compared to the editor built into the full PrintStream application. 

Editing is done in-browser. The actual slicing happens via Bambu Studio CLI on our x86 server (no way around that, sorry). We don't look at or hold onto anything, FWIW (zero interest in that, at all).

If you have an x86 machine lying around that you can use as a server, you could install / self-host PrintStream on it for free, and use it from your ARM64 machine. The Community Edition is free. Files don't need to go through our server in that case.

generally-ready:
+1 on @RyanEwen 's post above.  It's a much better solution given BL's unresponsiveness.

FWIW, I've had a support ticket open with BL for over a year now.  My latest request in it was for them to review this commit and give me any indication of when they might integrate it.  Here's the response (September 18, 2026):

_> Currently, there are no plans to release a Windows native ARM64 version of Bambu Studio. This request is still pending and under evaluation. We are unable to provide you with a specific timeline for when this feature will be launched. Regrettably, due to current scheduling constraints, we are unable to address this request in the short term. We sincerely apologize for the inconvenience this may cause._  

So, I'm afraid that's that.

RyanEwen:
I got a similar response regarding Bambu Connect and the Bambu Farm client/server applications, which simply don't launch at all on ARM last I checked.. Also, Bambu's farm software seemed doomed from the beginning simply by them deciding to make it a Windows-native client/server combo.

PR #11602, opened 2026-07-19
FIX: compile error on debug

tonghao-bbl:
fix compile error on Mac in debug mode


Haidiye00:
Hello, I suggest combining it into a patch. Thank you

Issue #11630, opened 2026-07-21
Support interface is not generated if "On build plate only" option is selected


NilsRo:
### Type of issue

Other (functional bug, wrong result, UI glitch, etc.)

### Bambu Studio Version

2.7.1.62

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11 x64

### Printer

Bambu Lab X1C

### How to reproduce

1. Open example 3MF
2. Slice with "On build plate only" option on and off

### Actual results

No interface layer generated

<img width="1829" height="995" alt="Image" src="https://github.com/user-attachments/assets/653e2c14-4a7b-4710-9efe-e27c17c7cb94" />

Really bad surface quality!

<img width="768" height="1024" alt="Image" src="https://github.com/user-attachments/assets/3004b2be-f5cb-41fb-9776-3315b39d05f7" />

### Expected results

Should generate an interface layer like without the option.

<img width="1862" height="1244" alt="Image" src="https://github.com/user-attachments/assets/5af946f6-cb31-439e-9cdc-00a222c5dcd8" />

### File uploads (project / log / crash dump / performance profile)

3MF with screenshots
[Support Interface Bug.zip](https://github.com/user-attachments/files/30220008/Support.Interface.Bug.zip)

### Checklist of files to include

- [x] Project file (always)
- [ ] Log file (always)
- [ ] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)
- [ ] Performance profile — required for performance / lag issues

ArthurBambulab:
It's not a bug. When "On build plate only" option is enabled, the supports has to grow around the object so there is no collision. In this case, the support nodes are all tree nodes, and the interfaces are all small circles.
On the other hand, I agree it's not the best solution. I'll keep this issue open until we find a better solution.

NilsRo:
> On the other hand, I agree it's not the best solution.

Yes. :-) I do not see any advantage that there is no interface layer on on top of the small circles. It perhaps reduces the space for the tree node but there should be a split-up in support that cannot have an interface layer and with enough space for an interface layer. Actually the print result with PETG is very very bad.

BenJule:
Thanks for the clarification. I understand that the current behaviour is intentional and therefore not classified as a bug.

[...]

Issue #27, opened 2022-07-18
[Bug Ux / 01.00.10.05 / Win ] conversion to meters/inches out of bed


MarcoPiazza83:
When converting a model to meters or inches, the model exits the print plane without a warning as in prusaSlicer.

<img width="284" alt="Schermata 2022-07-18 alle 11 03 37" src="https://user-images.githubusercontent.com/4783556/179479156-588d0fd6-d70f-49ec-870b-10996954c571.png">

lanewei120:
already fixed in the newest released version v1.2.0

PR #11635, opened 2026-07-21
Improve tree support interfaces in build-plate-only mode

BenJule:
## Summary

- improve tree support interface coverage when **On build plate only** is enabled
- replace the previous branch-distance dilation approach with a cumulative support-backed footprint
- allow recovered interface only where it is reachable from geometry that is actually printable on the support layer below
- keep the normal non-build-plate-only path unchanged

## Problem

Tree support normally uses the overhang area when generating the support interface. In **On build plate only** mode, branches that cannot reach the build plate are pruned first.

Reusing the complete overhang afterward is unsafe because it can recreate interface paths above branches that no longer exist. Restricting the interface to the individual surviving tree-node footprints, however, leaves considerably less interface coverage than necessary.

## Approach

For build-plate-only tree support, this change builds a cumulative supported footprint across the support layers.

For each layer:

1. Start with the geometry actually printed by the surviving tree branches on the current layer.
2. Take the geometry that was printable on the preceding support layer and allow it to grow by at most half of one support extrusion width.
3. Intersect that reachable area with the interface requested for the current layer.
4. Add only that recovered interface to the current printable footprint.
5. Apply the existing object-collision and machine-border clipping.
6. Only the resulting printable footprint is allowed to propagate to the next layer.

This allows neighbouring surviving tree branches to gradually form a more continuous interface while preventing interface regions above pruned support from being restored.

The previous `tree_support_branch_distance` dilation is no longer used by this fix.

When **On build plate only** is disabled, the existing full-overhang interface path remains unchanged.

## Validation

Tested commit:

`cd09fd64a4ac2619696759660caac2ea06d28f5b`

Source validation:

- `git diff --check` passes
- targeted `TreeSupport.cpp` `-fsyntax-only` compile check passes
- PR contains one signed commit and only changes `src/libslic3r/Support/TreeSupport.cpp`

CI:

- Jenkins Linux PR build #27: **SUCCESS**
- `BUILD_APP=true`
- Debian Trixie
- exact artifact: `BambuStudio-pr-11635-cd09fd64a4ac-debian-trixie.tar.gz`

Runtime validation with the supplied 3MF using Bambu Lab X1 Carbon, Tree Strong and the same slicing settings:

### On build plate only = ON

- Support: **5.78 m / 17.80 g**
- Support interface: **0.20 m / 0.57 g**
- Support transition: **0.12 m / 0.37 g**

For comparison, earlier conservative candidates produced only 0.10 m and 0.17 m of support interface.

Representative single-layer inspection around layers 325, 326 and 327 was also performed. In the inspected regions, the recovered interface remained aligned with generated support on the layer below and no large unsupported interface islands like those produced by the original dilation approach were observed.

### On build plate only = OFF

Control slice:

- Support: **4.24 m / 13.06 g**
- Support interface: **0.37 m / 1.06 g**
- Support transition: **0.37 m / 1.13 g**

These values match the previous non-build-plate-only reference, confirming that the normal path remains unchanged.

Addresses #11630

### DIFF
--- a/src/libslic3r/Support/TreeSupport.cpp
+++ b/src/libslic3r/Support/TreeSupport.cpp
@@ -2333,12 +2333,269 @@ void TreeSupport::draw_circles()
     const bool with_lightning_infill = m_support_params.base_fill_pattern == ipLightning;
     coordf_t support_extrusion_width = m_support_params.support_extrusion_width;
     const coordf_t line_width_scaled                 = scale_(support_extrusion_width);
+    const coord_t buildplate_roof_growth =
+        std::max<coord_t>(1, coord_t(line_width_scaled / 2));
     const float tree_brim_width = config.raft_first_layer_expansion.value;
     const bool bottom_expand_enabled = config.tree_support_wall_count > 1 || config.tree_support_wall_count < 0;
 
     if (m_object->support_layer_count() <= m_raft_layers)
         return;
     BOOST_LOG_TRIVIAL(info) << "draw_circles for object: " << m_object->model_object()->name;
+
+    // Build the same collision region used while drawing a node, but for an
+    // arbitrary node in the surviving tree.
+    auto node_collision = [&](const SupportNode &node) -> ExPolygons {
+        const size_t obj_layer_nr = node.obj_layer_nr;
+        const bool sharp_tail = node.is_sharp_tail && node.distance_to_top <= 0;
+
+        ExPolygons collision = offset_ex(
+            m_ts_data->m_layer_outlines[obj_layer_nr],
+            sharp_tail ?
+                scale_(top_z_distance) :
+                scale_((obj_layer_nr == 0) ?
+                    config.support_object_first_layer_gap :
+                    m_ts_data->m_xy_distance));
+
+        if (top_z_distance > EPSILON) {
+            float accum_height = 0;
+            for (size_t layer_id = obj_layer_nr + 1;
+                 layer_id < m_ts_data->m_layer_outlines.size() &&
+                 (accum_height += m_object->get_layer(layer_id)->height) &&
+                 accum_height <= top_z_distance;
+                 ++layer_id) {
+                collision = union_ex(
+                    collision,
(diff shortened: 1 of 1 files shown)


NothFie:
@BenJule Thank you for your support. However, as shown in the picture, simply expanding the area to get the interface will create a lot of overhanging interface lines. These lines may warp or peel off, which could cause the print to fail.
image

BenJule:
Thanks for pointing this out. You're right: expanding the surviving branch area can create interface paths that extend beyond the actually supported footprint, so those lines may warp or detach.

I'll drop the dilation-based approach rather than just tuning the expansion distance. A proper fix needs to guarantee that every added interface region is backed by generated support on the layers below. I'll rework the approach and validate the resulting toolpaths with the supplied 3MF before updating the PR.

BenJule:
@NothFie Thanks again for pointing out the problem with the original approach.

I have now replaced the dilation-based implementation completely. The updated version no longer expands the interface using `tree_support_branch_distance`.

In **On build plate only** mode, it now builds a cumulative layer-wide footprint from geometry that is actually printable. Additional requested interface on a layer is accepted only where it can be reached from the printable geometry on the preceding support layer, with growth limited to half of one support extrusion width. Only the resulting printable geometry can propagate further to the next layer.

I validated the updated implementation with the supplied 3MF:

- **On build plate only = ON:** support interface 0.20 m / 0.57 g, transition 0.12 m / 0.37 g
- **On build plate only = OFF:** support interface 0.37 m / 1.06 g, transition 0.37 m / 1.13 g
- representative single-layer checks around layers 325-327 showed the recovered interface remaining backed by generated support in the inspected regions
- the OFF control matches the previous normal-path result
- Jenkins Linux PR build 27 completed successfully for `cd09fd64a4ac2619696759660caac2ea06d28f5b`

So the goal is no longer to restore the complete overhang in build-plate-only mode. The recovered interface is deliberately limited by the support that can actually be carried through the surviving tree.

Could you please take another look at the updated approach?

PR #11641, opened 2026-07-21
Keep tool ordering layer height lookup in bounds

BenJule:
## Summary
- use the configuration fallback accessor when matching maximum layer heights to nozzles
- prevent an out-of-bounds read when `max_layer_height` is shorter than `nozzle_diameter`
- add a focused regression test for two nozzles with one maximum-layer-height entry

## Testing
- syntax-checked `ToolOrdering.cpp` with the existing Release compiler flags
- syntax-checked the isolated `test_tool_ordering.cpp` regression test
- verified the patch with `git diff --check`


XunZhangBambu (changes requested):
thanks for your refinement. But i think the test cases is not neccesarry

PR #11644, opened 2026-07-21
Avoid travel crash before initial tool selection

BenJule:
## Summary

- keep a safe process configuration index for travel moves before the first filament is selected
- refresh the cached index when an extruder is initialized or changed
- use the cached configuration for XY, XYZ, Z, and spiral-Z travel speeds
- add a regression test covering XY and Z travel without an active filament

## Problem

Some multi-extruder configurations may generate a travel move before initial tool selection. Travel speed lookup currently dereferences the active filament even though no filament has been selected yet, causing a null-pointer crash.

Until the first tool is selected, travel moves now use the base process configuration at index 0. Normal per-extruder configuration lookup resumes as soon as a filament becomes active.

## Testing

- `GCodeWriter.cpp` syntax-checked using the existing generated compiler command
- isolated regression test syntax-checked with the same compiler configuration
- verified that all affected travel-speed lookups use the cached process configuration index

### DIFF
--- a/src/libslic3r/GCodeWriter.cpp
+++ b/src/libslic3r/GCodeWriter.cpp
@@ -33,6 +33,7 @@ void GCodeWriter::set_extruders(std::vector<unsigned int> extruder_ids)
 {
     std::sort(extruder_ids.begin(), extruder_ids.end());
     m_curr_extruder_id = -1;
+    m_current_process_config_idx = 0;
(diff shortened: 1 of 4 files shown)


zhimin-zeng-bambulab (approved):

lanewei120:
thanks @BenJule

PR #11651, opened 2026-07-21
Avoid measure edge selection crash

BenJule:
## Summary

- position the second measurement gripper from the second selected feature
- use the geometric midpoint when a plain edge has no additional center point
- avoid dereferencing an empty optional during measurement rendering

The second gripper previously used the first selected feature. If that feature was a plain edge without an additional center point, updating the second gripper could access an empty optional and crash.

## Testing

- syntax-checked `GLGizmoMeasure.cpp` using the existing Release GUI build flags
- verified that the second gripper no longer references the first selected feature


Haidiye00 (changes requested):

PR #11652, opened 2026-07-21
Avoid crash when transforming the prime tower

BenJule:
## Summary

- ignore synthetic prime tower object IDs during instance update notifications
- cover both direct and selection-wide notification paths
- prevent prime tower transforms from indexing the model object list out of bounds

Prime towers use synthetic object IDs starting at 1000. Transform operations share the normal instance-update path, which previously treated those IDs as indices into `Model::objects`. Rotating, scaling, or mirroring a selected prime tower could therefore access invalid memory.

## Testing

- syntax-checked `Selection.cpp` using the existing Release GUI build flags
- verified guards for both direct and selection-wide update notifications


Haidiye00 (commented):

PR #11655, opened 2026-07-21
Guard stale PartPlate instance IDs

BenJule:
## Summary

- validate both object and instance IDs before dereferencing PartPlate entries
- reuse the existing instance bounds check across all affected plate scans
- allow the bounds helper to be used by the const CLI extruder scan

Deleting an instance can temporarily leave a stale object/instance pair in a plate set. Several scans checked only the object ID before indexing the object's instance vector, which could access a removed instance.

The affected paths include printable-state queries, CLI extruder collection, plate movement, and instance duplication.

## Testing

- syntax-checked `PartPlate.cpp` using the existing Release GUI build flags
- verified seven new instance bounds checks and matching const declarations

### DIFF
--- a/src/slic3r/GUI/PartPlate.cpp
+++ b/src/slic3r/GUI/PartPlate.cpp
@@ -341,7 +341,7 @@ void PartPlate::set_spiral_vase_mode(bool spiral_mode, bool as_global)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Haidiye00 (changes requested):

Issue #31, opened 2022-07-18
No 3rd party printers supported!


valentt:
Bambu Lab is based upon lots of open source code from Slic3r, Prusa slicer and SuperSlicer. All of them support multiple printers.

We would like to have print farm with Bambu Lab machines, but they will never be only machine in our shop. So why would you reduce usage of other 3D printers in Bambu Studio?

We would like to have one single workflow for all our machines.

wuyifanfancy:
Thanks for the suggestion. We are already working on it.
We will update the code and provide a guide when the developing work is done.

valentt:
Good to hear that we will see other printer profiles also, can't wait. Any ideas on the timeline?

SaltWei:
guide to add 3-rd party printer into BambuStudio
https://wiki.bambulab.com/en/software/add-third-party-printer-in-studio

SaltWei:
Already include 3-rd party printer in latest release.

PR #11687, opened 2026-07-24
Confirm before clearing recent projects

### LINKED ISSUE #8351: Bad placement of button 'Clear All' in Recently Opened
### Bambu Studio Version

2.2.2.56

Bambu Lab Official website

### OS version

macOS 15.6.1

_No response_

### Printer

Not related

### How to reproduce

1. Open BambuStudio
2. Click Home icon on top
3. If you have selected Library, click "MORE>" on the right of Recently Opened
4. And here is problem, in the same place you have button for clearing all last opened projects, It should be in different place to avoid accidental clearing 

### Actual results

after clicking more, there is a bit of lag and sometimes I click one more time to see more last opened models, and then it clears all of last opened models

### Expected results

Clear All button should be in different place
Or have confirmation that you are sure to clear all last opened models

It is just bad design of application. It is not related with projects.

- [ ] Log file
- [ ] Project file

BenJule:
## Summary

- add a confirmation dialog before clearing all recent projects
- make **No** the default choice to prevent accidental deletion
- leave removal of individual recent projects unchanged

## Testing

- ran `git diff --check`
- verified the Clear All confirmation control flow
- verified that the dialog only appears for the empty-filename Clear All action
- no local full build; the change is limited to the existing recent-project handler

Fixes #8351

### DIFF
--- a/resources/web/homepage3/js/home.js
+++ b/resources/web/homepage3/js/home.js
@@ -482,22 +482,11 @@ function OnDeleteRecentFile( )
 
 function OnDeleteAllRecentFiles()
 {
-	showConfirmDialog({
-		title: GetCurrentTextByKey("t39"),
-		message: GetCurrentTextByKey("t149"),
-		okText: GetCurrentTextByKey("t21"),
-		cancelText: GetCurrentTextByKey("t22"),
-		onOk: function() {
-			$('#FileList').html('');
-			UpdateRecentClearBtnDisplay();
-			
-			var tSend={};
-			tSend['sequence_id']=Math.round(new Date() / 1000);
-			tSend['command']="homepage_delete_all_recentfile";
-			
-			SendWXMessage( JSON.stringify(tSend) );
-		}
-	});
+	var tSend={};
+	tSend['sequence_id']=Math.round(new Date() / 1000);
+	tSend['command']="homepage_delete_all_recentfile";
+
+	SendWXMessage( JSON.stringify(tSend) );
 }
 
 function UpdateRecentClearBtnDisplay()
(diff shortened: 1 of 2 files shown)


tonghao-bbl:
image

Hi @BenJule can we use a native dialog for this?

BenJule:
Thanks @tonghao-bbl. Yes, I've updated the confirmation to use the native `wxMessageDialog` instead of Bambu Studio's custom `MessageDialog` wrapper.

`No` remains the default choice, so the existing protection against accidentally clearing the recent-project list is unchanged.

I also ran a targeted compile check of `MainFrame.cpp` using the existing Release build flags, and `git diff --check` passes.

Updated commit: `767237da8e`

Could you please take another look? Thanks!

tonghao-bbl:
> Thanks @tonghao-bbl. Yes, I've updated the confirmation to use the native `wxMessageDialog` instead of Bambu Studio's custom `MessageDialog` wrapper.
> 
> `No` remains the default choice, so the existing protection against accidentally clearing the recent-project list is unchanged.
> 
> I also ran a targeted compile check of `MainFrame.cpp` using the existing Release build flags, and `git diff --check` passes.
> 
> Updated commit: `767237da8e`
> 
> Could you please take another look? Thanks!

Thanks for you quick action. Actually I mean we can use a native looking widget, anyway can you take a screen short of this message dialog so we can make this review faster?

BenJule:
Thanks @tonghao-bbl. I retested the final version with Jenkins build #31 (commit `ab863b406d81`).

The previous web confirmation has been removed, so **Clear all** now opens only the native-looking `wxMessageDialog`.

I verified both paths:
- `No` is the default choice and leaves the recent-project list unchanged.
- `Yes` clears the recent-project list as expected.

### Before

The previous web confirmation:

Before - web confirmation dialog

### After

The final native-looking confirmation dialog:

After - native wxMessageDialog

tonghao-bbl:
@zzsoccer214 Please take a look on Mac if the dialog looks good when available

PR #11694, opened 2026-07-25
feat: add option to exclude timelapse time from the print time estimate

pdelteil:
### What

Adds an optional print setting, **Exclude timelapse time from estimate** (`exclude_timelapse_from_estimate`), which subtracts the duration of the timelapse blocks from the reported print time estimate. It defaults to `false`, so existing behaviour is unchanged unless the user opts in.

### Why

With timelapse enabled, the estimate includes the per-layer timelapse moves. On prints with many short layers this inflates the number noticeably relative to the same model sliced without timelapse, which makes it awkward to compare profiles or to estimate machine time when the timelapse footage is a by-product rather than the goal.

The timelapse still runs — only the reported estimate changes.

### How

The timelapse duration is already tracked as `skippable_part_time[SkipType::stTimelapse]`. When the option is on, that value is subtracted in two places so the preview and the G-code agree:

- `GCodeProcessor::update_estimated_times_stats()` — the statistics shown in the preview.
- `GCodeProcessor::TimeProcessor::post_process()` — the `; model printing time:` / `; estimated printing time` header written into the G-code. The value is carried in via `TimeProcessContext`.

`skippable_part_time` is accumulated across all enabled time modes, so the total is spread evenly over the enabled modes rather than subtracted in full from each one; otherwise it would over-subtract whenever more than one mode is enabled. Both call sites clamp at zero.

The preview legend (`BaseRenderer.cpp`) zeroes its local `timelapse_time` when the option is on, because `modes[].time` has already been reduced by it — this keeps the legend self-consistent and avoids subtracting the timelapse portion twice from the displayed model printing time.

The option is registered in `s_Preset_print_options` and appears in the Process tab under Others, directly below the existing `timelapse_type` setting.

### Scope / non-goals

- The printer's live remaining-time countdown is unaffected; this only changes what the slicer reports.
- No change to how timelapse G-code is generated, and no change in behaviour when the option is off.

### Testing

Built on Linux (GCC, Release, `SLIC3R_GUI=ON`) against `master`; the branch is a clean cherry-pick onto the current head and no upstream commit since has touched the seven files involved.

Happy to adjust the option name, its tooltip wording, or where it sits in the Process tab if you'd prefer something different.

### DIFF
--- a/src/libslic3r/GCode/GCodeProcessor.cpp
+++ b/src/libslic3r/GCode/GCodeProcessor.cpp
@@ -667,22 +667,36 @@ void GCodeProcessor::TimeProcessor::post_process(const std::string& filename, st
                 }
             }
             else if (line == reserved_tag(ETags::Estimated_Printing_Time_Placeholder)) {
+                // BBS: optionally drop the timelapse block time from the reported estimate.
+                // timelapse_part_time is accumulated over all enabled modes, so spread it evenly.
+                float timelapse_time_per_mode = 0.0f;
+                if (context.exclude_timelapse_time && context.timelapse_part_time > 0.0f) {
+                    size_t enabled_modes = 0;
+                    for (const auto& m : machines)
+                        if (m.enabled)
+                            ++enabled_modes;
+                    if (enabled_modes == 0)
+                        enabled_modes = 1;
+                    timelapse_time_per_mode = context.timelapse_part_time / static_cast<float>(enabled_modes);
+                }
                 for (size_t i = 0; i < static_cast<size_t>(PrintEstimatedStatistics::ETimeMode::Count); ++i) {
                     const TimeMachine& machine = machines[i];
                     PrintEstimatedStatistics::ETimeMode mode = static_cast<PrintEstimatedStatistics::ETimeMode>(i);
                     if (mode == PrintEstimatedStatistics::ETimeMode::Normal || machine.enabled) {
                         char buf[128];
+                        const float total_time = std::max(0.0f, machine.time - timelapse_time_per_mode);
+                        const float model_time = std::max(0.0f, total_time - machine.prepare_time);
                         if (!s_IsBBLPrinter) {
                             // Klipper estimator
                             sprintf(buf, "; estimated printing time (normal mode) = %s\n",
-                                get_time_dhms(machine.time).c_str());
+                                get_time_dhms(total_time).c_str());
                             ret += buf;
                         }
                         else {
                             // BBS estimator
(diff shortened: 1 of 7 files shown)


BenJule (commented):
Thanks for working on this. The idea is useful and the PR is cleanly scoped, but I found two correctness issues that should be addressed before merging.

## 1. Blocking: Timelapse time is averaged across time modes

`skippable_part_time[stTimelapse]` is a single shared value. However, `handle_time_block()` adds the timelapse block duration for every enabled `TimeMachine` mode.

The block duration is calculated separately for each mode using different speed, acceleration, and jerk limits, so Normal and Stealth mode can have different timelapse durations.

The PR later divides the combined value only by the number of enabled modes:

    timelapse_time_per_mode =
        it->second / static_cast<float>(enabled_modes);

and subtracts that average from every mode.

For example:

    Normal mode:   40 s timelapse
    Stealth mode:  80 s timelapse
    Stored total: 120 s
    Average:       60 s

    Normal:  20 s too much is subtracted
    Stealth: 20 s too little is subtracted

This makes both the preview estimate and the values written to the G-code header incorrect whenever multiple time modes are enabled and their movement durations differ.

The timelapse duration should be stored per `ETimeMode`, for example in an array or a mode-specific map. `handle_time_block()` already receives `activate_machine_idx`, so the relevant mode index is available.

## 2. The detailed time breakdown no longer matches the reduced total time

The PR reduces only:

    data.time

but these values still include the timelapse blocks:

    data.custom_gcode_times
    data.moves_times
    data.roles_times
    data.layers_times

The preview then calculates percentages using the unchanged role or movement times divided by the already reduced total time.

This can result in:

- accumulated percentages exceeding 100%,
- layer times no longer matching the displayed total time,
- inflated travel percentages,
- division by zero or infinite percentages if all estimated time is removed.

Setting `timelapse_time` to zero in `BaseRenderer.cpp` only changes the small summary display. It does not make the underlying statistics consistent.

## Tests / CI

The PR description mentions a local Linux Release build, but GitHub did not run any actual CI jobs for this commit. The “Build all” workflow completed with no jobs, and there are no automated regression tests covering this behavior.

Before approval, I recommend testing at least:

1. Option disabled and enabled with Normal mode only.
2. Normal and Stealth modes enabled with different machine limits.
3. Preview time compared against the G-code header values.
4. Consistency between total time, layer times, movement percentages, and role percentages.

The single-mode case may appear correct, but the current implementation loses the association between timelapse duration and its corresponding time mode. Because that affects the core result of this feature, I recommend addressing these issues before merging.

pdelteil:
Thanks. 

Is more complicated than expected. Just closing it.

PR #11699, opened 2026-07-25
Guard empty float-or-percent input

BenJule:
## Summary

- guard both float-or-percent parsing paths before calling `std::string::back()`
- preserve the existing percentage parsing behavior
- let empty input continue into the existing exception-based validation path

## Problem

`change_opt_value()` called `str.back()` for `coFloatOrPercent` and `coFloatsOrPercents` without first checking whether the input string was empty.

Calling `std::string::back()` on an empty string is undefined behavior and may crash before the surrounding exception handler can report the invalid value.

## Testing

- verified both parsing paths contain the empty-string guard
- ran `git diff --check`
- ran a targeted `GUI.cpp` syntax check using an existing compile database


tonghao-bbl:
LGTM @lanewei120

lanewei120:
thanks @BenJule

PR #11702, opened 2026-07-25
Recover from incomplete framebuffer creation

BenJule:
## Summary

Recover cleanly when framebuffer creation is only partially completed.

## Changes

- release partially created framebuffer resources after setup failures
- make framebuffer binding report success or failure
- avoid caching a failed framebuffer as currently bound
- validate both blit endpoints before changing read/draw framebuffer bindings
- avoid reads and resolves through invalid framebuffer IDs

## Validation

- `git diff --check`
- changed-file scope verified
- one GPG-signed commit

A local GUI compile entry is not available. Full native validation remains delegated to Jenkins.

### DIFF
--- a/src/slic3r/GUI/OpenGLManager.cpp
+++ b/src/slic3r/GUI/OpenGLManager.cpp
@@ -560,8 +560,13 @@ void OpenGLManager::_bind_frame_buffer(const std::string& name, EMSAAType msaa_t
             current_framebuffer->unbind();
         }
     }
-    m_name_to_framebuffer[name]->bind();
-    m_current_binded_framebuffer = m_name_to_framebuffer[name];
+    const auto& target_framebuffer = m_name_to_framebuffer[name];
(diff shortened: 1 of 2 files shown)


src/slic3r/GUI/OpenGLManager.cpp
```diff
@@ -1311,10 +1358,10 @@ void FrameBuffer::resolve()
             glsafe(::glDrawBuffers((GLsizei)1, bufs));
         }
 
-        const bool rt = check_frame_buffer_status();
-        if (!rt)
+        if (!check_frame_buffer_status())
         {
-            glsafe(BBS_GL_EXTENSION_FUNC(::glBindFramebuffer)(BBS_GL_EXTENSION_PARAMETER(GL_FRAMEBUFFER), 0));
+            release();
```
guanyun-gudujian, line 1363: why need to release all resource here?

just to bind to the default fb0 zero is enough i think

PR #11704, opened 2026-07-25
Use global AMS tray ID for calibration hold

BenJule:
## Summary

Use the AMS-type-aware global tray identifier when protecting recently written calibration values.

## Root cause

The calibration hold logic compared against an AMS identifier that remained initialized to zero and reconstructed tray positions with `/ 4` and `% 4`. That does not represent all AMS types correctly.

## Changes

- resolve the current tray through `DevAms::GetTrayId()`
- reuse that global identifier for existence bits
- compare the complete global tray ID during the calibration hold window
- keep the change limited to the device parser

## Validation

- source-contract checks for global tray-ID creation and comparison
- `git diff --check`
- one GPG-signed commit

Full native validation remains delegated to Jenkins.

### DIFF
--- a/src/slic3r/GUI/DeviceCore/DevFilaSystem.cpp
+++ b/src/slic3r/GUI/DeviceCore/DevFilaSystem.cpp
@@ -831,11 +831,17 @@ DevAmsTray* DevFilaSystemParser::ParseAmsTrayInfo(const json& j_tray, MachineObj
         }
     }
 
-    int ams_id_int = 0;
     int tray_id_int = 0;
+    int global_tray_id = -1;
     try {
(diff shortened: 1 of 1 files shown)


hemai249:
@BenJule  Thank you for your patch. Due to rather tricky compatibility issues with tray_id across firmware versions for different device models, we have been gradually reducing the use of tray_id during development, as ams_id and slot_id provide clearer expression. Regarding the issue you reported, we will arrange for our developers to investigate and fix it.

BenJule:
Understood, thank you for the explanation and for taking the time to review the patch. That makes sense given the tray_id compatibility concerns across device models and firmware versions. I appreciate that your developers will investigate the underlying issue. I’ll close this PR accordingly.

PR #11712, opened 2026-07-25
Update BambuStudio_tr.po

fatih5228:
Update Turkish language pls


qusheng-bambu:
@fatih5228 Thank you very much for your contribution. However, we are very sorry to inform you that we have already translated it into Turkish internally and it will be included in the next version. I'm very sorry that we didn't share our roadmap with the community earlier, which led to duplicated work. We really appreciate the effort you put in. Unfortunately, we have to close this PR.

PR #11741, opened 2026-07-28
Add geometry-only 3MF import deep link

jsandfordhughes:
## Why

We need a way to open a remote 3MF in the current Bambu Studio project without starting a new project or replacing the user’s existing setup.

The existing `bambustudio://open` action treats the downloaded 3MF as a project. This replaces the current project and may change its printer, filament, process, and plate settings.

This PR introduces a separate import action for cases where only the models from the remote 3MF should be added to the current project.

## What changed

A new backward-compatible deep-link action is available:

`bambustudio://import?file=<encoded-url>`

After downloading the 3MF, Bambu Studio sends it through the normal geometry-only model import workflow. This means:

- Models are added to the currently open project.
- The current printer, filament, process, and plate settings are preserved.
- The current project filename is retained.
- The import is added to the undo history.
- The project is marked as modified after a successful import.
- If no editable project is open, an empty project is created first.

The existing `bambustudio://open` behaviour is unchanged and continues to open the downloaded 3MF as a project.

The new action works during startup and through single-instance forwarding on Windows, macOS, and Linux. Legacy macOS deep links remain supported.

## Security

The import action uses the existing remote-download pipeline and does not widen its trust boundary. It retains the existing:

- Trusted-site confirmation
- URL decoding and sanitisation
- `.3mf` file validation
- 500 MB size limit
- Retry and cancellation handling
- Download progress UI
- Temporary-file and atomic-rename handling

Forwarded arguments are accepted as deep links only when they match a supported action and contain a non-empty `file` parameter.

## Verification

- Added focused tests for `open`, `import`, legacy macOS links, malformed links, unsupported actions, missing parameters, and forwarding classification.
- Built `slic3rutils_tests`, `libslic3r_gui`, and `BambuStudio`.
- Ran the focused deep-link tests successfully.
- Validated both macOS plist templates.
- Ran `git diff --check`.

### DIFF
--- a/src/slic3r/GUI/DeepLink.hpp
+++ b/src/slic3r/GUI/DeepLink.hpp
@@ -1,0 +1,39 @@
+#ifndef slic3r_GUI_DeepLink_hpp_
+#define slic3r_GUI_DeepLink_hpp_
+
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace Slic3r {
+namespace GUI {
+
+enum class RemoteModelAction : unsigned char
+{
+    OpenProject,
+    ImportModel
+};
+
+struct StudioDeepLink
+{
+    RemoteModelAction action;
+    std::string       encoded_download_info;
+    bool              legacy_macos_scheme { false };
+};
+
+struct RemoteModelDownloadRequest
+{
+    RemoteModelAction action;
+    std::string       download_info;
+};
+
+// Recognizes only Bambu Studio's supported model-download deep links. The
+// returned payload is intentionally left encoded; URL decoding and trust
+// validation remain in GUI_App, where the existing confirmation UI lives.
+std::optional<StudioDeepLink> parse_studio_deep_link(std::string_view url);
+bool is_studio_deep_link_candidate(std::string_view url);
+
+} // namespace GUI
+} // namespace Slic3r
+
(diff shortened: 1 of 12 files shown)


src/slic3r/GUI/DeepLink.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace Slic3r {
+namespace GUI {
+
+enum class RemoteModelAction : unsigned char
+{
```
tonghao-bbl, line 14: comment is appreciated

src/slic3r/GUI/DeepLink.hpp
```diff
@@ -0,0 +1,39 @@
...
+    RemoteModelAction action;
+    std::string       encoded_download_info;
+    bool              legacy_macos_scheme { false };
+};
+
+struct RemoteModelDownloadRequest
+{
+    RemoteModelAction action;
+    std::string       download_info;
```
tonghao-bbl: a url or something more?

src/slic3r/GUI/DeepLink.hpp
jsandfordhughes: Done 👍

PR #11833, opened 2026-08-08
FIX: Lower-case domain-name in D-Bus interface names

hadess:
Match D-Bus interface names to the (verified) Flatpak app on Flathub, as well as the AppStream naming guidelines:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

and lower-case the reverse domain name portion of the D-Bus interface and object names.

This will also fix the Flatpak being able to own the Instance object.

Fixes: a803118282d8 ("Fix DBus name case mismatch breaking Linux single-instance file passing")

### DIFF
--- a/src/slic3r/GUI/InstanceCheck.cpp
+++ b/src/slic3r/GUI/InstanceCheck.cpp
@@ -236,10 +236,10 @@ namespace instance_check_internal
 			dbus_uint32_t 	serial = 0;
 			const char* sigval = message_text.c_str();
 			//std::string		interface_name = "com.prusa3d.prusaslicer.InstanceCheck";
-			std::string		interface_name = "com.BambuLab.BambuStudio.InstanceCheck.Object" + version;
+			std::string		interface_name = "com.bambulab.BambuStudio.InstanceCheck.Object" + version;
 			std::string   	method_name = "AnotherInstance";
 			//std::string		object_name = "/com/prusa3d/prusaslicer/InstanceCheck";
-			std::string		object_name = "/com/BambuLab/BambuStudio/InstanceCheck/Object" + version;
+			std::string		object_name = "/com.bambulab/BambuStudio/InstanceCheck/Object" + version;
 
 
 			// initialise the error value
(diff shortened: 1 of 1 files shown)


src/slic3r/GUI/InstanceCheck.cpp
```diff
@@ -538,7 +538,7 @@ namespace MessageHandlerDBusInternal
 	        "       <arg name=\"data\" direction=\"out\" type=\"s\" />"
 	        "     </method>"
 	        "   </interface>"
-	        "   <interface name=\"com.BambuLab.BambuStudio.InstanceCheck\">"
+	        "   <interface name=\"com.bambulab.BambuStudio.InstanceCheck\">"
```
tonghao-bbl, line 541: `BambuStudio` still uses capital letters, will you fix this too?

hadess: No, that's not part of the "reverse domain name", and that already matches the [com.bambulab.BambuStudio application ID](https://github.com/flathub/com.bambulab.BambuStudio/).

tonghao-bbl: image
To be honest, I'm not familar with this. but isn't all the string `com.BambuLab.BambuStudio.InstanceCheck` an `ID`？

does this PR want to make the ID comply to the AppStream naming guidelines or fix a bug in flatpak or something compromised of the two?

hadess: > To be honest, I'm not familar with this. but isn't all the string `com.BambuLab.BambuStudio.InstanceCheck` an `ID`？
> 
> does this PR want to make the ID comply to the AppStream naming guidelines or fix a bug in flatpak or something compromised of the two?

It's a compromise, but the important part is that we're basically making it match the **verified** Bambu Studio application ID, which we can't change without a huge amount of work.

I can change the commit message if that's not clear enough.

tonghao-bbl: > I can change the commit message if that's not clear enough.

That would be great since I was confused about this.

BTW, Thank you for the work on the flatpak, we don't have so many resources on linux.

hadess: Done, should be clearer now. Please reopen the conversation if it's still too confusing.

tonghao-bbl:
@lanewei120 LGTM

Issue #6220, opened 2025-03-27
Feature Request: Customizable Axes for 3Dconnexion Mice in Bambu Studio


AdditivprintMICHA:
Dear Bambu Lab Team,

I would like to request a feature enhancement for Bambu Studio regarding the use of 3Dconnexion mice. Currently, it is not possible to customize or invert the axes of these devices within the software.

Adding an option to adjust the axes via simple checkboxes would significantly improve the user experience, allowing for greater flexibility and compatibility with individual preferences. Many users rely on 3Dconnexion devices for precise navigation, and this feature would make Bambu Studio even more user-friendly.

I appreciate your consideration and hope this feature can be implemented in a future update.

Best regards,
Michael Legelr

PR #11928, opened 2026-08-17
ENH: Add "Lock horizon" option for 3Dconnexion devices

timkarnold:
## What
Add a **Lock horizon** option to the 3Dconnexion settings dialog. When enabled, 3D mouse rotation orbits on a sphere about the target instead of rotating freely about the camera local axes, so the horizon stays level.

BambuStudio SpaceMouse Lock Horizon

## Why
Rotation is applied with `rotate_local_around_target()`, which has no world-up constraint. Roll accumulates as the view is reoriented and never returns to level, leaving the scene progressively tilted. 3Dconnexion drivers offer this as "Lock Horizon" for NavLib applications, but Bambu Studio uses the legacy 3DconnexionClient path, so the driver shows no Navigation tab for it and the option cannot come from that side.

## Changes (4 files, +31/-4)
- `slic3r/GUI/Mouse3DController.{cpp,hpp}` — `lock_horizon` param, dialog checkbox, and the rotation branch: `rotate_on_sphere(-rot.z(), rot.x(), true)`, so azimut turns about the world Z axis and zenit about the camera right vector — what a mouse drag already does. The roll axis is dropped, having nowhere to go under a locked horizon.
- `libslic3r/AppConfig.{cpp,hpp}` — persist `lock_horizon` per device, alongside `swap_yz`.

Ticking the box also levels the current view through the existing `Camera::recover_from_free_camera()`. Orbiting on a sphere only stops roll from growing, so without this any roll already accumulated would be frozen in place as a permanently tilted horizon.

## Testing
Manual, macOS 26 / arm64, 3Dconnexion SpaceMouse Wireless BT: the horizon holds level through combined orbit/pan/zoom, levels correctly when enabled from an already-tilted view, and the setting survives a restart and opening other projects. Top view is unaffected — `recover_from_free_camera()` no-ops when the right vector is already horizontal.

## Scope
Defaults to off, so existing free rotation is unchanged for anyone who prefers it. Axis direction/inversion is a separate concern (#6220) and is not touched here.

### DIFF
--- a/src/libslic3r/AppConfig.cpp
+++ b/src/libslic3r/AppConfig.cpp
@@ -1284,7 +1284,7 @@ void AppConfig::set_recent_projects(const std::vector<std::string>& recent_proje
 }
 
 void AppConfig::set_mouse_device(const std::string& name, double translation_speed, double translation_deadzone,
-                                 float rotation_speed, float rotation_deadzone, double zoom_speed, bool swap_yz)
+                                 float rotation_speed, float rotation_deadzone, double zoom_speed, bool swap_yz, bool lock_horizon)
 {
     std::string key = std::string("mouse_device:") + name;
     auto it = m_storage.find(key);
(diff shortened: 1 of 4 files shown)


Haidiye00 (approved):
approve

Issue #11785, opened 2026-08-02
Better 3D viewport UX - Navigation and extra options.


alizreik-au:
**Is your feature request related to a problem? Please describe.**
The viewport is killing me! It rotates around the centre of the plate and the FOV is insanely zoomed in.
It has been a pain to navigate and perform work in the viewport.
- Tight FOV
- Unintuitive camera orbit and rotation
- Gizmo is finicky and fragile with a punishingly small hit box on the axes.

**Describe the solution you'd like**
Extra options and setup for the viewport.
- Scalar input for FOV, generally 14-140 degrees with a 90 degree default. But be generous with limits please.
- Create options for navigation:
-> Allow users to set the centre of interest to the centre of the viewport for orbiting.
-> Optional hotkeys set same to industry standard for 3D software, like:
--> `Q` to select
--> `W` to translate
--> `E` to rotate
--> `R` to scale
--> `F` to focus on object selection.
- For the Gizmo, add a padding of at least 3-5px around each axes for the user to easily click on.
- Scalar input for front clipping point of camera. So we can visually clip through objects as we get closer.
- Add custom clipping cutouts on X, Y and Z axes to allow users to visually see cutout profiles of their models and still work on them.

**Describe alternatives you've considered**
Using Blender instead of BambuStudio.

Issue #6806, opened 2025-05-03
Visual and "object" issue


SmogDragon:
### Bambu Studio Version

2.0.3.54

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 10 Home Build 19045.5737

### Additional system information

CPU : 8th Generation Intel(R)Core(TM ) i7-8750H Processor (6-Core, 9MB Cache, up to 4.1GHz w/ Turbo Boost)
Memory : 16GB, 2x8GB, DDR4, 2666MHz
GPU : NVIDIA(R) GeForce(R) GTX 1060 with NVIDIA(R) Max Q Design te chnology, 6GB GDDR5 video memory

### Printer

X1C

### How to reproduce

-> I opened a 3mf with 10 plates and a for models
-> Dragged a model on an existing empty plate
-> Selected Rotate
-> New object and plate was close to the bottom edge of the screen
-> All "arrows" was visible but at the edge of the screen. let say +/- 10mm from the edge
-> Went to clic on the red arrow
-> Animation keep flikering as the new bigger "wheel" and probably "hit box" aren't exactly at the same place

### Actual results

-> I had 9 plates with a few models on them. I needed to rotate one to be flat on the bed. Selected the rotation tool, but couldn't clic on the red arrow because there is an animation making this "wheel" bigger. Since it was close to the edge, the animation seems to move the "hit box" and it was going back and forth in that animation. As the new hit box doesn't seems to be a the same place as the previous one.
-> A bit later, moving around in the environement wasn't constent at all. Some mouse drag moved everything normally and others moved REALLY slow. Not in a laggy, jumping was as if I was missing powers, could be though this hardward is far from new and recent. But more like the value of the "drag motion" was different. The animation was smooth. The total movement value was way different.
-> Still in the rotation function, after moving around a bit to avoid the hit box issue, no model was visible. They all came back fine when I selected a different fonction.

All three issue was fresh after the update in the same 5min. I ended fine formating and sending the print. It's more a annoyance and might lead to mistrust of part being at the right place.

### Expected results

Was expecting to be able to rotate the part without having visual issue.

### Project file & Debug log uploads

Sorry, those files have been created after sending the print, not exactly after the issue, as it's somewhat a minor issue that didn't prevent me to get the print ready to send.

[log.zip](https://github.com/user-attachments/files/20024118/log.zip)

[...]

Issue #5845, opened 2025-01-23
Turn, Scale and move Anchors gets VERY small


KimA998:
### Bambu Studio Version

1.10.1.50

### Where is the application from?

Bambu Lab github releases

### OS version

Windows 11

### Additional system information

Quite often the anchors when moving, scaling or turning an object get VERY small and hard to hit with the mouse.

Only solution is to close Bambu Studio and restart it.

### Printer

X1C

### How to reproduce

I haven´t figured it out yet how it happens - but it does!

### Actual results

You need to restart Studio to have "normal" size anchors again.

### Expected results

I didn´t expect the anchors to change in size!!

### Project file & Debug log uploads

[Nimbus Print 1.1.zip](https://github.com/user-attachments/files/18523367/Nimbus.Print.1.1.zip)

![Image](https://github.com/user-attachments/assets/798e2967-6f15-4deb-bec8-d1e55c793788)

### Checklist of files to include

- [ ] Log file
- [ ] Project file

Haidiye00:
Hello, we will try to reproduce it. If you can find a stable way to reproduce it, you can record a GIF and send it to us. Thank you very much.

And can you download ScreenToGif software and record a GIF of the operation video and upload it? Thank you
[Release ScreenToGif 2.41.1 · NickeManarin/ScreenToGif](https://github.com/NickeManarin/ScreenToGif/releases/tag/2.41.1)

Haidiye00:
Hello, we have released software V2.0. Please download and use it
https://github.com/bambulab/BambuStudio/releases/tag/v02.00.00.95

Issue #5035, opened 2024-10-18
Dynamic scaling of arrow handles (grabber) 


suchabadperson:
In the latest beta version of Bambu Studio (v1.10.0 - v01.10.00.74, I appreciate the newly added Grabber scale option, which helps adjust the size of the handles for selecting and moving objects. However, there is still room for improvement in terms of usability.

I would like to suggest implementing a dynamic arrow handle similar to what is found in PrusaSlicer. In PrusaSlicer, the handles scale dynamically based on the camera’s distance from the object: as you zoom in closer to an object, the handles grow in size, and as you zoom out, they become smaller. This would significantly improve the experience of manipulating objects, as it allows for more precise interaction when zoomed in and avoids clutter when zoomed out.

![](https://user-images.githubusercontent.com/65729856/221362564-a84cd774-84a5-490b-8c94-b113e485a925.png)

PR #11960, opened 2026-08-20
FIX: keep gizmo screen size constant under a perspective camera

vvzvlad:
## The problem

With default settings the gizmos (move / rotate / scale) do not keep a constant size on screen. Their size depends on where the selected object is relative to the point the camera orbits around. In a project with many plates, the gizmo of an object on a far plate becomes tiny, and it becomes huge when the object is closer to the camera than the orbit point. It also changes while panning the view, without touching the zoom.

Default settings here means `gizmo_keep_screen_size = true` and the perspective camera, both are defaults (`AppConfig.cpp` sets `gizmo_keep_screen_size` to true, `Camera.hpp` has `EType m_type{ EType::Perspective }`).

Reports that look like this problem: #5845, #3753, #6806, #5035, #11785.

## Where it comes from

Gizmos compute a world size and cancel the zoom with `INV_ZOOM = 1 / camera.get_zoom()`:

```cpp
// GLGizmoBase.cpp, get_grabber_size()
grabber_size = Grabber::FixedGrabberSize * Grabber::GrabberSizeFactor * INV_ZOOM;

// GLGizmoBase.cpp, modify_radius()
radius  = 0.2f * std::min(t_width, t_height);
radius *= GLGizmoBase::INV_ZOOM;
```

This cancels the projection only when the projection is orthographic. In `Camera::apply_projection()` the perspective branch does:

```cpp
// scale near plane to keep w and h constant on the plane at z = m_distance
const double scale = m_frustrum_zs.first / m_distance;
w *= scale;
h *= scale;
```

If you expand the resulting frustum matrix, `near_z` cancels out and the projected size of a world space object becomes

```
screen_px = world_size * zoom * (m_distance / depth)
```

`zoom` is cancelled by `INV_ZOOM`, but `m_distance / depth` is not cancelled by anything. `m_distance` is the distance from the camera to its orbit target, `depth` is the eye space depth of the object. They are equal only for an object sitting exactly on the orbit plane, and the camera orbits the center of the plate, not the selection. So the further the selected object is from that plane, the more wrong the size is.

Two more things make it move on its own:

- panning changes the target (`camera.set_target(...)` in `GLCanvas3D.cpp`), and with `zoom_to_mouse` the wheel calls `camera.translate(displacement)`, so `depth` changes on almost every navigation action;
- `Camera::calc_tight_frustrum_zs_around()` calls `set_distance(m_distance + delta)` when the near plane hits `FrustrumMinNearZ`, so `m_distance` itself changes depending on what is inside the frustum.

## Numbers

This does not need a build. The script below builds the same `glOrtho` / `glFrustum` matrices from the values used in `apply_projection()`, and projects the gizmo radius produced by `modify_radius()`:

<details>
<summary>proof.py</summary>

```python
#!/usr/bin/env python3
# Reproduces Camera::apply_projection() + GLGizmoBase::modify_radius() with plain arithmetic.
# No OpenGL and no build needed: it only builds the same glOrtho / glFrustum terms the code builds.

VP, ORBIT = (2560, 1440), 1000.0      # viewport in px, Camera::DefaultDistance in mm


def ring_px(depth, perspective, zoom=3.0):
    inv_zoom = 1.0 / zoom
    radius = 0.2 * min(VP) * inv_zoom                  # GLGizmoBase::modify_radius()
    w = 0.5 * VP[0] * inv_zoom                         # Camera::apply_projection()
    near = max(depth - 60.0, 100.0)                    # tight frustum, FrustrumMinNearZ = 100
    if perspective:
        w *= near / ORBIT                              # "scale near plane to keep w and h constant"
        clip_x, clip_w = near * radius / w, depth      # glFrustum
    else:
        clip_x, clip_w = radius / w, 1.0               # glOrtho
    return clip_x / clip_w * 0.5 * VP[0]


print("gizmo ring, expected constant %d px
" % (0.2 * min(VP)))
print("depth mm |    ortho | perspective")
for depth in (250, 500, 1000, 2000, 3000, 4000):
    ortho, persp = ring_px(depth, False), ring_px(depth, True)
    print("%8d | %6.1f px | %7.1f px  (%.2fx)" % (depth, ortho, persp, persp / ortho))

print("
same object at depth 3000, user zooms in and out:")
for zoom in (1.0, 2.0, 4.0, 8.0):
    print("  zoom %4.1f -> %5.1f px" % (zoom, ring_px(3000, True, zoom)))
```

</details>

Output:

```
gizmo ring, expected constant 288 px

depth mm |    ortho | perspective
     250 |  288.0 px |  1152.0 px  (4.00x)
     500 |  288.0 px |   576.0 px  (2.00x)
    1000 |  288.0 px |   288.0 px  (1.00x)
    2000 |  288.0 px |   144.0 px  (0.50x)
    3000 |  288.0 px |    96.0 px  (0.33x)
    4000 |  288.0 px |    72.0 px  (0.25x)

same object at depth 3000, user zooms in and out:
  zoom  1.0 ->  96.0 px
  zoom  2.0 ->  96.0 px
  zoom  4.0 ->  96.0 px
  zoom  8.0 ->  96.0 px
```

The orthographic column is exactly `0.2 * min(viewport)` at any depth, which is what the code intends. The perspective column is off by exactly `orbit_distance / depth`, from 4x too big to 4x too small in one scene. The second block changes the zoom by 8x and the size does not move at all, which shows `INV_ZOOM` compensates the zoom and only the zoom.

## Steps to reproduce in the app

1. View -> Use Perspective View (default).
2. Put the same object on plate 1 and on a far plate, say plate 10.
3. Select the one on plate 1, press R, look at the ring. Then select the one on the far plate, press R again.
4. The ring is much smaller on the far plate, and the arrow heads too.
5. Now View -> Use Orthogonal View and repeat: both rings are the same size.

Step 5 is the check that this is the projection and not something else, because the sizing code is the same in both cases.

## The fix

Cancel the missing term: multiply the screen space sizes by `depth / orbit_distance`, where `depth` is the eye space depth of the selection center.

`Camera::get_distance()` and `Camera::get_target()` are not const, so they cannot be called from the const render path. But the perspective branch already stores `m_gui_scale = near_z / m_distance`, so the orbit distance is `get_near_z() / get_gui_scale()`, and both getters are const.

The correction only applies when it is needed, and stays `1.0f` otherwise:

- orthographic camera: nothing to cancel, the old math is already exact;
- `gizmo_keep_screen_size == false`: that path is left bit identical, see the known limits below;
- nothing selected, or the anchor is in front of the near plane.

The factor is clamped to `[0.1, 8.0]`. The lower end keeps the grabber model matrix from becoming degenerate (`normal_matrix` takes an inverse of it), the upper end keeps the gizmo AABB from inflating too much, because it is merged into `_max_bounding_box()` and ends up in `apply_projection()` again.

`Selection::get_screen_scalling_matrix()` gets the same factor. It sizes the sidebar position and rotation hints under the same flag, from the same anchor, and they are drawn one line before the gizmo in the same frame, so leaving it out would make the hint arrows and the gizmo arrows differ in length by the same factor.

Because of that the update moved from `_render_current_gizmo()` up into `render()`, right after `apply_projection()` fills in `near_z` and `gui_scale`. Both consumers now read the value of the current frame. `_render_current_gizmo()` is the only call site of the old assignment, so nothing else loses the refresh.

Existing preferences keep their meaning, `grabber_size_factor` still scales the same way.

## Known limits of this patch

- The correction uses one anchor per frame, the selection center. Grabbers that sit away from it, the Scale corners, the Cut plane grabber, the Text/SVG cube, keep a residual error proportional to their own depth offset. One anchor keeps the parts of a gizmo proportional to each other, which looked more important than making each grabber exact.
- With `gizmo_keep_screen_size` off nothing changes at all. That path is also foreshortened, but there the grabber size and the arrow offset both scale as `1/zoom` and stay consistent with each other, so correcting only one of them would make it worse.
- The gizmo geometry used by the picking pass is still built in `on_render()`, which runs after `_picking_pass()`, so hit boxes lag one frame as before. Both passes use the same world sizes, so this patch does not change that, but it may be part of why grabbers are hard to hit while the camera moves (#6806).

### DIFF
--- a/src/slic3r/GUI/GLCanvas3D.cpp
+++ b/src/slic3r/GUI/GLCanvas3D.cpp
@@ -2807,6 +2807,9 @@ void GLCanvas3D::render(bool only_init)
     camera.apply_projection(_max_bounding_box(true, true, true,is_volumes_limit_to_expand_plate()));
     camera.update_frustum();
 
+    //BBS update gizmo camera scaling (inv_zoom + perspective depth correction)
+    GLGizmoBase::update_camera_scaling(camera, m_selection);
+
     m_frame_callback_list.clear();
 
     const std::array<int, 4>& viewport = camera.get_viewport();
@@ -9428,8 +9431,6 @@ void GLCanvas3D::_render_volumes_for_picking() const
 
 void GLCanvas3D::_render_current_gizmo() const
 {
-    //BBS update inv_zoom
-    GLGizmoBase::INV_ZOOM = (float)get_active_camera().get_inv_zoom();
     m_gizmos.render_current_gizmo();
 }
 
--- a/src/slic3r/GUI/Gizmos/GLGizmoBase.cpp
+++ b/src/slic3r/GUI/Gizmos/GLGizmoBase.cpp
@@ -5,6 +5,7 @@
 #include <GL/glew.h>
 #include <algorithm>
 
+#include "slic3r/GUI/Camera.hpp"
 #include "slic3r/GUI/GUI_App.hpp"
 #include "slic3r/GUI/GUI_Colors.hpp"
 #include "slic3r/GUI/OpenGLManager.hpp"
@@ -15,6 +16,7 @@ namespace Slic3r {
 namespace GUI {
 
 float GLGizmoBase::INV_ZOOM = 1.0f;
+float GLGizmoBase::DEPTH_CORRECTION = 1.0f;
 
 
 const float GLGizmoBase::Grabber::SizeFactor = 0.05f;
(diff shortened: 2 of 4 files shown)


Haidiye00 (commented):
Hi, please record the GIFs before and after the changes, and then submit them. Thanks.

guanyun-gudujian (changes requested):
Currently, we already have a constant gizmo screen size. Although there are some artifacts, it is good enough for now.

Issue #3566, opened 2024-02-09
Select infill patern for bridge surface


M-Spit:
**Is your feature request related to a problem? Please describe.**
The bridge infill pattern is always rectalinniar, which is fine when you print square parts.
However, I try to print a round part (a cillinder) with a widening of said cilinder in layer 150. Instead of the rectaliniar infill, I'd like to use concentric or even archymedian spiral patterns.
Especially when trying to print waterproof parts, concentric is worse to deal with (even with a sealant, in certain directions some gaps between the lines will leak)
It will also add to printing speeds for these layers.

**Describe the solution you'd like**
I'd  like a setting in the bridge settings where you can choose the surface infill pattern of a bridge.

**Describe alternatives you've considered**
Manually ironing the surface of the bridge, which is labor intensive.

**Additional context**
![afbeelding](https://github.com/bambulab/BambuStudio/assets/28558448/6bb26429-f5e4-4dcb-a48a-40bb14337e24)
The blue bridge would be a lot nicer when printed as a concentric infill or a spiral.
https://www.reddit.com/r/BambuLab/comments/15tfkrb/how_to_change_the_bridge_pattern/
A reddit post describing the same issue

Thanks in advance!

BotSpace:
I second this, would be very helpful for my project.

riccalisci:
please include this feature! i am sure it will be very helpful

n3dwards:
**Throwing my my support for this here, along with some context/suggestions for how this would work.**

I've just started using PLA to support PETG and it works so well that I can now add supports for floating flat surfaces that have zero z-distance between them and the support material. This functionally allows me to have cleanly printed bottom surfaces that are not directly on the build plate. However, I have no options for configuring the pattern since these lines are considered "bridge" and not "bottom surface".

I presume the bridge line type doesn't support patterns because normally any pattern other than linear would not really bridge properly in a normal bridging scenario. However, with 100% support underneath (or close to it), they are not actually bridging any gaps at all.

**My suggestion:** change this so that if "bridge" lines are fully supported, the slicer will stop treating them as bridges and switch to "bottom surface", which would allow for pattern selection.

chadwallacehart:
+1 for this feature. I also have a project where I could also use a non-linear pattern for bridges

PLAmenPRINT:
+1

hoyeungw:
+1

cisbrh:
+1

PR #11964, opened 2026-08-20
ENH: add bridge_bottom_surface_pattern process setting

bandesz:
## Summary
- Adds a process setting to choose the infill pattern for the **underside of bridges** (overhangs printed over support or air).
- Pattern list matches top/bottom surface infill (Concentric, Rectilinear, Monotonic, etc.), with **Default** first.
- Default uses monotonic when the top surface is monotonic, otherwise rectilinear — the same fallback as before this setting existed.
- Addresses https://github.com/bambulab/BambuStudio/issues/3566.

The option lives under **Strength → Top/bottom shells**, is Advanced-mode, and is also available as a per-object/per-part override.

## Why
Bridge infill is currently hardcoded to rectilinear (or monotonic only when the top surface pattern is monotonic). That works for rectangular overhangs, but round or fully supported bottoms (for example PLA supporting PETG with zero Z gap) benefit from concentric or other solid patterns.

## Behavior
- `Default` is setting behaviour, not a fill algorithm. It uses a dedicated `BridgeBottomSurfacePattern` enum so it is not on the shared `InfillPattern` map.
- First-layer bottoms still use `bottom_surface_pattern` (`is_bridge` is only true for `layer.id() > 0`).
- Non-first-layer `stBottomBridge` surfaces use `bridge_bottom_surface_pattern`.
- Internal bridges (`stInternalBridge`) are unchanged.
- Bridge flow is still used regardless of the selected pattern.
- Changing the setting invalidates infill only (`posInfill`), same as other surface patterns.
- `bottom_surface_pattern` tooltip still notes that it is not bridge infill.
- New strings are marked with `L()`; language catalogs are left to the existing translation CI.

## Test plan
- [x] Open Process settings → Strength → Top/bottom shells in Advanced mode and confirm **Bridge bottom surface pattern** appears next to Bottom surface pattern, with **Default** first, then the same pattern list.
- [x] Slice a model with a first-layer bottom: pattern still follows Bottom surface pattern, not the new setting.
- [x] Slice a model with an overhang/bridge (layer > 0) and **Default**: infill matches previous behaviour (monotonic if top surface is monotonic, otherwise rectilinear).
- [x] Change Bridge bottom surface pattern to Concentric / Monotonic / Hilbert and confirm preview infill updates without a full reslice.
- [x] Confirm per-object/per-part override in the object settings Strength section.
- [x] Load a BBL system process profile and confirm the default is Default (JSON key `default`).
- [x] Change the setting, switch presets, and confirm Unsaved Changes shows the human-readable pattern name.

## Screenshots

Layer before the overhang, white layers are the support/raft interface (interface pattern set to concentric).

Screenshot 2026-08-21 at 08 45 28

Layer showing the bottom of the overhang, top object has bridge bottom surface pattern left as "Default", bottom object has it set to "Concentric"

Screenshot 2026-08-21 at 08 45 39

### DIFF
--- a/resources/profiles/BBL/process/fdm_process_common.json
+++ b/resources/profiles/BBL/process/fdm_process_common.json
@@ -8,6 +8,7 @@
     "bottom_color_penetration_layers": "3",
     "bottom_shell_thickness": "0",
     "bottom_surface_pattern": "monotonic",
+    "bridge_bottom_surface_pattern": "default",
     "bottom_surface_density": "100",
     "bridge_flow": "0.95",
     "bridge_no_support": "0",
--- a/src/libslic3r/Fill/Fill.cpp
+++ b/src/libslic3r/Fill/Fill.cpp
@@ -20,6 +20,22 @@
 
 namespace Slic3r {
 
+static InfillPattern infill_for_bridge_bottom(BridgeBottomSurfacePattern pattern, InfillPattern top_surface_pattern)
+{
+    switch (pattern) {
+    case bbspDefault:            return top_surface_pattern == ipMonotonic ? ipMonotonic : ipRectilinear;
+    case bbspConcentric:         return ipConcentric;
+    case bbspRectilinear:        return ipRectilinear;
+    case bbspMonotonic:          return ipMonotonic;
+    case bbspMonotonicLine:      return ipMonotonicLine;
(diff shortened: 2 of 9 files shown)


bandesz:
cc @SaltWei

PR #11965, opened 2026-08-20
fix: Restore layer Z after spiral-vase raft tool changes

### LINKED ISSUE #11961: Floating extrusion when using raft and spiral vase mode with multi filament printing with X2D
### Type of issue

### Bambu Studio Version

2.6.0.51

Bambu Lab Official website

### OS version

Windows 11

_No response_

### Printer

X2D

### How to reproduce

1. Insert cylinder model with hollow core into slicer
2. Enable spiral vase mode and set raft layers to 2
3. Set main filament to main nozzle, support/raft to aux nozzle
4. Slice

### Actual results

The first bottom layer filament will be floating about 10 layers from where it should be

### Expected results

The first bottom layer should be on the bottom, not higher than the current layer

v

- [ ] Project file (always)
- [ ] Log file (always)

mvanhorn:
## Summary

Update the spiral-vase path in `GCode::travel_to` to restore the writer to the requested/nominal layer Z when an earlier tool change or injected G-code left it at a different height, then retain the existing XY-only travel behavior once Z is correct. Base the decision on the generator's tracked position and the travel's target Z rather than on X2D-specific commands, raft counts, or a hard-coded clearance value, so other custom tool-change sequences receive the same correction. Preserve the no-extra-Z-move behavior for ordinary spiral layers where the tracked and nominal heights already agree, avoiding changes to seam blending and spiral interpolation.

## Test plan

- Happy path: slice a hollow single-object spiral-vase print with two raft layers on filament 2 and the model on filament 1, with tool-change G-code ending at `max_layer_z + 3`; verify the first model extrusion after the raft is at the nominal first-object-layer Z.
- Edge case: slice the same spiral-vase/raft geometry without a Z-changing tool change; verify no redundant layer-height correction is introduced and the existing spiralized Z progression remains continuous.
- Error path: use custom tool-change G-code that moves to a clearance Z and leaves the current XY position unknown; verify the generated travel safely restores Z before any model extrusion and does not emit extrusion at the clearance height.

## Why

On an X2D print that uses an auxiliary filament for a two-layer raft and the main filament for a spiral-vase object, the first object extrusion is emitted roughly ten layers above its intended position. The raft-to-object filament change runs custom machine G-code that may finish at a clearance Z such as `max_layer_z + 3`, and the generator records that external Z position. The spiral-vase branch of `GCode::travel_to` then emits only an XY travel, unlike the normal branch that travels in XYZ, so it does not restore the nominal object-layer Z before extrusion. The issue is limited to the interaction between raft-driven tool changes, externally moved Z, and spiral-vase travel behavior; ordinary spiral interpolation and the X2D profile template should remain unchanged.

Closes #11961

### DIFF
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -7722,11 +7722,11 @@ std::string GCode::travel_to(const Point &point, ExtrusionRole role, std::string
 
         if (m_spiral_vase) {
             // No lazy z lift for spiral vase mode
+            for (size_t i = 1; i < travel.size(); ++i)
+                gcode += m_writer.travel_to_xy(this->point_to_gcode(travel.points[i]), comment, use_short_travel_accel);
             const double target_z = z == DBL_MAX ? m_nominal_z : z;
             if (std::abs(m_writer.get_position().z() - target_z) > EPSILON)
                 gcode += m_writer.travel_to_z(target_z, "restore spiral vase layer Z");
-            for (size_t i = 1; i < travel.size(); ++i)
-                gcode += m_writer.travel_to_xy(this->point_to_gcode(travel.points[i]), comment, use_short_travel_accel);
         } else {
             if (travel.size() == 2) {
                 // No extra movements emitted by avoid_crossing_perimeters, simply move to the end point with z change
--- a/src/libslic3r/GCode.cpp
+++ b/src/libslic3r/GCode.cpp
@@ -7722,6 +7722,9 @@ std::string GCode::travel_to(const Point &point, ExtrusionRole role, std::string
 
         if (m_spiral_vase) {
             // No lazy z lift for spiral vase mode
+            const double target_z = z == DBL_MAX ? m_nominal_z : z;
+            if (std::abs(m_writer.get_position().z() - target_z) > EPSILON)
+                gcode += m_writer.travel_to_z(target_z, "restore spiral vase layer Z");
             for (size_t i = 1; i < travel.size(); ++i)
                 gcode += m_writer.travel_to_xy(this->point_to_gcode(travel.points[i]), comment, use_short_travel_accel);
         } else {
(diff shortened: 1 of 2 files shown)


src/libslic3r/GCode.cpp
```diff
@@ -7722,6 +7722,9 @@ std::string GCode::travel_to(const Point &point, ExtrusionRole role, std::string
 
         if (m_spiral_vase) {
             // No lazy z lift for spiral vase mode
+            const double target_z = z == DBL_MAX ? m_nominal_z : z;
+            if (std::abs(m_writer.get_position().z() - target_z) > EPSILON)
```
Diatom-Bambu, line 7728: Here's a small suggestion: the action of descending the z-height should more reasonably be placed after the two lines (xy travel)
_"for (size_t i = 1; i < travel.size(); ++i)
gcode += m_writer.travel_to_xy(this->point_to_gcode(travel.points[i]), comment, use_short_travel_accel);"
（line 7728 below）
This is to ensure that the xy travel is performed at a higher height, and printing starts directly after descending, reducing the possibility of being scratched by the warped edges.

Diatom-Bambu:
Hello, thank you for your contribution.
The problem indeed need a fix, but the action of descending the z-height should more reasonably be placed after the xy travel, which make it far from scratch.

mvanhorn:
You were right about the ordering, and it is fixed in 5ebea7f.

The XY travel loop now runs before the target-Z calculation and the Z restoration, so the toolhead finishes its horizontal move and only then descends, instead of dropping to the layer Z first and travelling at that height.

It is a two-line reorder in `travel_to`; nothing else changed. clang-format reports no violations on the file. I could not run the full build here since it needs prebuilt dependency prefixes I do not have locally, so the CI build is the real check on it.

Diatom-Bambu (approved):
thanks for your fixing

zhimin-zeng-bambulab (approved):

mvanhorn:
Thanks for merging, @lanewei120. Restoring the nominal layer Z after a tool change should keep spiral-vase rafts from travelling at the wrong height.

PR #11988, opened 2026-08-21
FIX: Reload from disk fails when 3MF stores source as filename only

ClusterM:
## Summary

- **Reload from disk** often showed **Unable to reload** for models imported from a 3MF (for example a FreeCAD export), even when the file on disk loaded fine.
- 3MF projects store `source.input_file` as a **filename only** (`tpu_tube-Roller.3mf`) by design (`FullPathSources` is off so shared projects do not leak full paths). After the same file is read again, the new volume gets the **absolute path**. The matcher used `==` on those strings, so a valid reload was rejected.
- The name fallback never ran in this case either: it required the part name to equal the source **filename** (`Roller` vs `tpu_tube-Roller.3mf`). The 3MF object name and the part name on the plate already matched (`Roller`).

## Change

In `Plater::priv::reload_from_disk()`:

- Compare source files by **filename** (case-insensitive), not by the full stored string.
- If the stored object/volume slot does not match, search the just-loaded model by **part name**. Do not require that name to be the source filename.


Haidiye00 (commented):
Hello, please provide the GIF tests before and after the modification. Thank you

ClusterM:
> Hello, please provide the GIF tests before and after the modification. Thank you

Before the fix:

  
  
After the fix:

Haidiye00 (commented):
approve

ClusterM:
> approve

@Haidiye00 so why it's not merged? Shouldn't you press "approve" button?

lanewei120:
@ClusterM thanks~

Just some comment from gpt:
Comparing only filename() may introduce false matches when both sides contain valid full paths but point to different files with the same filename.

Could we keep the full-path comparison when both paths contain directory information, and only fall back to filename comparison when one side is filename-only? That should fix the 3MF case without weakening the existing matching semantics for normal files.

ClusterM:
@lanewei120 good point, thanks.

Updated `source_input_files_match()` accordingly:

- if **both** sides have a directory, keep the original full-path `==` (`/a/model.stl` ≠ `/b/model.stl`);
- if **either** side is filename-only (typical 3MF `source.input_file`), fall back to a case-insensitive filename compare.

That still fixes the FreeCAD/3MF reload case without changing the match for normal files that store a full path. Pushed to this branch (squashed).

PR #12009, opened 2026-08-23
add: initial greek translation

thanbimp:
Add initial greek translation for bambu studio UI.


tonghao-bbl:
Hi @thanbimp we've added greek and other four languages in this sprint, which is expected to be shipped in the next release.

I'm very sorry that we didn't share our roadmap with the community earlier, which led to duplicated work. We really appreciate the effort you put in. Unfortunately, we have to close this PR.

Issue #2249, opened 2023-08-10
Align Model Surface to Build Plate


zeffez:
The current "Lay on Face" alignment tool appears to provide alignment planes that are surfaces of a bounding polyhedron of the model. It would be helpful if we could select faces of the actual model to align with the build plate (similar to other slicers such as Cura). This would allow the user to align model faces with slicing planes, and then critically flat surfaces (that are not necessarily parallel to the model bottom or bounding polyhedron) could be printed as one continuous layer without layer ridges.

My current work around involves realigning the model in a 3rd party software (such as Cura), exporting a new stl file with the adjusted alignment, and then import into Bambu studio. Although this works, these additional steps to the workflow get tedious pretty quickly.

Thanks!

johnflux:
@lanewei120 Here's an example:
I have this mostly flat piece:
<img width="774" alt="image" src="https://github.com/bambulab/BambuStudio/assets/1230500/ee53576a-6a68-49ca-85ca-0e65320d4b5b">

But it lays out with the top surface not flat.  So the result is steps in the top surface:

<img width="772" alt="image" src="https://github.com/bambulab/BambuStudio/assets/1230500/4f630f64-017d-4fae-8485-6e0d3c8fdd05">

I would like to add support to prop it up a bit so that the top surface is flat

lanewei120:
> @lanewei120 Here's an example: I have this mostly flat piece: <img alt="image" width="774" src="https://user-images.githubusercontent.com/1230500/275341837-ee53576a-6a68-49ca-85ca-0e65320d4b5b.png">
> 
> But it lays out with the top surface not flat. So the result is steps in the top surface:
> 
> <img alt="image" width="772" src="https://user-images.githubusercontent.com/1230500/275341813-4f630f64-017d-4fae-8485-6e0d3c8fdd05.png">
> I would like to add support to prop it up a bit so that the top surface is flat

@johnflux 
thanks

is it possible to share this stl for us to have a check?

johnflux:
Here. Rename to just .stl
[Fore_Shield.stl.txt](https://github.com/bambulab/BambuStudio/files/12912112/Fore_Shield.stl.txt)

Lex-FX:
+1 For also being able to select a specific mesh facet/surface to orientate down. Up til now, I've mostly been working with geometric objects so being able to orientate to a projected plane is invaluable. (like having 3 specific corners touch the bed)

[...]

PR #12048, opened 2026-08-27
ENH: cut gizmo - set the cut plane from a clicked triangular facet

MatRanc:
Pick face demo

### Background

In #2249 the request was to align a model to the plate using an actual facet rather than a bounding-polyhedron face, and *Lay on Face* gained its **Triangular facet** mode as a result. In that thread @Haidiye00 wrote:

> You've given us a great idea, we'll consider it, thank you very much.

The Cut tool never received the equivalent, and it has the same problem — arguably a worse one, because a cut usually needs to be *precise*.

### Problem

The cut plane can only be positioned by dragging the grabbers or typing Euler angles into the Rotation fields. Neither is workable when the cut has to be taken relative to a specific surface of the model — a chamfer, an angled boss, a draft face. Today you either eyeball the rotation or realign the model in another tool first.

### What this adds

Press **Pick face** in the cut panel, then click a facet. The cut plane becomes coplanar with that facet, oriented along its normal.

The plane lands *flush* with the face, so the cut is degenerate at that instant — that is intended. The existing **Movement** field then supplies a real offset, which is what makes this useful for precise work: "3 mm in from this face". Orientation-only placement would throw away the positional reference that motivates the feature.

Everything downstream — clipper, connectors, the Rotation/Movement fields, undo — derives from `m_plane_center` and `m_rotate_matrix` and needed no changes.

### Implementation notes

- The rotation is built the same way `process_cut_line()` builds it, with the normal coming from a facet instead of a screen-space line.

- **The validity check is deliberately not `process_cut_line()`'s containment guard.** That one tests `m.inverse() * (plane_center - instance_offset) + new_tbb.center()`, which is not the plane's position in that frame — `transformed_bounding_box()` maps a point to `R⁻¹·(p_world − plane_center)`, so the plane sits at the origin. The terms cancel in X and Y because an instance origin sits at the model's bounding-box centre in those axes, but in Z it sits at the object's base, leaving the object's half-height as error. `process_cut_line()` hides this by always cutting through `m_bb_center`. A plane flush with a facet does not, so reusing that guard silently dropped clicks on cut pieces. This uses the test `set_center_pos()` itself applies, so the guard and the function it guards ask the same question. Flagging it in case the same latent issue matters elsewhere.

- `on_get_requirements()` now also requests `CommonGizmosDataID::Raycaster`.

- `MeshRaycaster::unproject_on_mesh` returns the hit normal in **mesh-local** coordinates while the cut gizmo works in world space, so the normal is transformed with the inverse-transpose of the linear part. Using the linear part directly leaves the plane visibly non-perpendicular to the picked facet on a non-uniformly scaled instance.

- The facet index is captured **inside** the nearest-hit branch. The equivalent loop in `GLGizmoFlatten::update_raycast_cache` captures it *after* the loop, so on a multi-volume object it can index the last-tested volume's facet into the nearest volume's mesh. I have left `GLGizmoFlatten` untouched to keep this PR scoped to the cut gizmo — happy to send that as a separate fix if you'd like.

- Clicks are deliberately **not** gated on `m_hover_id`. The cut plane is a large translucent quad and the wanted face is usually behind it; gating on hover made exactly those faces unclickable. Since the mode is armed by an explicit button press, a click while armed can only mean "pick that facet". `gizmo_event` runs before the manager's grabber handling, so returning `true` suppresses the drag rather than racing it.

### Scope

Planar cut mode only. The button is disabled in Dovetail / tongue-and-groove mode, which has its own plane-editing state machine and would need separate reasoning.

Three files changed: `GLGizmoAdvancedCut.{cpp,hpp}` plus two new strings in `bbl/i18n/BambuStudio.pot`. No changes to `GLGizmoFlatten`, `GLGizmosManager`, or the build system.

### Known limitations

- **Mesh quality.** Cutting exactly along a facet plane can leave slivers or non-manifold edges where the plane is coplanar with, or grazes, existing geometry. This is inherent to cutting triangle meshes and affects the existing drag-the-plane cut equally — aiming at real geometry just makes it easier to land on such a plane. No change to the cutting algorithm itself is included here.
- **One facet, no coplanar snapping.** Picking uses the single triangle under the cursor. On a flat face that is exact; on a curved or finely tessellated surface you get that one triangle's plane, which may not be what you want.
- **One-shot by design.** The mode disarms after a successful pick; press the button again to pick another face.

### Testing

Built and manually tested by the author. The `Build all` run on this PR is sitting at `action_required` pending maintainer approval, as fork PRs do — happy to act on whatever it reports once someone kicks it off.

| | |
|---|---|
| Base | `926a719` (`02.08.02.61`) |
| OS | macOS 26.5.2 (build 25F84) |
| Toolchain | Xcode 26.5 (17F42), CMake 3.31.12 |
| Target | arm64, `Release`, Xcode generator via `BuildMac.sh` |

Cases exercised:

- Cube, top face — plane lands flush, Rotation reads `0, 0, 0`, **Movement** offsets into the body and the preview cuts correctly.
- Wedge, 45° face — Rotation reads the true facet angle rather than a bounding-box angle.
- **Non-uniformly scaled instance** (Z scaled to 200%) — plane stays perpendicular to the highlighted facet. This is the case that catches the normal-transform error described above.
- **Rotated instance** (30° about Z) — plane matches the face in world space.
- **Cut pieces** — repeatedly picking faces on the bodies produced by a previous cut. This is what surfaced the containment-guard problem above; it was diagnosed from a trace of a failing pick and fixed before submitting.
- Picking a face *behind* the translucent cut plane.
- Undo/redo — one `Cmd+Z` restores the previous plane exactly; redo reapplies.
- Connectors — with connectors placed, picking a new face moves them onto the new plane.
- Clicking empty space while armed changes nothing and pushes no undo step.

Not exercised: the multi-volume case, and Windows/Linux builds — the change is platform-independent but I only have a Mac to verify on.

### DIFF
--- a/bbl/i18n/BambuStudio.pot
+++ b/bbl/i18n/BambuStudio.pot
@@ -968,6 +968,15 @@ msgstr ""
 msgid "Flip cut plane"
 msgstr ""
 
+msgid "Pick face"
+msgstr ""
+
+msgid "Click a face of the model."
+msgstr ""
+
+msgid "Click a face of the model to set the cut plane. The plane lands flush with that face; use Movement to offset it."
+msgstr ""
+
 msgid "Planar"
 msgstr ""
 
--- a/bbl/i18n/BambuStudio.pot
+++ b/bbl/i18n/BambuStudio.pot
@@ -968,15 +968,6 @@ msgstr ""
 msgid "Flip cut plane"
 msgstr ""
 
-msgid "Pick face"
-msgstr ""
-
-msgid "Click a face of the model."
-msgstr ""
-
-msgid "Click a face of the model to set the cut plane. The plane lands flush with that face; use Movement to offset it."
-msgstr ""
-
 msgid "Planar"
 msgstr ""
 
--- a/src/slic3r/CMakeLists.txt
+++ b/src/slic3r/CMakeLists.txt
@@ -138,6 +138,8 @@ set(SLIC3R_GUI_SOURCES
     GUI/Gizmos/GLGizmosManager.hpp
     GUI/Gizmos/GLGizmosCommon.cpp
     GUI/Gizmos/GLGizmosCommon.hpp
+    GUI/Gizmos/FacetPicker.cpp
+    GUI/Gizmos/FacetPicker.hpp
     GUI/Gizmos/GLGizmoBase.cpp
     GUI/Gizmos/GLGizmoBase.hpp
     GUI/Gizmos/GLGizmoMove.cpp
--- a/src/slic3r/GUI/Gizmos/FacetPicker.cpp
+++ b/src/slic3r/GUI/Gizmos/FacetPicker.cpp
@@ -1,0 +1,158 @@
+#include "FacetPicker.hpp"
+
+#include "GLGizmoBase.hpp"
+#include "GLGizmosCommon.hpp"
+#include "slic3r/GUI/Camera.hpp"
+#include "slic3r/GUI/GUI_App.hpp"
+#include "slic3r/GUI/Selection.hpp"
+#include "libslic3r/Model.hpp"
+
+#include <GL/glew.h>
+
+#include <limits>
+
+namespace Slic3r {
+namespace GUI {
+
+Vec3d FacetPicker::world_normal(const Transform3d &trafo, const Vec3f &mesh_normal)
+{
+    const Vec3d  n   = trafo.linear().inverse().transpose() * mesh_normal.cast<double>();
+    const double len = n.norm();
+    return (len < EPSILON) ? Vec3d::Zero() : Vec3d(n / len);
+}
+
+void FacetPicker::set_active(bool active)
+{
+    if (m_active == active)
+        return;
+    m_active = active;
+    if (!m_active)
+        reset();
+}
+
+void FacetPicker::reset()
+{
+    m_hit            = Hit();
+    m_rendered_facet = -1;
+    m_highlight.reset();
+}
+
+bool FacetPicker::update(const Vec2d &mouse_position, const CommonGizmosDataPool *pool, const Selection &selection, const Camera &camera)
+{
+    m_hit = Hit();
+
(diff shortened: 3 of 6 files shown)


Haidiye00 (commented):
good job,Can the newly added code be encapsulated into a class for easier maintenance in the future? Thank you

MatRanc:
Encapsulated in `FacetPicker` (`src/slic3r/GUI/Gizmos/FacetPicker.{hpp,cpp}`). `GLGizmoAdvancedCut` goes from 8 members + 4 methods down to a single `FacetPicker` member plus `apply_picked_facet()`.

The class answers only "which triangle is under the cursor, and draw it highlighted" — it knows nothing about cut planes, so turning a facet into a plane stays in the gizmo. That keeps it reusable: `GLGizmoFlatten`'s "Triangular facet" mode already performs the same raycast-and-highlight, but written directly inside that gizmo, so it could be moved onto `FacetPicker` to remove the duplication. I've deliberately left that out of this PR to avoid changing a working gizmo here — happy to do it as a follow-up if you'd like.

Also corrected a `class`/`struct` mismatch on the `Camera` forward declaration that could have caused linker errors under the MSVC ABI.

Pushed as a separate commit rather than a force-push, so the change against your review stays visible. Re-verified by hand on macOS, including repeated picks on the pieces produced by a previous cut.

Haidiye00 (approved):
approve

lanewei120:
hello @MatRanc 
thanks for this patch, could you remove the modification of bbl/i18n/BambuStudio.pot
this file will be updated automatically

MatRanc:
Removed in b7877ecf6276ae993acea8f0e34fe506c3e23198 — `bbl/i18n/BambuStudio.pot` is back to its `master` contents and no longer appears in the diff.

The `_L()` calls in the gizmo are unchanged, so the three new strings will be picked up whenever the template is regenerated.

lanewei120:
thanks @MatRanc

Issue #6362, opened 2025-03-31
Improved Rich Prescence Support and Discord Integration


H34TW4V3:
**1. Improved Rich Presence Support**

- Dynamic Thumbnails: Displaying the current model open in Bambu Studio as the Rich Presence thumbnail would give a more personalized and visually informative experience. Friends can see the exact model being worked on, sparking conversations and collaboration.

- Enhanced Detail: Alongside the model thumbnail, including details like the print progress, estimated time remaining, and printer status would make the presence more comprehensive and appealing.

**2. Discord Integration**

- Seamless Sharing: Allow users to share print status or completed prints directly with friends on Discord. This could include the model thumbnail, print progress, and key details such as print time and filament used.

- Real-Time Updates: Friends could receive notifications when a print starts, completes, or encounters an issue. This fosters community interaction and helps share accomplishments effortlessly.

-  Social Engagement: Integrating a "Show Off" button would allow users to highlight their prints directly in a Discord channel, promoting creativity and discussion within the 3D printing community.

PR #12059, opened 2026-08-28
ENH: add Discord Rich Presence support

Sn3akySqu1d:
Optionally publish current activity to the user's Discord profile: the open project, slicing progress, and live print progress with a completion countdown when a printer is connected.

Off by default and opt-in from Preferences, alongside a second toggle that hides project and print job names while keeping progress and status. Only the displayed strings, a printer model name and timestamps leave the machine; no geometry, thumbnails, file paths, G-code or account details. The application talks to a local unix socket or named pipe and opens no network connection.

No new third-party dependency. Discord's local IPC is a length-prefixed JSON protocol, implemented directly rather than vendoring discord-rpc into deps/. JSON uses the bundled nlohmann/json.

Socket work runs on one worker thread; the GUI thread only stores a snapshot under a mutex. Every wait carries a deadline, so an unresponsive Discord cannot stall the UI or delay shutdown. Discord renders both the elapsed and countdown timers itself from absolute timestamps, so frames are sent only when the state actually changes, rate limited to one per five seconds.

Art assets are rasterised from existing icons in resources/images and committed under resources/images/discord for upload to the Discord developer portal. Nothing loads them at runtime.

DISCORD_APPLICATION_ID in DiscordPresence.cpp points at a contributor-registered application so the feature can be evaluated as-is. Rich Presence requires an application id, and the owning account controls the name and artwork every user sees, so it should be re-pointed at an application owned by Bambu Lab before release.

Partially addresses #6362. Dynamic model thumbnails and sharing prints into Discord channels are deliberately out of scope: Rich Presence cannot accept image bytes from the application, and channel posting would require a bot identity, an OAuth2 flow and server-side credentials.


src/slic3r/Utils/DiscordPresence.cpp
```diff
@@ -0,0 +1,337 @@
...
+#else
+    #include <unistd.h>
+#endif
+
+namespace Slic3r {
+
+// Contributor-registered application, so the feature can be evaluated as-is.
+// The owning account controls the name and artwork every user sees, so this
+// should be re-pointed at an application owned by Bambu Lab before release.
+static const char *DISCORD_APPLICATION_ID = "1542693447339745472";
```
tonghao-bbl: we can't use this hardcoded application clident id, maybe a configurable string?

src/slic3r/Utils/DiscordIPC.cpp
```diff
@@ -0,0 +1,386 @@
...
+    return true;
+}
+
+bool DiscordIPC::write_all(const char *src, size_t count)
+{
+    size_t written   = 0;
+    int    budget_ms = 2000;
+
+    while (written < count) {
+        const ssize_t n = ::write(m_fd, src + written, count - written);
```
tonghao-bbl: we should disable or hanle `SIGPIPE` for write operation

src/slic3r/Utils/DiscordPresence.cpp
Sn3akySqu1d: no id in the source now. added SLIC3R_DISCORD_APP_ID (CMake) which sets the default, discord_rich_presence_app_id in the config overrides it, so you can retarget it without a rebuild. Feature is disabled if neither is set.

-DSLIC3R_DISCORD_APP_ID=1542693447339745472 a test app I set up with the assets from resources/images/discord/.

src/slic3r/Utils/DiscordIPC.cpp
Sn3akySqu1d: Fixed per socket without changing the global handler. MSG_NOSIGNAL on Linux and SO_NOSIGPIPE on macOS. The EPIPE was already handled as a write failure.

src/slic3r/Utils/DiscordIPC.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    m_pipe = pipe;
+    m_open = true;
+    return true;
+}
+
+bool DiscordIPC::write_all(const char *src, size_t count)
+{
+    size_t written = 0;
+    while (written < count) {
```
tonghao-bbl, line 177: can we have a 2000ms timeout like on linux?

Sn3akySqu1d: done, pipe is now overlapped so we get the same 2000ms as linux on write. reads also moved to overlapped.

Issue #12078, opened 2026-08-29
Custom-filament spools are permanently "Unsupported": AddEditDialog writes a preset setting_id into a field matched against filament_id


marciogranzotto:
**Version:** 02.08.02.61 (Windows)

## Summary

Spools added via Filament Manager for a **custom filament** (any vendor not in Bambu's global
cloud catalogue) are written with an id from the wrong namespace, so they can never resolve and
are shown under **Unsupported Filaments** forever.

On my install this affected 13 of 35 spools.

## Root cause

The reader accepts only a `filament_id`. `PresetBundle::get_filament_by_filament_id`
(`src/libslic3r/PresetBundle.cpp`) matches:

```cpp
if (filament_preset.filament_id == filament_id) { ... }
```

and `AMSMaterialsSetting.cpp:1557` feeds it the spool's `setting_id`:

```cpp
bool has_preset = bundle && bundle->get_filament_by_filament_id(sp->setting_id).has_value();
if (has_preset) { lib_brand_to_spools[brand].push_back(*sp); }
else            { unsupported_spools.push_back(*sp); }
```

The writer picks that value with this precedence
(`.../device_page/src/features/filament-manager/AddEditDialog.tsx:1092`, mirrored at `:528`):

```js
setting_id: matchedCloudFilamentId
  || matchedPresetItem?.setting_id     // <-- wrong namespace
  || matchedPresetItem?.filament_id
  || initSpool?.setting_id
  || '',
```

`matchedCloudFilamentId` resolves only on an exact match against the **global** cloud
catalogue:

```js
settings.find((it) => it.filamentVendor === brand && getCloudSettingDisplayName(it) === name)
```

A user's own custom filament is never in that catalogue, so the first term is empty and the
code falls through to `matchedPresetItem?.setting_id`. Preset `setting_id`s are `PFUS` + 14 hex,
while `filament_id`s are `P` + 7 hex (or `GFxxx`) — so the stored value can never equal any
preset's `filament_id`.

Concretely, from the `config/fetched` payload on my machine the catalogue has **99 entries**:
Bambu Lab (49), Generic (29), Polymaker (11), SUNLU (7), Overture (2), eSUN (1 — `PLA+` only).
Zero Voolt3D.

## Evidence

Every broken spool held a `PFUS…` value that is the `setting_id` of the very preset it should
have pointed at:

| filament | stored | should be |
|---|---|---|
[...]

Issue #11937, opened 2026-08-18
Filament Manager / Third-party filaments cannot be added to the AMS from the Filament Manager


Spocky510:
### Type of issue

Other (functional bug, wrong result, UI glitch, etc.)

### Bambu Studio Version

2.8.2.60

### Where is the application from?

Bambu Lab GitHub releases, Bambu Lab Official website

### OS version

Win 11

### Additional system information

CPU: AMD Ryzen 7 5800 X3D
RAM: 64GB
GPU: RTX 4060ti

### Printer

P2S

### How to reproduce

1: Open Bambu Studio
2: Manually add third-party filaments to the Filament Manager.
3: Added filament is displayed in the Filament Manager.
4: When assigned in the AMS, the third-party filament is displayed as "unsupported filament."
5: The third-party filament cannot be selected during AMS assignment. Original Bambu filament, however, does
6: Third-party filament with multiple colors cannot be changed.
7: The filament quantity is not updated after printing is complete even though updated

### Actual results

<img width="2515" height="783" alt="Image" src="https://github.com/user-attachments/assets/17dbc037-1571-4111-a7fc-8902cf4a5f43" />

<img width="647" height="610" alt="Image" src="https://github.com/user-attachments/assets/78de46be-dc1d-4578-9bac-9a74a82a2a48" />

<img width="645" height="601" alt="Image" src="https://github.com/user-attachments/assets/a6451697-b6db-4725-bfd9-dfa326d66749" />

### Expected results

It should also be possible to select and insert unsupported filaments. However, this does not happen.

### File uploads (project / log / crash dump / performance profile)

[log_iotc.txt](https://github.com/user-attachments/files/31175971/log_iotc.txt)

[stand with backing.zip](https://github.com/user-attachments/files/31175998/stand.with.backing.zip)

### Checklist of files to include

- [x] Project file (always)
- [x] Log file (always)
- [ ] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)
- [ ] Performance profile — required for performance / lag issues

beuster:
I have the same problem, only difference with P1S.

finch6:
I have the exact same issue! I hope Bambu Labs sorts this out!

Spocky510:
I think I've found a workaround.
Try adding your third-party filaments using the "Read from AMS" option. After that, I had all the third-party brands listed separately in the settings and have been able to add them so far...
[...]

Issue #11871, opened 2026-08-11
Filament Manager cannot use custom user presets; editing Handy filaments in Studio marks them as Unsupported


fhack2:
### Type of issue

Other (functional bug, wrong result, UI glitch, etc.)

### Bambu Studio Version

2.8.1.55

### Where is the application from?

Bambu Lab Official website

### OS version

Windows 11

### Printer

Bambu Labs P2S

### How to reproduce

1. In Bambu Handy, create or select a custom user filament preset.
2. Add a filament using that preset.
3. Open Bambu Studio → Filament Manager.
4. Attempt to add a filament using the same custom preset.
5. Preset does not appear.
6. Edit the Handy‑created filament inside Studio (change color, add notes, etc.).
7. Save the filament.
8. Studio marks the filament as Unsupported Filaments.
9. Handy still shows the filament, but with Unknown color and Unknown filament type.

### Actual results

Studio cannot access or display custom user presets.
Editing a Handy filament in Studio breaks preset compatibility.
Filament becomes unusable in Studio and partially corrupted in Handy.

### Expected results

Custom user presets created in Handy should be selectable in Studio’s Filament Manager.
Editing a Handy filament inside Studio should not corrupt or invalidate the preset.
Filament metadata (color, type, preset association) should remain consistent across both apps.

### File uploads (project / log / crash dump / performance profile)

N/A

### Checklist of files to include

- [ ] Project file (always)
- [ ] Log file (always)
- [ ] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)
- [ ] Performance profile — required for performance / lag issues

fhack2:
When adding filament through Bambu Handy, I can select and apply custom user presets without any problem.
However, when adding filament through Filament Manager in Bambu Studio, those same user presets do not appear and cannot be selected.

Additionally, if I add a filament in Handy and then later edit that filament inside Bambu Studio (for example, changing the color or adding notes), Studio immediately marks the filament as:

Unsupported Filaments

Once this happens:

The filament cannot be selected in Bambu Studio.

The filament can still be selected in Bambu Handy, but Handy loses the correct color and filament type, showing it as Unknown.

This behavior makes it impossible to maintain consistent custom filament presets across both apps.

kennetjuh:
Just experienced the same issue. It seems like a data mismatch between the app and the slicer.

Issue #11654, opened 2026-07-21
Why are standard custom filaments (PETG/ABS) listed as "Unsupported Filaments" when assigning to AMS?


SUPEROHM88:
### Type of issue

Other (functional bug, wrong result, UI glitch, etc.)

### Bambu Studio Version

2.8.1.55

### Where is the application from?

Bambu Lab Official website

### OS version

win11

### Additional system information

cpu R7 7xxx
gpu rtx4060
ram16

### Printer

p1s

### How to reproduce

While I appreciate the recent update that allows assigning custom presets directly from the Filament Manager to AMS slots, I am currently facing an issue where many of my standard custom filaments (such as R3D PETG Basic / ABS Basic) are categorized under "Unsupported Filaments" and cannot be selected.

These are standard materials (PETG, ABS, PLA) that are completely AMS-compatible, but the AMS slot assignment menu greys them out or blocks them under the "Unsupported Filaments" dropdown.

### Actual results

While I appreciate the recent update that allows assigning custom presets directly from the Filament Manager to AMS slots, I am currently facing an issue where many of my standard custom filaments (such as R3D PETG Basic / ABS Basic) are categorized under "Unsupported Filaments" and cannot be selected.

These are standard materials (PETG, ABS, PLA) that are completely AMS-compatible, but the AMS slot assignment menu greys them out or blocks them under the "Unsupported Filaments" dropdown.

### Expected results

While I appreciate the recent update that allows assigning custom presets directly from the Filament Manager to AMS slots, I am currently facing an issue where many of my standard custom filaments (such as R3D PETG Basic / ABS Basic) are categorized under "Unsupported Filaments" and cannot be selected.

These are standard materials (PETG, ABS, PLA) that are completely AMS-compatible, but the AMS slot assignment menu greys them out or blocks them under the "Unsupported Filaments" dropdown.

### File uploads (project / log / crash dump / performance profile)

While I appreciate the recent update that allows assigning custom presets directly from the Filament Manager to AMS slots, I am currently facing an issue where many of my standard custom filaments (such as R3D PETG Basic / ABS Basic) are categorized under "Unsupported Filaments" and cannot be selected.

These are standard materials (PETG, ABS, PLA) that are completely AMS-compatible, but the AMS slot assignment menu greys them out or blocks them under the "Unsupported Filaments" dropdown.

### Checklist of files to include

- [ ] Project file (always)
- [ ] Log file (always)
[...]

PR #12079, opened 2026-08-29
fix(filament-manager): don't write a preset setting_id into a spool's filament id

marciogranzotto:
> **Defers to #12060.** @JaKuBisz identified this root cause first and #12060 is the more
> complete fix — it also adds a `Generic <type>` fallback for free-typed brands, a one-time
> migration that backfills the broken `setting_id` on existing spools, and the remaining-weight
> fix. This PR is offered only as a minimal one-file variant of the same correction, in case a
> focused change is easier to land first. **If #12060 is going in, close this one.**

Relates to #11937 (also reported in #11562, #11654, #11871). My duplicate issue #12078 is closed.

## Problem

A spool's `setting_id` field must hold a **filament_id** — that is the only thing it is ever matched against:

```cpp
// src/libslic3r/PresetBundle.cpp — get_filament_by_filament_id
if (filament_preset.filament_id == filament_id) { ... }
```

```cpp
// src/slic3r/GUI/AMSMaterialsSetting.cpp:1557
bool has_preset = bundle && bundle->get_filament_by_filament_id(sp->setting_id).has_value();
if (has_preset) { lib_brand_to_spools[brand].push_back(*sp); }
else            { unsupported_spools.push_back(*sp); }
```

`AddEditDialog` resolved that value as:

```js
matchedCloudFilamentId
|| matchedPresetItem?.setting_id
|| matchedPresetItem?.filament_id
|| initSpool?.setting_id
```

`matchedCloudFilamentId` only resolves on an exact match against the **global** cloud catalogue:

```js
settings.find((it) => it.filamentVendor === brand && getCloudSettingDisplayName(it) === name)
```

A user's own custom filament is never in that catalogue, so the first term is empty and the chain falls through to the preset's `setting_id`. Preset setting ids are `PFUS` + 14 hex while filament ids are `P` + 7 hex (or `GFxxx`), so the stored value can never equal any preset's `filament_id` — the spool is shown under **Unsupported Filaments** permanently. Re-picking the filament in the dialog re-runs the same fallback and does not help.

On my install this affected 13 of 35 spools. Every broken spool held a `PFUS…` value that was the `setting_id` of the very preset it should have pointed at. The same filament appeared in **both** states (PLA Velvet OK on 6 spools / broken on 7), confirming it is per-spool write data rather than a preset problem — the healthy ones were created earlier via a path that resolved the real `filament_id`.

## Change

Drop the wrong-namespace term at both resolution sites (`handleSubmit` and the `filaId` memo that mirrors it), and correct the comment that documented the old order.

## Why this is safe

In the `preset/list` payload, **all 325 preset items carry both `filament_id` and `setting_id`** — zero items have `setting_id` without `filament_id`. So `matchedPresetItem?.setting_id` could never act as a useful fallback: whenever it was set, `filament_id` was set too. It could only shadow the correct value.

That said, this is one machine's profile set — please confirm there is no profile source where an item legitimately has a `setting_id` but no `filament_id`. If such a case exists, ordering the term last rather than removing it would be the safer variant.

Note this fixes newly created spools only; records already written with a `PFUS…` id need their `setting_id` rewritten to the matching `filament_id`.

## Difference from #12060

#12060 orders the term last (`cloud filamentId > preset filament_id > preset setting_id`);
this PR removes it. Both fix the reported bug — ordering it last is the more conservative
choice and is fine. The note above about all 325 preset items carrying both ids is the
argument for removal, not a requirement.

This PR also fixes newly created spools only; it does **not** migrate records already written
with a `PFUS…` id. #12060 does, which is a further reason to prefer it.

### DIFF
--- a/src/slic3r/GUI/DeviceWeb/device_page/src/features/filament-manager/AddEditDialog.tsx
+++ b/src/slic3r/GUI/DeviceWeb/device_page/src/features/filament-manager/AddEditDialog.tsx
@@ -520,13 +520,12 @@ export function AddEditDialog({
 
   // STUDIO-17977 / Task 10: drive the colour palette from
   // `filament.colors.query_for_id`. The same fila_id resolution as
-  // handleSubmit (cloud filamentId > preset setting_id > preset filament_id)
+  // handleSubmit (cloud filamentId > preset filament_id)
   // is reused here so the candidate query keys off whatever the spool will
   // actually serialise as `setting_id` on save.
   const filaId = useMemo(
     () => (
       matchedCloudFilamentId
-      || matchedPresetItem?.setting_id
       || matchedPresetItem?.filament_id
       || initSpool?.setting_id
       || ''
@@ -1090,7 +1089,6 @@ export function AddEditDialog({
       remain_percent: remainPct,
       note,
       setting_id: matchedCloudFilamentId
-        || matchedPresetItem?.setting_id
         || matchedPresetItem?.filament_id
         || initSpool?.setting_id
         || '',


kovaacs:
Thanks, I'm using this build locally and it works great

PR #12094, opened 2026-08-31
FIX: filament vendor submenu never opens when each group has one preset

joranv1:
Hovering a category row (Custom, Bambu, Generic, ...) in the Project Filaments dropdown is supposed to open a submenu listing that vendor's presets. Often it does nothing at all.

`DropDown::hoverIndex()` short-circuits on `count == items.size()` and returns the row index unchanged, assuming equal counts mean nothing is grouped. That does not hold when every group contains exactly one item: each group still collapses to a single row, so the counts match while grouping is present. The `-i-2` group-row encoding is then never produced and `mouseMove()` cannot open the submenu.

This is why it looks intermittent. It depends only on the preset list: as soon as one vendor has two or more compatible presets the counts differ, the shortcut is skipped, and every submenu works again.

Track `has_groups` in `messureSize()` next to count and require it to be false before taking the shortcut. `selectedItem()` had the same flaw, mispositioning the check mark in the affected lists. Also guard the `subDropDown` dereference in `mouseMove()`, which this makes reachable.

### DIFF
--- a/src/slic3r/GUI/Widgets/DropDown.cpp
+++ b/src/slic3r/GUI/Widgets/DropDown.cpp
@@ -466,7 +466,9 @@ int DropDown::selectedItem()
 {
     if (selection < 0)
         return -1;
-    if (count == items.size())
+    // Same caveat as hoverIndex(): equal counts do not imply "no grouping",
+    // because single-member groups make them coincide. See messureSize().
+    if (count == items.size() && !has_groups)
         return selection;
     auto & sel = items[selection];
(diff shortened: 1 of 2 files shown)


joranv1:
7b728cae9 fixes the `hoverIndex()` half of this, the same root cause described above. I've dropped that part and rebased; what remains is the rest of the original report, which that commit didn't cover:

- `selectedItem()` has the identical stale `count == items.size()` shortcut, so the check mark lands on the wrong row in the affected lists.
- `mouseMove()` dereferences `subDropDown` without a null check on the path this makes reachable.

PR #12116, opened 2026-09-03
perf: avoid copying rvalue meshes when adding model volumes

VailElla:
## Summary
- avoid deep-copying `TriangleMesh` when `ModelObject::add_volume` receives an rvalue
- preserve convex-hull initialization after the mesh is moved into the volume
- add a regression test that verifies the vertex and index buffers are transferred

## Validation
- `cmake --build build/arm64 --target libslic3r --parallel 8`
- focused `fff_print_tests` regression: `ModelObject moves rvalue mesh storage into a volume`
- `git diff --check`

The full `fff_print_tests` target is currently blocked by pre-existing test sources that use an older `GCodeWriter` API (`set_extruder`, `lift`, and `retract_lift`); those failures are unrelated to this change.

### DIFF
--- a/src/libslic3r/Model.hpp
+++ b/src/libslic3r/Model.hpp
@@ -1243,6 +1243,12 @@ private:
         if (mesh.facets_count() > 1)
             calculate_convex_hull();
     }
+	ModelVolume(ModelObject *object, TriangleMesh &&mesh, ModelVolumeType type = ModelVolumeType::MODEL_PART) :
+		ModelVolume(object, std::make_shared<const TriangleMesh>(std::move(mesh)), type)
+	{
+		if (m_mesh->facets_count() > 1)
+			calculate_convex_hull();
+	}
     ModelVolume(ModelObject *object, const std::shared_ptr<const TriangleMesh> &mesh, ModelVolumeType type = ModelVolumeType::MODEL_PART) : m_mesh(mesh), m_type(type), object(object)
     {
 		assert(this->id().valid());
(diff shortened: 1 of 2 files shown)


Haidiye00 (commented):
Hi, under what bug is this? Does it need to be fixed this way? Can you tell me what this bug is?

VailElla:
Hi, thank you for taking a look.

This PR is not intended to fix a user-visible functional bug. It addresses an unnecessary deep copy on an existing rvalue code path.

ModelObject::add_volume(TriangleMesh&&) already forwards the mesh as an rvalue to ModelVolume. Before this change, ModelVolume did not have a matching TriangleMesh&& constructor, so the call selected the const TriangleMesh& constructor. That constructor creates a new TriangleMesh(mesh), copying the vertex and index buffers even though the rvalue no longer needs to retain its original storage.

The added constructor keeps the existing std::shared_ptr<const TriangleMesh> storage model, but moves the mesh into that shared object. This preserves immutable mesh storage, leaves the lvalue overload and other ownership behavior unchanged, and keeps the existing convex-hull initialization.

The regression test checks that the vertex and index buffer addresses are transferred rather than copied, and that the convex hull is still initialized. On high-poly models, this path can involve millions of vertices and indices, so the goal is to reduce loading memory traffic and peak memory without changing geometry or slicing behavior.

I chose this implementation to keep the change local to ModelVolume and avoid changing the existing ModelObject::add_volume API. If you prefer this to be tracked under a performance issue, or would like a dedicated benchmark added to the PR, I can add that and adjust the patch accordingly.

Haidiye00 (approved):
approve

PR #12164, opened 2026-09-08
Better filament dropdown and external bridge density

wildjokerss:
Sorry in advance for having 2 features in one. I wanted Bambu Studio to have a bridge density setting as it seems better than just increasing the bridge flow to have better bridges.

I've also felt a bit annoyed by the filament presets menu, as I like to have my own and see them once I open the filament profiles dropdown rather than having to go through sub menus.
CleanShot 2026-09-08 at 17 45 43
With this we have a better view of our filament profiles and, there's also a setting to choose how many you want to see (you may want to adjust based on screen size)
CleanShot 2026-09-08 at 17 46 56

The dropdown also adjusts to the profile name so it doesn't have to trim and me ending up guessing what's written in there.


Zulaner:
Simply fantastic, great job!
Thanks

PR #12176, opened 2026-09-09
Text tool: multi-line text with Enter and a Line Gap control

jeremykenedy:
## What

- Text tool input: **Enter inserts a line break**, like a normal text editor. Ctrl+Enter (Cmd+Enter on macOS) leaves the field.
- Multi-line text is generated correctly for flat text and for the three surface modes (Surround surface, Horizontal, Per character).
- New **Line Gap** slider/input under Text Gap (mm, default 0, negative allowed) to adjust the spacing between lines. Its range follows the active font, the negative end stops before the gap cancels the line height.
- Multi-line text and line gap survive project save/load.

## Demo

Flat text: Enter adds a second line, then the Line Gap slider changes the spacing between the lines.

two line text

Surface text on a cube (Surround surface mode), same steps:

two line surface text

## Why

Since 86641262a `ImGui::InputTextMultiline` no longer sets `ImGuiInputTextFlags_Multiline` itself and the text gizmo did not pass it, so Enter only unfocused the field and there was no way to type a second line. Even with a `\n` in the text, `GenerateTextJob` laid all glyphs out on one line and lost index alignment on the `\n` entry: `align_shape` skipped it in `offset_xy`, `create_all_char_mesh` dropped it from the mesh list, and positions/normals ended up with different sizes, which read out of bounds.

## How

- Text input: the gizmo passes `ImGuiInputTextFlags_Multiline`, so Enter inserts `\n` and Ctrl+Enter unfocuses (stock ImGui behaviour, `io.KeyCtrl` comes from `wxGetKeyState(WXK_CONTROL)` so it is Cmd+Enter on macOS). No change to `src/imgui`.
- Emboss: `\n` keeps its entry in every per-glyph array (align offsets, cursors), its cursor advance is 0, `EmbossShape` carries `line_height` and `first_line_offset_y` in mm, new `get_line_ranges()`. The backup-font `create_range_text` no longer counts white space as unknown glyphs, which used to flag every multi-line text with the "unsupported characters" tooltip.
- GenerateTextJob: `\n` gets an empty mesh placeholder so mesh, cursor and position arrays stay aligned. Flat text lays every line out separately (`calc_position_points_by_lines`), line Y offsets come from the glyph shapes. Surface text slices the object once per line at the line's base line and places each line on its own curve; the line offset baked in the shape is removed in `calc_mesh_offset`, which now uses the glyph's own align offset instead of glyph 0's.
- Line Gap: `m_line_gap` (mm) maps to `FontProp::line_gap` (font points), so it drives `get_line_height()` and the input preview. Stored in `TextInfo::m_line_gap`, 3mf attribute `line_gap` (0 for older files), undo/redo serialization.
- Line Gap range: slider and number input share one range derived from the active font, min `-0.9 * (ascent - descent + linegap)` in mm and max `max(10, 2x)` that. `get_line_height()` adds `line_gap` straight onto the font metrics, so without the lower bound a big negative gap drives the line height to zero or below and stacks the lines in reverse order. The stored value is only clamped when the user moves one of the two controls, never on its own: the Size field applies every keystroke, so clamping against a transient size would throw the gap away, and rewriting it while drawing would regenerate the volume and push an undo snapshot nobody asked for.
- Surface text is now robust to a base line that does not cross the object, which happens when the text sits near an edge or the gap is large. Every base line is cut first; a line whose own plane misses is placed on the nearest one that did, or on the text anchor when none did, since the anchor is on the surface by construction. `m_text_line_y` keeps the nominal per-line offset in that case, because `calc_mesh_offset` removes exactly the offset baked into that line's glyph shapes. The existing `JobException` is only raised when even the anchor misses, which is the pre-existing "text is off the object" case.
- 3mf: the `text` attribute is written with `xml_escape_double_quotes_attribute_value`, so a line break is stored as `&#xA;` instead of being normalized to a space by the XML parser on load.

## Tests

- New `tests/libslic3r/test_emboss.cpp`: line ranges, per-glyph array alignment for multi-line text, line gap effect on line distance. 3 test cases, 24 assertions, pass on macOS arm64. (`test_aabbindirect.cpp` and `test_3mf.cpp` in that directory do not compile on master, they are untouched.)
- Manual, macOS 26.5 / arm64 build, Bambu Lab P2S profile:
  - Flat text "Text" + Enter + "Line2": two lines, size Y 9.57 mm -> 22.57 mm at Line Gap 0, 29.61 mm at Line Gap 7.04. Cmd+Enter leaves the field.
  - Surface text (Surround surface) on a 60x40x40 box: two lines on the face, block centered on the anchor, height 22.57 mm -> 28.23 mm after Line Gap +5.67.
  - Surface text on a 25.6 mm cube with the anchor near the top edge, 8 mm font, three lines, Line Gap dragged to the low end: lines land on the part, the slider stops at -7.20 (0.9 x the 8 mm line height) and nothing is dropped. Before the base line handling above, adding the second line alone raised "The hit polygon is null" because re-centering the block pushed the only line's plane past the top of the cube.
  - Save/reopen: `text="Text&#xA;Line2" line_gap="5.67"` in `model_settings.config`, text comes back with the line break.

### DIFF
--- a/src/libslic3r/Emboss.cpp
+++ b/src/libslic3r/Emboss.cpp
@@ -1475,7 +1475,8 @@ void Slic3r::Emboss::text2vshapes(EmbossShape &                emboss_shape,
             text_scales.emplace_back(real_scale);
         }
         cur_x = cursor.x() * standard_scale;
-        text_cursors.emplace_back(cur_x - last_x);
+        // '\n' resets cursor to line start, it is not an advance
+        text_cursors.emplace_back(letter == '\n' ? 0.f : cur_x - last_x);
         text_absolute_cursors.emplace_back(cur_x);
         last_x = cur_x;
     }
@@ -1508,6 +1509,12 @@ void Slic3r::Emboss::text2vshapes(EmbossShape &                emboss_shape,
     }
     emboss_shape.shapes_with_ids = result;
     emboss_shape.align_type      = std::pair<int, int>((int) font_prop.align.first, (int) font_prop.align.second);
+    // line layout of multi line text (line Y offsets are already baked in shapes)
+    unsigned count_lines             = get_count_lines(text);
+    emboss_shape.line_height         = static_cast<float>(get_line_height(font, font_prop) * standard_scale);
+    emboss_shape.first_line_offset_y = (count_lines > 1) ?
+        static_cast<float>(get_align_y_offset_in_mm(font_prop.align.second, count_lines, font, font_prop) -
+                           get_align_y_offset_in_mm(font_prop.align.second, 1, font, font_prop)) : 0.f;
 }
 
 #include <boost/range/adaptor/reversed.hpp>
@@ -1551,6 +1558,20 @@ unsigned Emboss::get_count_lines(const ExPolygonsWithIds &shapes) {
     for (const ExPolygonsWithId &shape_id : shapes)
         if (shape_id.id == ENTER_UNICODE)
             ++result;
+    return result;
+}
+
+Emboss::LineRanges Emboss::get_line_ranges(const ExPolygonsWithIds &shapes)
+{
+    LineRanges result;
+    size_t     first = 0;
+    for (size_t i = 0; i < shapes.size(); ++i) {
+        if (shapes[i].id != ENTER_UNICODE)
+            continue;
+        result.emplace_back(first, i);
+        first = i + 1;
+    }
+    result.emplace_back(first, shapes.size());
     return result;
 }
 
@@ -1622,6 +1643,8 @@ std::string Slic3r::Emboss::create_range_text(std::string &text, std::vector<std
     *exist_unknown                              = false;
     bool                     temp_exist_unknown = false;
     std::wstring             not_dup_text       = remove_duplicates(boost::nowide::widen(text));
+    // white spaces are no glyphs, the single font range skips them too
+    not_dup_text.erase(std::remove_if(not_dup_text.begin(), not_dup_text.end(), [](wchar_t wc) { return wc == L'\n' || wc == L'\r' || wc == L'\t'; }), not_dup_text.end());
     std::sort(not_dup_text.begin(), not_dup_text.end());
     std::vector<std::string> results;
     results.reserve(fonts.size());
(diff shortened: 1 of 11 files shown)


Haidiye00 (commented):
Hi, please provide a test GIF, thanks

jeremykenedy:
Here is a test GIF (macOS build of this branch).

Flat text: Enter adds a second line, then the new Line Gap slider changes the spacing between the lines.

two line text

Surface text on a box (Surround surface mode), same steps:

two line surface text

Haidiye00:
Hello, why didn't you just use enter to wrap the line? Please help merge the other two patches into one. Thank you

jeremykenedy:
Hello, I chose shift and enter because that is usually the standard for doing a new line in other programs that normally have it when you hit enter to be the accept button. Such as in Slack if you hit enter it sends the message but if you hit shift, enter it does a new line. So I did it so that way people who use the keyboard only want to hit enter to hit OK they still can do so. On Sep 9, 2026, at 6:39 PM, Bambu_Haidiye00 ***@***.***> wrote:﻿Haidiye00 left a comment (bambulab/BambuStudio#12176)
Hello, why didn't you just use enter to wrap the line? Please help merge the other two patches into one. Thank you

Haidiye00:
Hello, I think using carriage return and line wrapping would be better, just like using Word software to do it

jeremykenedy:
@Haidiye00 

> Hello, I think using carriage return and line wrapping would be better, just like using Word software to do it

Switched it to Enter. Enter now inserts the line break like Word does, and Ctrl+Enter (Cmd+Enter on mac) leaves the field.

That also let me drop the imgui change completely. The gizmo just passes ImGuiInputTextFlags_Multiline now, so src/imgui is untouched.

Also squashed it down to one commit. The GIFs above still match, only the key changed.

jeremykenedy:
@Haidiye00 Anything else you want me to do with this?

Haidiye00:
Hi, the internal team, including interaction design, is still reviewing this, please be patient~

Haidiye00 (approved):
approve

Issue #8060, opened 2025-08-31
Arabic text


honeydashtii:
Hello,  
I would like to request support for Arabic language in Bambu Studio.  
Currently, when I try to add Arabic text, it appears broken because the software does not handle Right-to-Left (RTL) text correctly.  

Arabic is spoken by more than 400 million people worldwide, and supporting it would be very helpful for many users.  

Thank you for considering this feature request.

penanot:
> Hello, I would like to request support for Arabic language in Bambu Studio. Currently, when I try to add Arabic text, it appears broken because the software does not handle Right-to-Left (RTL) text correctly.
> 
> Arabic is spoken by more than 400 million people worldwide, and supporting it would be very helpful for many users.
> 
> Thank you for considering this feature request.

It could be a nice feature to add on.

alkobra1394:
I would like to request support for Arabic language in Bambu Studio

shadowsman22:
I second  that!

borfoo3:
Bumping to support this!

PR #12185, opened 2026-09-10
ENH: text tool accepts emoji and other characters above U+FFFF

### LINKED ISSUE #12144: Support Emojis in Text-Editor
I would like to be able to add emojis in the Text Editor.
If I use Noto Emoji as font and insert an emoji, it get's converted into a question mark:

Image

In OpenSCAD / Makerlab it's possible to add emojis with font "Noto Emoji:color=false", it would be great if that would work directly in bambu studio, too.

jeremykenedy:
Fixes #12144 (Support Emojis in Text-Editor).

This solves that issue: an emoji typed or pasted into the Text tool no longer turns into `?`. It shows in the input box and is generated as a real 3D shape, with any font, and with an installed emoji font such as Noto Emoji selected directly, which is the exact case from the report.

## What

- Emoji (and anything else above U+FFFF) can be typed or pasted into the text tool and come out as a real 3D shape, flat or on a surface.
- Works with any font: when the selected font has no outline for the character, the glyph comes from Symbola (already shipped in `resources/fonts`).
- An installed outline emoji font such as Noto Emoji can be selected directly, as in the report. A bitmap emoji font (Apple Color Emoji, Noto Color Emoji) can be selected too, the outlines then come from the backup fonts.
- The input box, the font list previews and the generated mesh all show the glyph. Variation selectors that emoji pickers append (`❤️` is `U+2764 U+FE0F`) no longer show up as `?`.

## Demo

Flat text, default font: the emoji come from the Symbola fallback. Then the font is switched to Noto Emoji.

emoji text

Surface text on a cube (Surround surface):

emoji surface text

## Why

Several places dropped any character outside the Basic Multilingual Plane, each one on its own enough to turn an emoji into `?`:

1. ImGui was built with a 16-bit `ImWchar`, so its input queue replaced every code point above U+FFFF with U+FFFD before `InputText` saw it. No font in the atlas has that glyph, so the box draws ImGui's fallback `?`, and the gizmo shapes a `?` for text it cannot draw.
2. On Windows `wxChar` is UTF-16, so a single emoji arrives as two `wxEVT_CHAR` events carrying a surrogate pair, and each half was pushed to ImGui as its own (invalid) character.
3. `Emboss` converted the UTF-8 text to `std::wstring`, which is 16 bit on Windows, so the same split happened again right in front of the glyph lookup. `EmbossJob::create_all_char_mesh` indexed the text as a `wxString` for the same reason.
4. Most fonts have no emoji glyph at all, and the emoji fonts that ship with the OS (Apple Color Emoji, Segoe UI Emoji, Noto Color Emoji) are bitmap or COLR fonts, so there is no outline to extrude even when the code point survives.

## How

- `src/imgui/imconfig.h`: `IMGUI_USE_WCHAR32`. Nothing else in `src/imgui` changes; the codebase already uses `ImWchar` generically everywhere.
- `ImGuiWrapper::update_key_data`: where `sizeof(wxChar) == 2` the char event goes through `io.AddInputCharacterUTF16`, which ImGui already provides for exactly this; other platforms keep `AddInputCharacter`.
- `Emboss`: `text2vshapes`, `align_shape`, `letter2shapes`, `get_count_lines` and `remove_duplicates` work on `std::u32string` / `char32_t` instead of `std::wstring` / `wchar_t`; `create_range_text` decodes the same way. New `Emboss::to_utf32()` / `to_utf8()` (`boost::locale::conv::utf_to_utf`, header only, already used in `src/slic3r`). `EmbossJob::create_all_char_mesh` uses the same decoding and keeps one mesh per glyph (empty or not) so the cursor arrays stay aligned with the meshes.
- `BackupFonts::generate_backup_fonts`: appends `resources/fonts/Symbola.ttf` (loaded from file, new `gener_font_with_cache_from_file`) after the OS fonts, so the existing CJK/Arabic fallbacks keep priority and Symbola only answers for what they do not cover. The backup loop in `letter2shapes` skips a font whose glyph has no outline instead of stopping there, and the base font of the input box is chosen the same way, judged on the text's own characters (`font_has_any_glyph_in_ranges` looks for an outline, not just a mapped code point), so selecting a bitmap emoji font still gives a readable input box.
- `align_shape` (backup font overload): the vertical offset of a backup glyph was multiplied by `int(text_scale / standard_scale)`, which is 0 for every main font with 1000 units per em (Symbola has 2048), pushing the glyph off the base line. It uses the real ratio now, on the left aligned path too.
- `create_range_text` (multi font): the glyph range for the input box is the union of what all fonts cover instead of one font's set, so text that needs both a CJK backup and Symbola shows every glyph; white space is not counted as an unknown glyph.
- Zero width marks (`U+FE00..FE0F`, `U+200C`, `U+200D`): `is_zero_width_mark()`, skipped in `letter2shapes` (no shape, no advance) and in `create_range_text`, and given an empty custom glyph in the text tool atlas so the input box draws nothing for them.
- `Emboss::create_font_file(const char*)` opens the file with `boost::nowide::fopen` so the UTF-8 resources path works on Windows.

## Tests

`tests/libslic3r/test_emboss.cpp` (new): UTF-8/UTF-32 round trip above U+FFFF, `create_range_text` with and without the glyph (single font and multi font union, white space and zero width marks), backup font supplying the emoji outline with aligned per-glyph arrays, zero width marks producing no advance, and a backup glyph keeping its position in mm between a 1000 and a 2048 units-per-em font. 8 test cases pass on macOS arm64. (`libslic3r_tests` as a whole does not link on master, the new file was linked and run on its own.)

Manual, macOS 26.5 / arm64 build:

- Flat text and surface text (Surround surface) on a cube with `Hi 😀🚀`, `🎉`, `👍`, `❤`, `🚀`: input box shows the glyph, mesh is generated, no mesh warnings.
- Noto Emoji (monochrome, from google/fonts) installed and selected in the font list: `🦄Hi 😀🚀` renders in Noto's outlines, no warnings. Font list previews show the emoji for every font.
- Apple Color Emoji selected: `Hi🚀` shows in the input box (outlines from the backup fonts) and on the plate.
- Helvetica Neue (1000 units per em): `Ag🚀` sits on one base line. Helvetica with `你好🚀`: CJK and emoji both visible in the input box. `Hi ❤️🚀` (heart with U+FE0F): no `?` in the box.
- Paste (Cmd+V) and typing through the emoji picker path both work on macOS.

Builds: the branch was built with this repository's `build_all.yml` on my fork for ubuntu-22.04, ubuntu-24.04, ubuntu-26.04, windows-latest, macos-15-intel, macos-15 and macos-26, all green: https://github.com/jeremykenedy/BambuStudio/actions/runs/34491495362. The Windows-only paths (surrogate pairs, 16-bit `wchar_t`) are the reason for the `u32string` change; runtime testing was done on macOS only.

## Known limits

- Symbola 9 covers emoji up to about Unicode 9; newer ones (🥺, 🫠, 🧑) need a selected font that has them, otherwise they show as `?` like any other unknown glyph.
- Skin tone modifiers and ZWJ sequences render as their separate glyphs, there is no shaping engine.
- Symbola's own unicorn (`U+1F984`) outline produces a "reversed faces" mesh warning; every other emoji tried is clean, and the same character from Noto Emoji is clean.

### DIFF
--- a/src/imgui/imconfig.h
+++ b/src/imgui/imconfig.h
@@ -53,7 +53,10 @@
 //#define IMGUI_USE_BGRA_PACKED_COLOR
 
 //---- Use 32-bit for ImWchar (default is 16-bit) to support unicode planes 1-16. (e.g. point beyond 0xFFFF like emoticons, dingbats, symbols, shapes, ancient languages, etc...)
-//#define IMGUI_USE_WCHAR32
+// Needed by the text tool. With a 16-bit ImWchar every code point above U+FFFF
+// (emoji live in plane 1) is replaced by IM_UNICODE_CODEPOINT_INVALID on the way into
+// InputText, so the character is destroyed before it ever reaches the emboss pipeline.
+#define IMGUI_USE_WCHAR32
 
 //---- Avoid multiple STB libraries implementations, or redefine path/filenames to prioritize another version
 // By default the embedded implementations are declared static and not available outside of Dear ImGui sources files.
--- a/src/libslic3r/Emboss.cpp
+++ b/src/libslic3r/Emboss.cpp
@@ -4,6 +4,8 @@
 #include <numeric>
 #include <cstdlib>
 #include <boost/nowide/convert.hpp>
+#include <boost/nowide/cstdio.hpp>
+#include <boost/locale/encoding_utf.hpp>
 #include <boost/log/trivial.hpp>
 #include <ClipperUtils.hpp> // union_ex + for boldness(polygon extend(offset))
 #include "IntersectionPoints.hpp"
@@ -1070,7 +1072,7 @@ std::unique_ptr<FontFile> Emboss::create_font_file(
 
 std::unique_ptr<FontFile> Emboss::create_font_file(const char *file_path)
 {
-    FILE *file = std::fopen(file_path, "rb");
+    FILE *file = boost::nowide::fopen(file_path, "rb");
     if (file == nullptr) {
         assert(false);
         BOOST_LOG_TRIVIAL(error) << "Couldn't open " << file_path << " for reading.";
@@ -1192,7 +1194,7 @@ int Emboss::get_line_height(const FontFile &font, const FontProp &prop) {
 }
 
 namespace {
-ExPolygons letter2shapes(wchar_t letter, Point &cursor, FontFileWithCache &font_with_cache, const FontProp &font_prop, fontinfo_opt &font_info_cache)
+ExPolygons letter2shapes(char32_t letter, Point &cursor, FontFileWithCache &font_with_cache, const FontProp &font_prop, fontinfo_opt &font_info_cache)
 {
     assert(font_with_cache.has_value());
     if (!font_with_cache.has_value())
@@ -1216,7 +1218,7 @@ ExPolygons letter2shapes(wchar_t letter, Point &cursor, FontFileWithCache &font_
         cursor.x() += count_spaces * space->advance_width;
         return {};
     }
-    if (letter == '\r')
+    if (letter == '\r' || is_zero_width_mark(letter))
         return {};
 
     int unicode = static_cast<int>(letter);
@@ -1239,7 +1241,7 @@ void letter2shapes(ExPolygons &       result,
                    float              cur_scale,
                    FontFileWithCache &real_use_font,
                    float &            real_scale,
-                   wchar_t            letter,
+                   char32_t           letter,
                    Point &            cursor,
                    FontFileWithCache &font_with_cache,
                    const FontProp &   font_prop,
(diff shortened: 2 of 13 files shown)


Haidiye00 (approved):
Approve

traien:
Can you push this?

Since this will solve the Arabic lang issue too: https://github.com/bambulab/BambuStudio/issues/8060

Haidiye00:
Hello, could you please help confirm if there are any conflicts in the latest code? Or do you need me to help you submit and set the author as you?

Haidiye00:
Hello, please merge into one patch. Thank you

jeremykenedy:
@Haidiye00 Done, squashed into one commit on top of the latest master, ready to merge.

PR #12194, opened 2026-09-11
ENH: per-filament control of purifying the chamber air at print end

### LINKED ISSUE #11276: Add support to purify air at end only for specific materials
Something I always get annoyed by is that If I'm printing ABS on my X2D, I always turn on purify Air at end because of the VOCs ABS produces, this is a cool feature however it takes time to do, I completely understand this. However if I start a PLA/PETG print I don't want to purify the air at end because these materials don't produce harmful VOCs. I understand that I could go to print settings and change it every time but thats not very convinient and If I already started the print I have to cancel it to turn the feature off then restart the task.

jeremykenedy:
**Closes #11276** — [Add support to purify air at end only for specific materials](https://github.com/bambulab/BambuStudio/issues/11276)

### The problem

"Purify air at print end" is a single printer-wide toggle under **Device → Print options**. It is worth the extra time for ABS or ASA and wasted on PLA or PETG, but it is not per-job, and the control is disabled while a job is running. So the only way to skip it for a PLA print is to cancel the job, change the setting, and start over. That is what the issue reports.

### The change

A per-filament option under **Filament → Cooling → Air purification**:

| Value | Effect when this filament is on the plate |
|---|---|
| **Follow printer setting** (default) | Studio does not touch the printer's setting |
| Off | Ask the printer not to purify after this print |
| Internal Circulation | Purify through internal circulation |
| Exhaust | Purify by exhausting |

Per-filament purify air at print end

Because the default is *Follow printer setting*, nothing changes for anyone who never opens the option. When at least one filament on the plate does name a value, `SelectMachineDialog::on_send_print` resolves it and pushes it to the printer with the existing `print_option` / `air_purification` command before the job starts. The most demanding request wins — Exhaust over Internal Circulation over Off — so a filament that wants purification is never overruled by one that does not. Exhaust falls back to internal circulation on a machine with no chamber exhaust duct.

The two mode names are the strings the device's own print options dialog already uses.

### Capability gating

`support_air_filtration` is **0** on the X2D — it gates the during-print exhaust fan, which is a different feature. Gating on it would have hidden this option on exactly the machine the issue is about. So this adds `support_purify_air_at_print_end`, set on the X2D profile, and the group is hidden on machines without it. Other models can be opted in by adding the flag to their profile.

Changing the option does not invalidate any slicing step — it is listed with the other options that have no influence on the generated G-code.

### What I verified

- Builds clean on macOS arm64.
- The option appears on an X2D profile and is hidden on P1S/P2S, which do not carry the new flag.
- Round-trip: setting it and saving writes `"purify_air_at_print_end": ["external"]` to the user preset; after a full restart the preset loads back showing **Exhaust**.

### What I could not verify

I do not have an X2D, so the send-time half — the printer actually receiving the `print_option` command and honouring it — is **not** verified against hardware. The command itself is the existing one used by the Print options dialog; what is new is the call site and the resolution logic. Worth a check on a real machine before this lands.

Happy to rework the shape of this — in particular, whether the setting should be per-filament at all versus a new per-job field, and whether writing a persistent device setting from the send path is acceptable. The current design leans on the default value being a no-op so that nothing is written unless a user explicitly opts a filament in.

### DIFF
--- a/resources/profiles/BBL/machine/Bambu Lab X2D 0.4 nozzle.json
+++ b/resources/profiles/BBL/machine/Bambu Lab X2D 0.4 nozzle.json
@@ -190,6 +190,7 @@
         "20"
     ],
     "support_chamber_temp_control": "1",
+    "support_purify_air_at_print_end": "1",
     "upward_compatible_machine": [
         "Bambu Lab A1 0.4 nozzle",
         "Bambu Lab P2S 0.4 nozzle",
--- a/src/libslic3r/Preset.cpp
+++ b/src/libslic3r/Preset.cpp
@@ -1113,6 +1113,7 @@ static std::vector<std::string> s_Preset_filament_options {/*"filament_colour", 
     "filament_start_gcode", "filament_end_gcode",
     //exhaust fan control
     "activate_air_filtration","during_print_exhaust_fan_speed","complete_print_exhaust_fan_speed",
+    "purify_air_at_print_end",
     // Retract overrides
     "filament_retraction_length", "filament_z_hop", "filament_z_hop_types", "filament_retraction_speed", "filament_deretraction_speed", "filament_retract_length_nc", "filament_retract_restart_extra", "filament_retraction_minimum_travel",
     "filament_retract_when_changing_layer", "filament_wipe", "filament_retract_before_wipe",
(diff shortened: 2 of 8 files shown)


jeremykenedy:
Anything I can do to help this on along? @XinZhangBambu @Haidiye00

Issue #12253, opened 2026-09-17
2.8.3.66 (open beta) mouse over tooltip on settings lock up appimage


DrBlackross:
### Type of issue

Crash / freeze (the app crashes, hangs, or stops responding)

### Bambu Studio Version

2.8.3.66 appimage

### Where is the application from?

Bambu Lab GitHub releases

### OS version

Linux 7.0.0-31-generic #31~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Aug 10 09:38:02 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Printer

Bambu A1

### How to reproduce

mouse over a setting to popup a tooltip, app freezes, need to kill via htop as root. (xkill will not work)

### Actual results

appimage freezes completely

### Expected results

(lol expected results) it should work?

### File uploads (project / log / crash dump / performance profile)

doesnt matter any 3mf does this

### Checklist of files to include

- [ ] Project file (always)
- [ ] Log file (always)
- [ ] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)
- [ ] Performance profile — required for performance / lag issues

DrBlackross:
rolling back to a previous version

BambulabRobot:
Hi, thank you for taking the time to report this issue and for the details you've already provided — they are very helpful!

To help us pinpoint the root cause more accurately, it would be great if you could also share:

- If possible, a crash dump (DMP) file would help us locate the exact issue. You can follow the steps here: https://wiki.bambulab.com/en/bambu-studio/troubleshoot/crash-freeze-issue
- Please upload the log file to assist in troubleshooting the issue. You can find instructions here: https://wiki.bambulab.com/en/x1/troubleshooting/bambu-studio-log

Of course, if you have any other context (screenshots, screen recordings, software version, OS info, etc.),feel free to add them as well. We appreciate your help and patience!

tanganyika:
I have the same issue and have a gdb trace. Thread 1 is stuck here 

gtk_main → event dispatch → ~10 nested gtk_container_propagate_draw/g_signal_emit calls (drawing the tooltip's widget hierarchy) → cairo_stroke() → a chain of internal Cairo calls → XRenderCompositeTrapezoids → _XSend → xcb_writev → poll().

[bambu-hang-backtrace.txt](https://github.com/user-attachments/files/32367934/bambu-hang-backtrace.txt)

[...]

PR #12262, opened 2026-09-18
Fix Linux freeze on slicer-setting tooltips

### LINKED ISSUE #12233: Bambu Studio (Linux) freezes permanently when hovering slicer-setting tooltips
### Type of issue

Crash / freeze (the app crashes, hangs, or stops responding)

### Bambu Studio Version

02.08.02.60

Bambu Lab GitHub releases

### OS version

Omarchy 4.0.3

| Item | Value |
|---|---|
| OS | Omarchy (Arch Linux), kernel + X11 tooling current |
| Session | Hyprland 0.56.2 (Wayland), XWayland 24.1.13 (xorg-server-xwayland 24.1.13-1) |
| GPU / mesa | AMD Radeon Pro 5500M, mesa 1:26.2.2-1 |
| GTK | gtk3 1:3.24.52-1 (system) |

### Printer

P2S

### How to reproduce

1. Open `NO_CameraQuickRelease-With_Easy_Disassemble.3mf`.
2. In the left settings panel, hover a slicer parameter row until its tooltip /
   description text appears.
3. Continue moving the pointer (into or out of the panel). The tooltip stops following
   the pointer and the entire window freezes: hover states no longer update, the 3D view
   does not redraw on scroll/click, tooltips stay on screen, no input is accepted.
4. Application must be force-killed.

### Actual results

The application freezes completely — window stops repainting, tooltips/UI text remain on screen forever, all input is ignored, and the process must be killed — **after hovering a slicer setting** in the left settings panel. The tooltip never dismisses. Fully reproducible (3/3).

### Expected results

Tooltip is dismissable and program remains interactive.

[476085_enc.log.part00.zip](https://github.com/user-attachments/files/32221424/476085_enc.log.part00.zip)
[476085_enc.log.part01.zip](https://github.com/user-attachments/files/32221422/476085_enc.log.part01.zip)
[1278575_enc.log.part00.zip](https://github.com/user-attachments/files/32221419/1278575_enc.log.part00.zip)
[1278575_enc.log.part01.zip](https://github.com/user-attachments/files/32221425/1278575_enc.log.part01.zip)
[1278575_enc.log.part02.zip](https://github.com/user-attachments/files/32221421/1278575_enc.log.part02.zip)
[1278575_enc.log.part03.zip](https://github.com/user-attachments/files/32221423/1278575_enc.log.part03.zip)
[1278575_enc.log.part04.zip](https://github.com/user-attachments/files/32221420/1278575_enc.log.part04.zip)
[1278575_enc.log.part05.zip](https://github.com/user-attachments/files/32221426/1278575_enc.log.part05.zip)

[NO_CameraQuickRelease-With_Easy_Disassemble.zip](https://github.com/user-attachments/files/32221457/NO_CameraQuickRelease-With_Easy_Disassemble.zip)

[threaddump.txt](https://github.com/user-attachments/files/32221465/threaddump.txt)

[frames.zip](https://github.com/user-attachments/files/32221484/frames.zip)

[live-main-tt.zip](https://github.com/user-attachments/files/32221488/live-main-tt.zip)

- [x] Project file (always)
- [x] Log file (always)
- [x] Crash dump / report — required for crash / freeze issues (Windows .dmp / macOS .ips / Linux gdb backtrace)

ClusterM:
Fixes #12233 and #12253: hovering a slicer-setting tooltip on Linux grabs all input (and can hang the X11 connection), so the card never dismisses and Studio has to be killed.

`ParamTooltip` is a hover card, but it inherited `wxPopupTransientWindow`. On GTK, `Show()` takes a pointer grab (`gtk_grab_add` + `gdk_device_grab`). The card is shown with `Show()` / `Hide()`, not `Popup()` / `Dismiss()`, so the grab is taken and the click-outside dismiss path is never armed. Mouse events never reach the settings row, `Hide()` is never called, and the UI stops accepting clicks.

This PR switches the card to `wxPopupWindow` (no grab), which matches how other hover toasts in this tree already work.

On GTK we also skip `SetShape()`. A shaped GDK popup plus cairo rounded-rect painting can block the X11 socket (`cairo_stroke` → `XRenderCompositeTrapezoids` → `poll`), which is the hang in #12253. `OnPaint` still draws a rounded fill, but the window itself stays rectangular — we give up true rounded corners on GTK for compatibility. Treat that as temporary until this UI is rewritten against native GTK widgets, which is a much larger change.

Windows and macOS keep the shaped rounded card.


tonghao-bbl:
@lanewei120 LGTM

KaHa6uc:
Tested the CI artifact from this PR (`Bambu_Studio_ubuntu-24.04_PR-12262.AppImage`) — **it fixes
the freeze.** Independent confirmation on a different stack from #12233, with before/after
instrumentation.

**Environment** — native X11, not XWayland (#12233 is Hyprland + XWayland on Arch):

- Debian 13, kernel 6.12.107
- KDE Plasma / KWin, X11 session
- NVIDIA proprietary driver
- Affected build: Bambu Studio Beta `02.08.03.66` (AppImage)

Repro: hover a parameter in the left slicer-settings panel until the tooltip card appears.
Was fully reproducible before the patch, and I can no longer reproduce it with the PR build.

### Before the patch

Both failure modes in your description show up as distinct, measurable symptoms:

1. **The grab is taken when the card is *shown*, not on click.** With the card up,
   `XGrabPointer()` from a separate client returns `AlreadyGrabbed (1)`. Every pointer event in
   the session routes to the card — the only clickable thing on the whole desktop is the
   hyperlink inside it. The keyboard is unaffected, so Alt+Tab and typing into other apps keep
   working, which is exactly why this presents as a compositor hang rather than an app bug.
   Consistent with `gtk_grab_add + gdk_device_grab` firing from `Show()`.

2. **Repaint dies while the event loop stays live.** CPU went 2 ticks/2s idle → 17 ticks during
   a mouse sweep → 36 ticks on Ctrl+S, so input was still being consumed and acted on. But
   `import -window <main>` captures with the pointer parked vs. swept gave
   `compare -metric AE` = **0 pixels differ** — the window never redrew.

   One refinement: all X11 socket `Recv-Q` were 0, so the connection was **not** blocked at the
   socket level; the client kept draining events throughout. The symptom was specifically that
   *painting* stopped while the loop stayed alive. That may help narrow where the
   `SetShape()`/cairo interaction actually stalls.

   Useful side effect — since input still works blind, `xdotool key ctrl+s` **successfully saved
   an unsaved project** in this state (dirty asterisk cleared, `.3mf` mtime updated). Anyone
   hitting this can rescue their slice before force-killing.

### With the PR build

- 12/12 `XGrabPointer` probes over 8s return free while actively using the settings panel.
- The tooltip card is left `IsUnMapped` when idle, rather than staying viewable and holding the
  grab.
- No freeze across normal use of the slicer-settings panel.

---
Honest P.S.: This comment was generated with the help of AI which was monitoring the freeze and its output is correct.

PR #12264, opened 2026-09-18
ENH: add 'C' keyboard shortcut to center the selection on the plate

jomixlaf:
## Summary

Adds a plain **C** keyboard shortcut in the 3D view that centers the selected object(s) on their plate. It reuses the exact logic of the existing right-click **Center** menu item, so behaviour is identical to that action.

## Changes

- New `EVT_GLCANVAS_CENTER_SELECTION` canvas event, posted from `GLCanvas3D::on_char` on a plain `C` press when the canvas is the 3D view and no painting gizmo is active (painting gizmos use `C` to pick the circle brush, so they keep it).
- `Plater` binds the event and applies the same guards as the context menu: nothing happens when the selection is empty or the object is outside every plate; otherwise it selects the object's plate and centers the selection.
- The dev-only gcode window toggle previously bound to `C` (`#if !BBL_RELEASE_TO_PUBLIC`) is preserved for the preview canvas in non-release builds.
- The keyboard shortcuts dialog now lists `C` as "Center selected objects on the plate". The previous entry said "Gizmo cut", but the cut gizmo is not bound to plain `C`.

## Testing

- Built `libslic3r_gui` on macOS (arm64) with no new warnings.


Haidiye00 (commented):
Hello, please provide both the original and modified GIFs so we can verify the changes you've made. Thank you

Issue #2967, opened 2023-11-26
Duplicate Variable Layer Height between objects


drpeppershaker:
**Is your feature request related to a problem? Please describe.**
Not a problem necessarily, but it would be an excellent feature. Currently you cannot use a prime tower when printing multiple objects with variable layer heights. The solution, or workaround, is to select all the objects on the plate and assemble them and apply the _same_ variable layer height to all of them. 

This works okay, except now you have one giant assembly--which renders the 'skip objects' feature of the printers useless. Because if all of your objects are one assembly you can't skip a single one any more.

**Describe the solution you'd like**
I would like to be able to apply the same variable layer height settings to multiple objects without needing to assemble them together. Similar to how you are able to copy & paste Height Range modifiers.

**Describe alternatives you've considered**
An alternative solution would be to manually go in and create a height range modifier that matches the layer height on a per layer basis and use that on each object. However, this is very difficult to do on models with many layer height changes.

**Additional context**

<img width="421" alt="Screen Shot 2023-11-26 at 12 01 41 PM" src="https://github.com/bambulab/BambuStudio/assets/67330126/273f7552-2d5c-417c-92fc-e9075556ceb7">

shiinee:
+1. I'd like to be able to have both the prime tower and a purge object(s) on my multicolor prints with variable layer height. It would be fine with me if the variable layer height calculation either adapted to the purge object's shape, or just copied the layer height settings from the primary object. Currently no such print configuration is possible even with the assembly workaround, because you can't designate one part of an assembly as a purge object.

lordbbk:
+1. The only useful way of utilizing purge to object, if you duplicate the main object, and use it as a purge target, ignoring the colors... 

Just simply allowing copying the layer heights to another object would be a simple solution - since it will be just an initial state, you can reset it via the standard way, using the variable layer height menu.

ShaiDiamant:
This would be very helpful to anyone using the flush to object option. 
We want to be able to print a high quality primary object and still purge to object

Bathsheba:
+1
I have multiple objects that are the same height and need the same Variable Layer Height profile. I don't want to merge them because then I can't skip a single failed object, and also they have many parts and it would ruin my organization.

A different feature which would also solve this problem is, a way to group objects without merging them, and Variable Layer Height can be applied to the group.

gilbsgilbs:
I made my own attempt at fixing this in #11868 . Originally for another fork of this slicer, so feel free to test.

PR #12275, opened 2026-09-19
Copy/paste variable layer height between objects

MartinKlefas:
## Summary

The painted Variable Layer Height profile had no way to move between objects. Height Range Modifiers already have cross-object copy/paste — this adds the same to the continuous painted profile, via two new buttons on the Variable Layer Height tool's own overlay.

Matches #2967 (open).

## Why

A purge/sacrificial object can never pick up a main object's variable layer height today without merging the two into one multi-part object — which loses per-object "skip on printer" support. #2967 asks for exactly this, specifically in the purge-object context.

## Implementation

- **Copy / paste plumbing**: `LayersEditing::copy_layer_height_profile()` stashes the selected object's working profile into a buffer that survives selection changes; `paste_layer_height_profile()` writes it onto whichever object is selected at paste time, via the same `ModelObject::layer_height_profile.set()` call Reset/Adaptive/Smooth already use — so `Print::apply()`'s existing diff picks it up with no new invalidation plumbing.
- **Height-mismatch handling**: pasting onto a shorter object uses the source schedule up to the destination's own top Z, interpolated at exactly that height. Pasting onto a taller object uses the source schedule up to its own top, then keeps the destination's existing profile above that, untouched and deliberately not smoothed at the graft — smoothing would defeat the point of matching one object's layers to another's.

A bottom-right notification reports the graft height and, when it introduces an actual jump, the before/after values and which real layer it lands on — so the decision to touch it up is left to the user.

## Test plan

**Manual** (during development): copy a profile from a taller object, paste onto a shorter one and confirm the graft lands exactly at the destination's top layer; paste onto a taller one and confirm the untouched upper section is preserved; confirm the notification's reported values.

**Not yet done** — no automated test coverage exists for this feature (the diff only touches GUI files: `GLCanvas3D.cpp/.hpp`, `GUI_App.cpp`, `Plater.cpp`). Open to adding a headless unit test of the profile-merge math (the interpolation at the graft point, and the two truncate/extend branches) if that would help review.

## Known limitations

- No automated test coverage yet — see above.
- The graft point is never auto-smoothed, on purpose — flagging this since it's the kind of choice that could otherwise look like an oversight rather than a deliberate call.

---
Drafted with assistance from Claude Code — flagging that so it gets a closer look in review; happy to adjust anything that needs it.


Haidiye00 (commented):
Hello, please provide both the original and modified GIFs so we can verify the changes you've made. Thank you

MartinKlefas:
Thanks for the flag — here's a before/after showing the actual failure this PR fixes.

Before

**Before** — two objects on one plate, Print Sequence = By Layer, Prime Tower enabled, mismatched variable layer height profiles. Slicing fails with "The prime tower is only supported if all objects have the same variable layer height"

After

**After** — same plate, using the new Copy/Paste buttons on the Variable Layer Height tool to give the second object an exactly-matching profile instead of hand-painting it. Slicing now completes cleanly with the tower

Happy to record anything else that would help verify the change.

Haidiye00 (commented):
ok

